PR #68, opened 2024-10-29
fix: sarif: add result and kind

woodruffw:
See https://github.com/woodruffw/zizmor/issues/66.

Blocked by https://github.com/psastras/sarif-rs/issues/738.


src/sarif.rs
```diff
@@ -43,6 +70,14 @@ fn build_result(finding: &Finding<'_>) -> SarifResult {
         .message(finding.ident)
         .rule_id(finding.ident)
         .locations(build_locations(&finding.locations))
+        .level(
+            serde_json::to_value(ResultLevel::from(finding.determinations.severity))
```
woodruffw, line 74: NB: For some reason `.level(...)` takes `Into<Value>` instead of `Into<ResultLevel>`, hence this explicit `serde_json` call. I think this is an API deficiency on `serde-sarif`'s side, but not related to the blocking bug.

PR #70, opened 2024-10-29
Link to GitHub workflow examples

ncoghlan:
The boilerplate needed in a GitHub workflow to run `zizmor` and upload the SARIF file is very similar to the ESLint example in the GitHub docs, so linking directly to that will help make it clear how such a workflow can be written.

PR #71, opened 2024-10-29
Create workflow that runs zizmor latest

colindean:
This is a ~copy of what's working for me in [keepachangelog-parser-java][1].

[1]: https://github.com/colindean/keepachangelog-parser-java/blob/master/.github/workflows/zizmor.yml

If accepted, this could probably be linked from README.md instead of the ESLint example link in #70.

### DIFF
--- a/.github/workflows/zizmor.yaml
+++ b/.github/workflows/zizmor.yaml
@@ -5,7 +5,7 @@ on:
   push:
     branches: ["master", "main"]
   pull_request:
-    branches: ["**"]
+    branches: ["*"]
 
 permissions:
   contents: read
--- a/.github/workflows/zizmor.yaml
+++ b/.github/workflows/zizmor.yaml
@@ -1,0 +1,37 @@
+# https://github.com/woodruffw/zizmor
+name: GitHub Actions Security Analysis with Zizmor
+
+on:
+  push:
+    branches: ["master", "main"]
+  pull_request:
+    branches: ["**"]
+
+permissions:
+  contents: read
+
+jobs:
+  zizmor:
+    name: Zizmor latest via Cargo
+    runs-on: ubuntu-latest
+    permissions:
+      security-events: write
+    steps:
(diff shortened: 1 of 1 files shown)


ncoghlan:
I reworked the PR in #70 to link to the location where this workflow will be in the repository.

I kept the ESLint link as well, though, since GitHub have some genuinely useful links in the preamble to their example.

.github/workflows/zizmor.yaml
```diff
@@ -0,0 +1,37 @@
+# https://github.com/woodruffw/zizmor
+name: GitHub Actions Security Analysis with Zizmor
+
+on:
+  push:
+    branches: ["master", "main"]
+  pull_request:
+    branches: ["**"]
```
woodruffw, line 8: Small thing: shouldn't this be `*`?

```suggestion
    branches: ["*"]
```

woodruffw (approved):
Thanks @colindean, this is great! One small comment (might be me not understanding)

.github/workflows/zizmor.yaml
woodruffw: Ah nvm, TIL `**` means "match including `/`".

colindean: I'm not certain. I think that made its way into my example repo's config when I copy pasted somebody else's. Let's ship it as a fix and fix it later if it doesn't work I guess.

.github/workflows/zizmor.yaml
```diff
@@ -0,0 +1,37 @@
+# https://github.com/woodruffw/zizmor
+name: GitHub Actions Security Analysis with Zizmor
+
+on:
+  push:
+    branches: ["master", "main"]
```
woodruffw, line 6: Just for parsimony in this particular repo :slightly_smiling_face: 

```suggestion
    branches: ["main"]
```

woodruffw:
Thanks @colindean!

PR #83, opened 2024-10-30
fix: Always use the plain formatter even when the output is not a terminal

### LINKED ISSUE #76: Do not change the output format when stdout is a pipe
I tried piping the output of zizmor to `pbcopy` to copy into an issue, but the resulting output was changed from the human readable output to JSON. I also often pipe command output into `less`.

The format of the output should be controlled by a command line flag, not by the type of the stdout file.

asmeurer:
Fixes #76

I was also looking at whether color support should be modified when the terminal isn't a tty. I think this can be done by enabling supports-colors in owo_colors, although I'm not really a Rust dev and I couldn't quite figure out how to get this working right https://docs.rs/owo-colors/4.1.0/owo_colors/index.html#only-style-on-supported-terminals

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -1,8 +1,4 @@
-use std::{
-    io::{stdout, IsTerminal},
-    path::PathBuf,
-    time::Duration,
-};
+use std::{io::stdout, path::PathBuf, time::Duration};
 
 use anyhow::{anyhow, Context, Result};
 use audit::WorkflowAudit;
@@ -47,7 +43,6 @@ struct Args {
     gh_token: Option<String>,
 
     /// The output format to emit. By default, plain text will be emitted
-    /// on an interactive terminal and JSON otherwise.
     #[arg(long, value_enum)]
     format: Option<OutputFormat>,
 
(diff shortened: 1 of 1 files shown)


woodruffw:
Thanks @asmeurer! This LGTM.

> I was also looking at whether color support should be modified when the terminal isn't a tty. 

This shouldn't be an issue, since we use [`anstream::println!`](https://docs.rs/anstream/latest/anstream/) to render the output and that will filter colors and other control codes when the output isn't a TTY 🙂

woodruffw (approved):

asmeurer:
To be clear, I was testing out piping the output to `less` or to a file and the color codes were not being filtered (even when I tried using `if_supports_color` like in the owo docs example).

asmeurer:
I ran `cargo build --release` then `./target/release/zizmor` on a repo, both with a pipe to less and redirecting into a file (which I confirmed with a hex editor contains the ANSI codes). I wouldn't be surprised if something went awry in my building somehow, though.

woodruffw:
Weird. I'm at a loss to explain that -- I ran the following:

```
./target/debug/zizmor ~/personal/devel/gha-hazmat/ > /tmp/zizmor.txt
```

and opened `zizmor.txt`, and it's 100% plain text. I also tried with `| tee /tmp/zizmor.txt | less` interposed and also got plain text.

This might be a host OS thing. Are you on Windows?

asmeurer:
I am on a Mac.

woodruffw:
Okay, me too. I'm stumped then 🙂

daviddavis:
FWIW, I ran:

```
cargo build --release
./target/release/zizmor ../gha-hazmat > /tmp/zizmor.txt
```

And got plain text, no colors.

PR #90, opened 2024-10-31
feat: pin workflow action sha

chenrui333:
use https://github.com/stacklok/frizbee to pin the workflow action sha :)


.github/workflows/site.yml
```diff
@@ -26,24 +26,24 @@ jobs:
...
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4
         with:
           persist-credentials: false
 
       - name: Install the latest version of uv
-        uses: astral-sh/setup-uv@v3
+        uses: astral-sh/setup-uv@c5df01c9e4233fcd789bd004850aa36809d07727 # v3
```
woodruffw, line 34: Hmm, this doesn't look right -- I wonder if `frizbee` has a bug here?

The latest v3 is https://github.com/astral-sh/setup-uv/commit/3b9817b1bf26186f03ab8277bab9b827ea5cc254

woodruffw: Ran locally and got the same (wrong) ref:

https://github.com/astral-sh/setup-uv/commit/c5df01c9e4233fcd789bd004850aa36809d07727

chenrui333: yeah, indeed a bug, I will revert to https://github.com/astral-sh/setup-uv/commit/3b9817b1bf26186f03ab8277bab9b827ea5cc254 and investigate a bit with frizbee

woodruffw: https://github.com/stacklok/frizbee/issues/206

woodruffw (approved):
Thanks @chenrui333!

PR #92, opened 2024-10-31
Exclude `github.run_*` from template injection check

### LINKED ISSUE #91: Exclude `github.run_*` from template injection test
Running zizmor against my workflow triggered a template_injection alert for `github.run_id`. Based on [GitHub's documentation](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/accessing-contextual-information-about-workflow-runs) I don't see that there's any way that this could result in shell injection (barring a compromise of GitHub which means we have bigger problems).

PR incoming

funnelfiasco:
These are unique numbers assigned by GitHub and can't be used to inject shell commands.

Fixes #91

### DIFF
--- a/src/audit/template_injection.rs
+++ b/src/audit/template_injection.rs
@@ -39,6 +39,10 @@ const SAFE_CONTEXTS: &[&str] = &[
     // Safe keys within the otherwise generally unsafe github.event context.
     "github.event.number",
     "github.event.workflow_run.id",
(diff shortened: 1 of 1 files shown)


woodruffw:
Thanks @funnelfiasco! Agreed, these are harmless to expand.

woodruffw (approved):

PR #97, opened 2024-11-01
test: adds acceptance tests on top of json-formatted output

### LINKED ISSUE #55: Testing: add known-output tests
The diagnostic-style output isn't stable, but the JSON output is stable enough. So we should have tests that confirm that it doesn't diverge except for when expected (e.g. on new audits/behavioral changes).

ubiratansoares:
Closes #55 

I tried to cover all scenarios* that don't depend on the Github API token for now, but if you want I can add then by tweaking a bit the CI pipeline and using the `GITHUB_TOKEN` issued by GHA on these tests.

In addition, I did my best to make the tests read as simple as possible, moving some boilterplate into helper functions.

Let me know your thoughs!

### DIFF
--- a/tests/acceptance.rs
+++ b/tests/acceptance.rs
@@ -1,0 +1,161 @@
+use assert_cmd::Command;
+use serde_json::Value;
+use serde_json_path::JsonPath;
+use std::env::current_dir;
+
+// Acceptance tests for zizmor, on top of Json output
+// For now we don't cover tests that depends on Github API under the hood
+
+fn zizmor() -> Command {
+    Command::cargo_bin("zizmor").expect("Cannot create executable command")
+}
+
+fn workflow_under_test(name: &str) -> String {
+    let current_dir = current_dir().expect("Cannot figure out current directory");
+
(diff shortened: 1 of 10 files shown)


tests/acceptance.rs
```diff
@@ -0,0 +1,161 @@
...
+    // Don't bother about surrounding formatting
+    assert!(queried.to_string().contains(value));
+}
+
+#[test]
+fn audit_artipacked() -> anyhow::Result<()> {
+    let auditable = workflow_under_test("artipacked.yml");
+    let cli_args = ["--format", "json", &auditable];
+
+    let execution = zizmor().args(cli_args).unwrap();
```
ubiratansoares, line 40: Quick note : this `unwrap()` is not the same one from `Result` or `Option`. 

`args()` returns a `Command` type, hence there is no way to propagate an error here

woodruffw: Makes sense, thanks for explaining. Also TIL that `#[test]` functions can return result types!

woodruffw (approved):
Thanks @ubiratansoares!

Issue #12, opened 2024-09-05
Allow users to ignore findings


woodruffw:
In theory we could do something like:

```yaml
permissions: write-all # zizmor: ignore[excessive-permissions]
```

...but this is (1) annoying since it requires us to scan the YAML comments, and (2) might interfere with other tools that unfortunately read comments, like Dependabot.

ljharb:
It'd be nice to not have to put comments directly in the yml, too - iow, to optionally support a `.github/workflows/zizmor` config file of some kind.

woodruffw:
> It'd be nice to not have to put comments directly in the yml, too - iow, to optionally support a `.github/workflows/zizmor` config file of some kind.

That's a good idea!

`.github/zizmor.yml` would be a good place to put that, and it could be something like:

```yaml
version: 1
workflows:
  foobar.yml:
    ignore:
      - some-rule-name
```

(as a rough sketch)

ljharb:
(That would also allow CODEOWNERS to control access to the overrides)

michaelmior:
I personally would strongly prefer to put these as comments directly in the workflow YAML. That way whoever is responsible for that is also responsible for the overrides. Although I understand that might not be what everyone wants. If both approaches are possible, then great.

Another problem with the external config file is that you might want to ignore one rule only for part of a workflow. That gets messy to keep the two in sync.

ljharb:
Both sounds good to me; it's how eslint does it, and in practice it's fine to keep the two in sync.

woodruffw:
Yeah, I'm OK with having both. Comments will be little thorny in zizmor's case since there are other tools that use YAML comments for signaling (e.g. Dependabot supports version comments next to `uses:` clauses), but those can be documented edge cases.

(In terms of getting _some_ form of this in, I'm personally going to prioritize the config approach, since it requires fewer design thoughts about extending `yamlpath`. But I happily welcome people to help me out on the comment approach as well!)

funnelfiasco:
I think my ideal situation would be to ignore on a line-by-line basis, because there might be, e.g., one template injection that I know is safe in practice but I wouldn't want to ignore others that come up and might be unsafe. I recognize that it's more difficult.

One approach could be to have optional `line`s in the config file. For example:

```yaml
version: 1
workflows:
  foobar.yml:
    ignore:
      - template-injection
         line:
         - 8
         - 42
```

[...]

PR #104, opened 2024-11-02
feat: accept multiple arguments as inputs

miketheman:
Accept multiple individual files as inputs.
Useful with other mutli-file processing like xargs or other tools, so as to not have to invoke zizmor multiple times, and distinct from operating on an entire directory.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -47,8 +47,8 @@ struct Args {
     #[arg(long, value_enum)]
     format: Option<OutputFormat>,
 
-    /// The workflow filename or directory to audit.
-    input: PathBuf,
+    /// The workflow filenames or directories to audit.
+    inputs: Vec<PathBuf>,
 }
 
 #[derive(Debug, Copy, Clone, ValueEnum)]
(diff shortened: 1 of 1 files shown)


miketheman:
Easier to review ignoring whitespace: https://github.com/woodruffw/zizmor/pull/104/files?w=1

woodruffw (approved):

woodruffw (approved):
Thanks @miketheman!

woodruffw:
I thought a bit about this, and I think it might make sense to restrict the inputs slightly (but not as much as they were before). 

Concretely, instead of `<dir | file> [dir | file] ...`, I'm thinking `<dir | file [file ...]>` so that there's only ever one directory at most, but the user can still pass in multiple files explicitly if they please.

My rationale for this is that it'll make config (#12) discover a little easier: with just one directory, there's only one place where the config needs to be discovered.

Thoughts @miketheman?

woodruffw:
I thought about this more, and I'm probably overcomplicating it. For now I'll do autodiscovery via `${CWD}` and also support a `--config` flag; the rest can come later if needed.

PR #105, opened 2024-11-02
feat: pre-commit hook for other repos to use

miketheman:
Depends on #104

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -1,0 +1,6 @@
+- id: zizmor
+  name: zizmor
+  description: 'Find security issues in GitHub Actions CI/CD setups'
+  language: rust
+  files: \.github/workflows/.*\.yml$
+  entry: zizmor
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -102,3 +102,30 @@ as  GitHub's example of [running ESLint] as a security workflow.
 [repository workflow scan]: https://github.com/woodruffw/zizmor/blob/main/.github/workflows/zizmor.yml
 
 [running ESLint]: https://docs.github.com/en/code-security/code-scanning/integrating-with-code-scanning/uploading-a-sarif-file-to-github#example-workflow-that-runs-the-eslint-analysis-tool
+
+### Use with `pre-commit`
+
+`zizmor` can be used with the [`pre-commit`] framework.
+To do so, add the following to your `.pre-commit-config.yaml` `repos` section:
+
+```yaml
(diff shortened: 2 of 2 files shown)


docs/usage.md
```diff
@@ -102,3 +102,30 @@ as  GitHub's example of [running ESLint] as a security workflow.
...
 [running ESLint]: https://docs.github.com/en/code-security/code-scanning/integrating-with-code-scanning/uploading-a-sarif-file-to-github#example-workflow-that-runs-the-eslint-analysis-tool
+
+### Use with `pre-commit`
+
+`zizmor` can be used with the [`pre-commit`] framework.
+To do so, add the following to your `.pre-commit-config.yaml` `repos` section:
+
+```yaml
+-   repo: https://github.com/woodruffw/zizmor
+    rev: v0.1.6
```
miketheman, line 113: I _think_ this will only work after we push a new tag of this value, but I haven't tested that yet, I was using a local config to test in `.pre-commit-config.yaml` (not added to this scope):

```yaml
repos:
- repo: local
  hooks:
  - id: zizmor
    name: zizmor
    language: rust
    files: \.github/workflows/.*\.yml$
    entry: ./target/release/zizmor
```

.pre-commit-hooks.yaml
```diff
@@ -0,0 +1,6 @@
+- id: zizmor
+  name: zizmor
+  description: 'Find security issues in GitHub Actions CI/CD setups'
+  language: rust
+  files: \.github/workflows/.*\.yml$
+  entry: zizmor
```
woodruffw, line 6: Out of curiosity: how does `pre-commit` know to build + run `zizmor` here? Does it recognize that the repo contains a crate and then build it locally?

(I don't use pre-commit hooks much, so sorry if this is a silly question.)

miketheman: Not a silly question! `language: rust` tells pre-commit what to do.
Refs : https://pre-commit.com/index.html#rust

woodruffw: That's very cool!

docs/usage.md
woodruffw: Makes sense -- I can queue this into a release.

On a related note: it'd probably be nice to keep this `rev:` up-to-date automatically somehow, but I don't know of a good way to do that. Not a blocker, just something that came to mind :slightly_smiling_face:

woodruffw (approved):
Thanks @miketheman!

PR #122, opened 2024-11-04
Include a count of ignored rules when no issues are found

funnelfiasco:
This helps the user rest comfortably at night, knowing that zizmor found an issue, but correctly ignored it per user request.


woodruffw (approved):
Thanks!

PR #128, opened 2024-11-06
feat: Exclude safe values from template injection rule

funnelfiasco:
GitHub provides the following and I don't see a way an attacker could hijack them:

* runner.debug
* runner.os
* runner.temp

(runner.arch was already excluded from the check)

### DIFF
--- a/src/audit/template_injection.rs
+++ b/src/audit/template_injection.rs
@@ -51,6 +51,10 @@ const SAFE_CONTEXTS: &[&str] = &[
     "github.workspace",
     // GitHub Actions-controller runner architecture.
     "runner.arch",
+    // Debug logging is (1) or is not (0) enabled on GitHub Actions runner.
+    "runner.debug",
+    // GitHub Actions runner operating system.
+    "runner.os",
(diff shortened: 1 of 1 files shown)


woodruffw:
Thanks @funnelfiasco! The first two make sense to me, but `runner.temp` is hazy -- I think in practice it can't be attacker controlled, but I'm having trouble proving that (and in particular, if it can contain spaces anything else special, it can potentially be an injection source).

So for now, I think it makes sense to do the first two, and punt on `runner.temp`. Does that work for you?

funnelfiasco:
Updated commit with `runner.temp` removed for the time being. (Humorously, it's what inspired this PR in the first place. os and debug were things I saw along the way). I'll open an issue to have a discussion about temp, because I think it's worth excluding, but there's no reason to block this PR on it.

woodruffw:
> Edit: or would you rather it be a Discussion instead of an issue?

Discussion would be good, thanks!

woodruffw (approved):
Thanks!

PR #131, opened 2024-11-06
fix: bump github-actions-models

woodruffw:
Brings in https://github.com/woodruffw/github-actions-models/issues/14.


funnelfiasco:
I can confirm that this fixes the parsing error I was facing.

PR #136, opened 2024-11-07
fix: exclude information about the repo and owner

funnelfiasco:
This would seem to fall into the category of "if these values are unsafe, we have bigger problems to deal with"

### DIFF
--- a/src/audit/template_injection.rs
+++ b/src/audit/template_injection.rs
@@ -39,6 +39,13 @@ const SAFE_CONTEXTS: &[&str] = &[
     // Safe keys within the otherwise generally unsafe github.event context.
     "github.event.number",
     "github.event.workflow_run.id",
+    // Information about the GitHub repostory
+    "github.repository",
+    "github.repository_id",
+    "github.repostoryUrl",
+    // Information about the GitHub repository owner (account/org or ID)
+    "github.repository_owner",
+    "github.repository_owner_id",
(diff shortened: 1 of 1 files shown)


woodruffw:
Thanks @funnelfiasco! I think this is going to be similar to https://github.com/woodruffw/zizmor/pull/128 -- excluding `repository_id` and `repository_owner_id` seems 100% right to me, but I need to think about the others. 

(AFAIK repo/owner information is _probably_ safe, but URL quoting/escaping isn't always super consistent, and there may be funny things that someone can do with Enterprise deployment URLs.)

funnelfiasco:
Of course GitHub doesn't provide documentation on what's actually possible, but in the case of these values there's nothing an attacker could do to modify them. So even if it's possible and someone names their repository `Robert'); DROP TABLE Students;--` then plenty of other things will probably break before they ever get to running a GitHub workflow.

woodruffw:
> Of course GitHub doesn't provide documentation on what's actually possible, but in the case of these values there's nothing an attacker could do to modify them. 

I think that's right, except for in the fork context -- in that context the attacker controls all of these, but I agree that there's still no injection risk with them (I suppose someone could name their GitHub account `reboot` or something like that, but they still don't have the escape primitives needed).

Given that, I'm good with this. Thanks for the PR!

woodruffw (approved):

PR #150, opened 2024-11-12
docs: update version in pre-commit hook

baggiponte:
Ciao! Didn't open an issue for this as it seemed straightforward. I know you suggest the users to update the hook version, but I figured it wouldn't be bad to update from time to time.

This pull request consists of an update to the documentation to reflect the latest version of zizmor in the suggested pre-commit-config file.

* [`docs/usage.md`](diffhunk://#diff-72376d0b487d7231cf31959bf266f6aea098e899743bae02e36d176cef4db476L186-R186): Updated the version of the `zizmor` repository from `v0.1.6` to `v0.3.0` in the `.pre-commit-config.yaml` file.

### DIFF
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -183,7 +183,7 @@ To do so, add the following to your `.pre-commit-config.yaml` `repos` section:
 
 ```yaml
 -   repo: https://github.com/woodruffw/zizmor
-    rev: v0.1.6 # (1)!
+    rev: v0.3.0 # (1)!
     hooks:
     - id: zizmor
 ```


woodruffw (approved):
Thanks! Yeah, no issue needed, since this is trivial.

woodruffw:
Longer-term, we should keep this example updated like we do with `zizmor --help`, via a snippet: https://github.com/woodruffw/zizmor/blob/ec24c60280bad9d3ff78212da805c6f4e247acf0/docs/quickstart.md?plain=1#L7-L9

But that's a separate concern 🙂

woodruffw:
Thanks @baggiponte!

baggiponte:
> Longer-term, we should keep this example updated like we do with `zizmor --help`, via a snippet:
> 
> https://github.com/woodruffw/zizmor/blob/ec24c60280bad9d3ff78212da805c6f4e247acf0/docs/quickstart.md?plain=1#L7-L9
> 
> But that's a separate concern 🙂

I did not know about this trick!

> Thanks @baggiponte!

De nada, always a pleasure. Amazing work with the tool - I don't really remember how I found out about it, but since I know so frighteningly little about GHA (despite running a bunch of them) it immediately ended up in my precommit config 🤣

PR #164, opened 2024-11-16
Fix typos including `github.repostoryUrl` -> `github.repositoryUrl`

hugovk:
One of these likely causes a bug in functionality.

Here's the correct spelling also in the docs:

https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/accessing-contextual-information-about-workflow-runs#github-context

Found with https://github.com/codespell-project/actions-codespell.

### DIFF
--- a/src/audit/ref_confusion.rs
+++ b/src/audit/ref_confusion.rs
@@ -2,7 +2,7 @@
 //!
 //! This is similar to "impostor" commit detection, but with only named
 //! refs instead of fully pinned commits: a user may pin a ref such as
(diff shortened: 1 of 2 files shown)


woodruffw (approved):
Thanks for catching the @hugovk! Looks like this would have caused some false positives, since that's in the safe contexts list.

CCing @funnelfiasco for viz since I believe this is one of the ones you added.

funnelfiasco:
Thanks for catching this, @hugovk. I'm impressed that I managed to get it wrong twice 😅

PR #187, opened 2024-11-22
feat: adds support to inlined ignores

### LINKED ISSUE #124: Support ignoring results via inline comments?
`zizmor` now supports a config file that can be used to write ignore rules. However, some users also want inline ignore support, e.g.:

```yaml
permissions: write-all # zizmor: ignore[excessive-permissions]
```

See #12 for some earlier context.

ubiratansoares:
Closes #124 

There are 2 approaches here

- https://github.com/woodruffw/zizmor/commit/8f1a18cd3ec00da493502df35abacd1ec32756f4 propose a change at `FindingRegistry` level
- https://github.com/woodruffw/zizmor/commit/14d2dd04b0341fa176b2ad6e2ebbc5073243e076 proposes a change at `FindingBuilder` level

In both of them the core remains the same : 

- we cache the YAML string contents at `Workflow` 
- we use `Finding.location` to extract the start and end rows related to the impacted snippet
- we loop through the lines, checking whether or not we have a comment disabling that particular check

Happy to iterate on this idea. If one of them makes sense, happy to add a commit with proper docs

How to test

- Pull this branch
- Run

```bash
cargo run -- tests/test-data/inlined-ignores.yml
```

```
➜  zizmor git:(ufs/patch-06) cargo run -- tests/test-data/inlined-ignores.yml
   Compiling zizmor v0.4.0 (/Users/lonewolf/RustroverProjects/zizmor)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 1.20s
     Running `target/debug/zizmor tests/test-data/inlined-ignores.yml`
🌈 completed inlined-ignores.yml
No findings to report. Good job! (3 ignored)
```

### DIFF
--- a/src/finding/mod.rs
+++ b/src/finding/mod.rs
@@ -300,7 +300,7 @@ impl<'w> FindingBuilder<'w> {
             })
             .collect::<Vec<_>>();
 
-        let inlined_ignore = format!("zizmor: ignore[{}]", id);
+        let inlined_ignore = format!("# zizmor: ignore[{}]", id);
         for (start, end) in line_ranges {
             for document_line in start..(end + 1) {
                 if let Some(line) = document_lines.get(document_line) {
--- a/src/finding/mod.rs
+++ b/src/finding/mod.rs
@@ -304,7 +304,7 @@ impl<'w> FindingBuilder<'w> {
         for (start, end) in line_ranges {
             for document_line in start..(end + 1) {
                 if let Some(line) = document_lines.get(document_line) {
-                    if line.contains(&inlined_ignore) {
+                    if line.rfind(&inlined_ignore).is_some() {
                         return true;
                     }
                 }
--- a/src/finding/mod.rs
+++ b/src/finding/mod.rs
@@ -222,7 +222,7 @@ pub(crate) struct Finding<'w> {
     pub(crate) url: &'static str,
     pub(crate) determinations: Determinations,
     pub(crate) locations: Vec<Location<'w>>,
-    pub(crate) ignored_from_inlined_comment: bool,
+    pub(crate) ignored: bool,
 }
 
 pub(crate) struct FindingBuilder<'w> {
@@ -279,7 +279,7 @@ impl<'w> FindingBuilder<'w> {
                 severity: self.severity,
             },
             locations,
-            ignored_from_inlined_comment: should_ignore,
+            ignored: should_ignore,
         })
     }
 
@@ -289,7 +289,7 @@ impl<'w> FindingBuilder<'w> {
         locations: &[Location],
         id: &str,
     ) -> bool {
-        let document_lines = &workflow.contents.lines().collect::<Vec<_>>();
+        let document_lines = &workflow.document.source().lines().collect::<Vec<_>>();
         let line_ranges = locations
             .iter()
             .map(|l| {
--- a/src/finding/mod.rs
+++ b/src/finding/mod.rs
@@ -222,6 +222,7 @@ pub(crate) struct Finding<'w> {
     pub(crate) url: &'static str,
     pub(crate) determinations: Determinations,
     pub(crate) locations: Vec<Location<'w>>,
+    pub(crate) ignored_from_inlined_comment: bool,
 }
 
 pub(crate) struct FindingBuilder<'w> {
@@ -261,6 +262,14 @@ impl<'w> FindingBuilder<'w> {
     }
 
     pub(crate) fn build(self, workflow: &'w Workflow) -> Result<Finding<'w>> {
+        let locations = self
+            .locations
+            .iter()
+            .map(|l| l.clone().concretize(workflow))
+            .collect::<Result<Vec<_>>>()?;
+
+        let should_ignore = self.ignored_from_inlined_comment(workflow, &locations, self.ident);
+
         Ok(Finding {
             ident: self.ident,
             desc: self.desc,
@@ -269,11 +278,39 @@ impl<'w> FindingBuilder<'w> {
                 confidence: self.confidence,
                 severity: self.severity,
             },
-            locations: self
-                .locations
-                .into_iter()
-                .map(|l| l.concretize(workflow))
-                .collect::<Result<Vec<_>>>()?,
+            locations,
+            ignored_from_inlined_comment: should_ignore,
         })
+    }
+
+    fn ignored_from_inlined_comment(
+        &self,
+        workflow: &Workflow,
+        locations: &[Location],
+        id: &str,
+    ) -> bool {
+        let document_lines = &workflow.contents.lines().collect::<Vec<_>>();
+        let line_ranges = locations
+            .iter()
+            .map(|l| {
+                (
+                    l.concrete.location.start_point.row,
+                    l.concrete.location.end_point.row,
+                )
+            })
+            .collect::<Vec<_>>();
+
+        let inlined_ignore = format!("zizmor: ignore[{}]", id);
+        for (start, end) in line_ranges {
+            for document_line in start..(end + 1) {
+                if let Some(line) = document_lines.get(document_line) {
+                    if line.contains(&inlined_ignore) {
+                        return true;
+                    }
+                }
+            }
+        }
+
+        false
     }
 }
--- a/src/main.rs
+++ b/src/main.rs
@@ -177,13 +177,12 @@ fn run() -> Result<ExitCode> {
             workflow = workflow.filename().cyan()
         ));
         for (name, audit) in audit_registry.iter_workflow_audits() {
-            let findings = audit.audit(workflow).with_context(|| {
+            results.extend(audit.audit(workflow).with_context(|| {
                 format!(
                     "{name} failed on {workflow}",
                     workflow = workflow.filename()
                 )
-            })?;
-            results.extend(findings);
+            })?);
             bar.inc(1);
         }
         bar.println(format!(
--- a/src/main.rs
+++ b/src/main.rs
@@ -177,12 +177,13 @@ fn run() -> Result<ExitCode> {
             workflow = workflow.filename().cyan()
         ));
         for (name, audit) in audit_registry.iter_workflow_audits() {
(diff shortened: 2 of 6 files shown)


src/models.rs
```diff
@@ -17,6 +17,7 @@ use crate::finding::{Route, SymbolicLocation};
 /// providing access to the underlying data model.
 pub(crate) struct Workflow {
     pub(crate) path: String,
+    pub(crate) contents: String,
```
woodruffw, line 20: I might be missing something, but why do we need another copy of the YAML document here? `workflow.document.source()` should return a borrowed (`&str`) copy that should be reusable here :slightly_smiling_face:

woodruffw:
Thanks @ubiratansoares! I like the `FindingBuilder` approach, let's move forwards with that :slightly_smiling_face: 

(Apologies for the merge, that was a force of habit. If it makes things annoying here, please feel free to force-push over it.)

src/finding/mod.rs
```diff
@@ -221,6 +221,7 @@ pub(crate) struct Finding<'w> {
     pub(crate) desc: &'static str,
     pub(crate) determinations: Determinations,
     pub(crate) locations: Vec<Location<'w>>,
+    pub(crate) ignored_from_inlined_comment: bool,
```
woodruffw, line 224: Let's shorten this, since `ignored` is unambiguous in this context:

```suggestion
    pub(crate) ignored: bool,
```

src/finding/mod.rs
```diff
@@ -267,18 +268,54 @@ impl<'w> FindingBuilder<'w> {
...
+        let inlined_ignore = format!("zizmor: ignore[{}]", id);
+        for (start, end) in line_ranges {
+            for document_line in start..(end + 1) {
+                if let Some(line) = document_lines.get(document_line) {
+                    if line.contains(&inlined_ignore) {
+                        return true;
+                    }
+                }
+            }
+        }
```
woodruffw, line 317: This approach looks decent, but I *think* I'd prefer if we try and do this with the `yamlpath` APIs. My thinking there is that `yamlpath` will ensure that we're only ever parsing within the comment's own span.

The situations in which this matters are pretty obscure, but it's good to avoid them if we can, e.g.:

```yaml
uses: blah
with:
  # false positive
  foo: "zizmor: ignore[rule]"
  # false positive even if we anchor on #
  bar: "# zizmor: ignore[rule]"
  # two matches, unclear which is right
  baz: "something" # zizmor: ignore[rule] # zizmor: ignore[rule] 
```

`yamlpath` should avoid these cases, or at least give us a clear decision procedure about what to do with each :slightly_smiling_face:

src/finding/mod.rs
```diff
@@ -267,18 +268,54 @@ impl<'w> FindingBuilder<'w> {
...
+            .iter()
+            .map(|l| {
+                (
+                    l.concrete.location.start_point.row,
+                    l.concrete.location.end_point.row,
+                )
+            })
+            .collect::<Vec<_>>();
+
+        let inlined_ignore = format!("zizmor: ignore[{}]", id);
```
woodruffw, line 308: Flagging for discussion: we can think about this more in a follow-up, but I think we want a bit more syntax here to enable multiple ignores.

For example:

1. Ignore everything: `# zizmor: ignore[*]`
2. Ignore two specific rules: `# zizmor: ignore[rulea,ruleb]`

We can start with the current behavior though, to avoid making this PR too big :slightly_smiling_face:

src/main.rs
```diff
@@ -176,12 +176,13 @@ fn run() -> Result<ExitCode> {
...
         for (name, audit) in audit_registry.iter_workflow_audits() {
-            results.extend(audit.audit(workflow).with_context(|| {
+            let findings = audit.audit(workflow).with_context(|| {
                 format!(
                     "{name} failed on {workflow}",
                     workflow = workflow.filename()
                 )
-            })?);
+            })?;
+            results.extend(findings);
```
woodruffw, line 185: Is this new `findings` variable needed?

src/models.rs
ubiratansoares: You definetely did not miss anything, I actually missed that `document.source()` API! Super handy! I'll change that

src/main.rs
ubiratansoares: Definetely not needed! Just a leftover from the first take 🙈

ubiratansoares: Solved [here](https://github.com/woodruffw/zizmor/pull/187/commits/0de3177ed5dd2f3e944b3e2edc527742e4531a5e)

src/finding/mod.rs
ubiratansoares: Done [here](https://github.com/woodruffw/zizmor/pull/187/commits/0de3177ed5dd2f3e944b3e2edc527742e4531a5e)

src/models.rs
ubiratansoares: Fixed [here](https://github.com/woodruffw/zizmor/pull/187/commits/0de3177ed5dd2f3e944b3e2edc527742e4531a5e)

src/finding/mod.rs
ubiratansoares: I totally agree with you. The best I can think of to improve a bit over those false positives is trying with match with regex instead.

On the other hand, perphaps this can be only a first version for this support (we could tag it as `experimental`). Later on, we can iterate over `yamlpath` and replace the implementation which based on the comment extraction API.

You have the call, if you decide this impl. does not work, I'm happy with a few things I've learned along the way 🙂

woodruffw: > On the other hand, perphaps this can be only a first version for this support (we could tag it as `experimental`). Later on, we can iterate over `yamlpath` and replace the implementation which based on the comment extraction API.

That works for me -- I'm okay with a non-`yamlpath` approach as long as we don't box ourselves in here.

In the mean time, I *think* we can reduce the false positives here by matching on `# zizmor: ignore[...]` (note the `#` anchor) and searching from the string's right instead of left (i.e. `rfind` instead of `contains`).

(We could then make sure it's a comment by making sure nothing follows it, but that might be overkill.)

src/finding/mod.rs
```diff
@@ -267,18 +268,54 @@ impl<'w> FindingBuilder<'w> {
...
     }
+
+    fn ignored_from_inlined_comment(
+        &self,
+        workflow: &Workflow,
+        locations: &[Location],
+        id: &str,
+    ) -> bool {
+        let document_lines = &workflow.document.source().lines().collect::<Vec<_>>();
+        let line_ranges = locations
```
woodruffw, line 298: Small thing: I'm wondering if it makes sense to scan all locations like this, or require that the ignore comment is present only on the first location in the finding.

I don't feel strongly either way, but my thinking is that forcing it to be only on the first location keeps it at the "top" in general, and reduces the amount of scanning we need to do. Thoughts?

ubiratansoares: I gave a small try on the following change (local only)

```rust
        if let Some(first_location) = locations.first() {
            let start_line = first_location.concrete.location.start_point.row;
            let end_line = first_location.concrete.location.end_point.row + 1;
            let inlined_ignore = format!("# zizmor: ignore[{}]", id);

            for document_line in start_line..end_line {
                if let Some(line) = document_lines.get(document_line) {
                    if line.rfind(&inlined_ignore).is_some() {
                        return true;
                    }
                }
            }
        }
```

and tried against the following workflow, with exercises an `artipacked` finding

```yml
on:
  push:
    branches:
      - master
  workflow_dispatch:

jobs:
  artipacked:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v4

      - name: Upload artifact
        uses: actions/upload-artifact@v4
        with:
          name: artipacked
          path: . # zizmor: ignore[artipacked]
```

I got

```
error[artipacked]: credential persistence through GitHub Actions artifacts
  --> tests/test-data/artipacked.yml:13:9
   |
13 |         - uses: actions/checkout@v4
   |           ^^^^^^^^^^^^^^^^^^^^^^^^^ does not set persist-credentials: false
14 |
15 |         - name: Upload artifact
   |  _________^
16 | |         uses: actions/upload-artifact@v4
17 | |         with:
18 | |           name: artipacked
19 | |           path: . # zizmor: ignore[artipacked]
   | |______________________________________________^ may leak the credentials persisted above
   |
```

I have mixed feelings whether or not users may get confused with this, since we are not ignoring as requested. Strictly speaking from the user perspective, may take is going after for all `locations`. 

Thinking on future audits that may track two or more `locations` (right now only `artipacked` and `template-injection` does that from what I could learn), the behavior for inlined ignores stay the same if we scan all of them, so we don't need to care about special cases in which the meaningful `location` to check for an `ignore` is not the first one.

Wdyt?

woodruffw: Oh yeah, that's a good point -- I could see people expecting that `ignore` to work.

src/models.rs
woodruffw: Thanks!

woodruffw (commented):
Thanks @ubiratansoares!

I think I'd like to see https://github.com/woodruffw/zizmor/pull/187#discussion_r1855269066 in this PR, but otherwise everything looks fantastic to me (and we can start follow-up issues/PRs for the other things).

src/finding/mod.rs
woodruffw: https://github.com/woodruffw/zizmor/issues/189

src/finding/mod.rs
```diff
@@ -269,11 +278,39 @@ impl<'w> FindingBuilder<'w> {
...
+            .iter()
+            .map(|l| {
+                (
+                    l.concrete.location.start_point.row,
+                    l.concrete.location.end_point.row,
+                )
+            })
+            .collect::<Vec<_>>();
+
+        let inlined_ignore = format!("zizmor: ignore[{}]", id);
```
woodruffw, line 303:
```suggestion
        let inlined_ignore = format!("# zizmor: ignore[{}]", id);
```

woodruffw (approved):
Thanks @ubiratansoares!

PR #192, opened 2024-11-24
feat: adds new github-env audit

ubiratansoares:
- Pulls dangerous triggers detection logic from `WorkflowAudit` and moves to `impl DangerousTriggers` instance methods
- Reuses only that logic in the context of to implement a new audit

```bash
cargo run -- tests/test-data/github_env.yml

error[github-env]: dangerous use of $GITHUB_ENV
  --> tests/test-data/github_env.yml:12:9
   |
12 |           run: |
   |  _________^
13 | |           message=$(echo "$TITLE" | grep -oP '[{\[][^}\]]+[}\]]' | sed 's/{\|}\|\[\|\]//g')
14 | |           echo "message=$message" >> $GITHUB_ENV
   | |_________________________________________________^ GITHUB_ENV used in the context of a dangerous Workflow trigger
   |

2 findings (1 ignored): 0 unknown, 0 informational, 0 low, 0 medium, 1 high
```

Some questions below!

### DIFF
--- a/docs/audits.md
+++ b/docs/audits.md
@@ -650,10 +650,12 @@ In general, users should use for [Github Actions environment files]
 
 [github-env.yml]: https://github.com/woodruffw/gha-hazmat/blob/main/.github/workflows/github-env.yml
 
-Detects dangerous usages of `$GITHUB_ENV` environment variable.
+Detects dangerous usages of the `GITHUB_ENV` environment variable.
 
-This audit compounds over `dangerous-triggers`: given the proper circumstances, an attacker
-may achieve code execution by controlling this environment variable.
+When used in workflows with dangerous triggers (such as `pull_request_target` and `workflow_run`), 
+`GITHUB_ENV` can be an arbitrary code execution risk. In particular, if the attacker is able to set
+arbitrary variables or variable contents via `GITHUB_ENV`, they made be able to set `LD_PRELOAD`
+or otherwise induce code execution implicitly within subsequent steps.
 
 Other resources:
 
@@ -662,9 +664,10 @@ Other resources:
 
 ### Remediation
 
-In general, you should avoid using `$GITHUB_ENV` when relying on dangerous Workflow triggers,
-especially for use cases like caching state or passing state around. For those, `$GITHUB_OUTPUT`
-may be an alternative.
+In general, you should avoid setting `GITHUB_ENV` within workflows that are attacker-triggered, 
+like `pull_request_target`.
+
+If you need to pass state between steps, consider using `GITHUB_OUTPUT` instead.
 
 
 [ArtiPACKED: Hacking Giants Through a Race Condition in GitHub Actions Artifacts]: https://unit42.paloaltonetworks.com/github-repo-artifacts-leak-tokens/
--- a/docs/audits.md
+++ b/docs/audits.md
@@ -642,6 +642,31 @@ In general, users should use for [Github Actions environment files]
         echo "$HOME/.local/my-bin" >> "$GITHUB_PATH"
     ```
 
+## `github-env`
+
+| Type     | Examples           | Introduced in | Works offline  | Enabled by default |
+|----------|--------------------|---------------|----------------|--------------------|
+| Workflow  | [github-env.yml]   | v0.6.0        | ✅             | ✅                 |
+
+[github-env.yml]: https://github.com/woodruffw/gha-hazmat/blob/main/.github/workflows/github-env.yml
+
+Detects dangerous usages of `$GITHUB_ENV` environment variable.
+
+This audit compounds over `dangerous-triggers`: given the proper circumstances, an attacker
+may achieve code execution by controlling this environment variable.
+
+Other resources:
+
+* [GitHub Actions exploitation: environment manipulation]
+* [GHSL-2024-177: Environment Variable injection in an Actions workflow of Litestar]
+
+### Remediation
+
+In general, you should avoid using `$GITHUB_ENV` when relying on dangerous Workflow triggers,
+especially for use cases like caching state or passing state around. For those, `$GITHUB_OUTPUT`
+may be an alternative.
+
+
 [ArtiPACKED: Hacking Giants Through a Race Condition in GitHub Actions Artifacts]: https://unit42.paloaltonetworks.com/github-repo-artifacts-leak-tokens/
 [Keeping your GitHub Actions and workflows secure Part 1: Preventing pwn requests]: https://securitylab.github.com/resources/github-actions-preventing-pwn-requests/
 [What the fork? Imposter commits in GitHub Actions and CI/CD]: https://www.chainguard.dev/unchained/what-the-fork-imposter-commits-in-github-actions-and-ci-cd
@@ -654,3 +679,5 @@ In general, users should use for [Github Actions environment files]
 [were deprecated by Github]: https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/
 [Github Actions environment files]: https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/workflow-commands-for-github-actions#environment-files
 [Semgrep audit]: https://semgrep.dev/r?q=yaml.github-actions.security.allowed-unsecure-commands.allowed-unsecure-commands
+[GitHub Actions exploitation: environment manipulation]: https://www.synacktiv.com/en/publications/github-actions-exploitation-repo-jacking-and-environment-manipulation
+[GHSL-2024-177: Environment Variable injection in an Actions workflow of Litestar]: https://securitylab.github.com/advisories/GHSL-2024-177_Litestar/
--- a/src/audit/dangerous_triggers.rs
+++ b/src/audit/dangerous_triggers.rs
@@ -15,7 +15,7 @@ audit_meta!(
 
 impl WorkflowAudit for DangerousTriggers {
     fn new(_: AuditState) -> Result<Self> {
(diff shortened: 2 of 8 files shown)


src/audit/github_env.rs
```diff
@@ -0,0 +1,49 @@
...
+        let workflow = step.workflow();
+
+        let dangerous_triggers_detector = DangerousTriggers {};
+
+        let has_dangerous_triggers = dangerous_triggers_detector.has_workflow_run(workflow)
+            || dangerous_triggers_detector.has_pull_request_target(workflow);
+
+        if has_dangerous_triggers {
+            if let StepBody::Run { run, .. } = &step.deref().body {
+                if run.contains("$GITHUB_ENV") {
```
ubiratansoares, line 33: Super naive here. Advice request 😄

src/audit/github_env.rs
```diff
@@ -0,0 +1,49 @@
...
+                    findings.push(
+                        Self::finding()
+                            .severity(Severity::High)
+                            .confidence(Confidence::High)
```
ubiratansoares, line 37: Not sure about these two!

woodruffw:
Thanks @ubiratansoares! I'll give this a more detailed review later today.

> Pulls dangerous triggers detection logic from `WorkflowAudit` and moves to `impl DangerousTriggers` instance methods

Rather than use sharing APIs across audits, I think we can probably turn these into APIs on `models::Workflow` -- `Workflow::has_pull_request_target` and similar would probably be good.

src/audit/github_env.rs
woodruffw: Thanks! We could definitely be at least a little more general here: we should also catch the `${GITHUB_ENV}` variant.

src/audit/github_env.rs
woodruffw: `Severity::High` makes sense to me, but I think our confidence should be lower since we're not doing any real analysis here yet. Maybe `Low` or `Medium`?

src/audit/dangerous_triggers.rs
```diff
@@ -7,38 +7,40 @@ use crate::finding::{Confidence, Finding, Severity};
 use crate::models::Workflow;
 use crate::state::AuditState;
 
-pub(crate) struct DangerousTriggers {
-    pub(crate) _state: AuditState,
-}
+pub(crate) struct DangerousTriggers {}
```
woodruffw, line 10: Nit: you can make this a ZST:

```suggestion
pub(crate) struct DangerousTriggers;
```

src/audit/dangerous_triggers.rs
```diff
@@ -7,38 +7,40 @@ use crate::finding::{Confidence, Finding, Severity};
...
             Trigger::BareEvents(events) => events.contains(&BareEvent::WorkflowRun),
             Trigger::Events(events) => !matches!(events.workflow_run, OptionalBody::Missing),
-        };
+        }
+    }
+}
 
+impl WorkflowAudit for DangerousTriggers {
+    fn new(_: AuditState) -> Result<Self> {
+        Ok(Self {})
```
woodruffw, line 38:
```suggestion
        Ok(Self)
```

src/audit/github_env.rs
ubiratansoares: `Low` sounds good to me!

src/audit/dangerous_triggers.rs
ubiratansoares: Done here https://github.com/woodruffw/zizmor/pull/192/commits/ee0f5ef3bc7e05a5c3860e22d1346059c7430b8a

src/audit/dangerous_triggers.rs
ubiratansoares: Done here https://github.com/woodruffw/zizmor/pull/192/commits/ee0f5ef3bc7e05a5c3860e22d1346059c7430b8a

ubiratansoares:
I address all the comments, including moving dangerous triggers detection logic to  `models::Workflow`. 

I also added some docs covering the basics based on the discussions we have in the related issue, but I did not go deep there are safe ways to use `$GITHUB_ENV`. Happy to extended those docs based on additional input!

src/audit/github_env.rs
ubiratansoares: Done here https://github.com/woodruffw/zizmor/pull/192/commits/efa8d04ec30971590eb9b8bf0f433a595d11db0a

ubiratansoares:
@woodruffw we miss an example [here](https://github.com/woodruffw/gha-hazmat/blob/main/.github/workflows/github-env.yml), happy to raise a PR there, lmk!

woodruffw:
> @woodruffw we miss an example [here](https://github.com/woodruffw/gha-hazmat/blob/main/.github/workflows/github-env.yml?rgh-link-date=2024-11-24T21%3A49%3A34Z), happy to raise a PR there, lmk!

Please do, thanks!

src/audit/github_env.rs
```diff
@@ -0,0 +1,56 @@
...
+                            .severity(Severity::High)
+                            .confidence(Confidence::Low)
+                            .add_location(step.location().with_keys(&["run".into()]).annotated(
+                                "GITHUB_ENV used in the context of a dangerous Workflow trigger",
+                            ))
+                            .build(step.workflow())?,
+                    )
+                }
+            }
+        }
```
woodruffw, line 52: We could save ourselves an indent here by doing an early return instead:

```rust
if !has_dangerous_triggers {
    return Ok(findings)
}
```

src/audit/github_env.rs
```diff
@@ -0,0 +1,56 @@
...
+
+audit_meta!(GitHubEnv, "github-env", "dangerous use of $GITHUB_ENV");
+
+impl GitHubEnv {
+    fn uses_github_environment(&self, run_step_body: &str) -> bool {
+        // In the future we can improve over this implementation,
+        // eventually detecting how $GITHUB_ENV is being used
+        // and returning an Option<Confidence> instead
+
+        run_step_body.contains("$GITHUB_ENV") || run_step_body.contains("${GITHUB_ENV}")
```
woodruffw, line 18: Just thinking out loud: in terms of improving this, we could probably use a regexp like:

```
^.+\s*>>?\s*"?\$\{?GITHUB_ENV\}?"?.*$
```

That'll catch a bunch of common cases, but it's also a mess. So we can defer that for a follow-up.

docs/audits.md
```diff
@@ -642,6 +642,31 @@ In general, users should use for [Github Actions environment files]
...
 
+## `github-env`
+
+| Type     | Examples           | Introduced in | Works offline  | Enabled by default |
+|----------|--------------------|---------------|----------------|--------------------|
+| Workflow  | [github-env.yml]   | v0.6.0        | ✅             | ✅                 |
+
+[github-env.yml]: https://github.com/woodruffw/gha-hazmat/blob/main/.github/workflows/github-env.yml
+
+Detects dangerous usages of `$GITHUB_ENV` environment variable.
```
woodruffw, line 653:
```suggestion
Detects dangerous usages of the `GITHUB_ENV` environment variable.
```

docs/audits.md
```diff
@@ -642,6 +642,31 @@ In general, users should use for [Github Actions environment files]
...
+
+This audit compounds over `dangerous-triggers`: given the proper circumstances, an attacker
+may achieve code execution by controlling this environment variable.
```
woodruffw, line 656:
```suggestion
When used in workflows with dangerous triggers (such as `pull_request_target` and `workflow_run`), `GITHUB_ENV` can be an arbitrary code execution risk.

In particular, if the attacker is able to set arbitrary variables or variable contents via `GITHUB_ENV`, they made be able to set `LD_PRELOAD` or otherwise induce code execution implicitly within subsequent steps.
```

docs/audits.md
```diff
@@ -642,6 +642,31 @@ In general, users should use for [Github Actions environment files]
...
+
+* [GitHub Actions exploitation: environment manipulation]
+* [GHSL-2024-177: Environment Variable injection in an Actions workflow of Litestar]
+
+### Remediation
+
+In general, you should avoid using `$GITHUB_ENV` when relying on dangerous Workflow triggers,
+especially for use cases like caching state or passing state around. For those, `$GITHUB_OUTPUT`
+may be an alternative.
+
```
woodruffw, line 668:
```suggestion
In general, you should avoid setting `GITHUB_ENV` within workflows that are attacker-triggered, like `pull_request_target`.

If you need to pass state between steps, consider using `GITHUB_OUTPUT` instead.
```

src/audit/github_env.rs
ubiratansoares: Great catch! Done [here](https://github.com/woodruffw/zizmor/pull/192/commits/b7aef161fb9cb7adccae49d625db94aa4b2ef3e9)

src/audit/github_env.rs
ubiratansoares: Done [here](https://github.com/woodruffw/zizmor/pull/192/commits/f222f8655f3f0d4a184982ef551822fbd9c07fea) with all other suggestions for docs!

docs/audits.md
ubiratansoares: Done [here](https://github.com/woodruffw/zizmor/pull/192/commits/f222f8655f3f0d4a184982ef551822fbd9c07fea) with all suggestions for docs 🙂

woodruffw (approved):

woodruffw (approved):
LGTM, thanks @ubiratansoares!

I'll make a follow-up to track refinements here.

Issue #197, opened 2024-11-24
github-env: refinements


woodruffw:
The `github-env` audit merged in #192 is barebones. Some things we should do to improve it:

* Use a regex to match suspicious `>> $GITHUB_ENV` shell lines; this won't be perfect, but it'll avoid false positives like `echo $GITHUB_ENV`.
    * Candidate expression: `^.+\s*>>?\s*"?\$\{?GITHUB_ENV\}?"?.*$`
* Handle `actions/github-script` bodies as well, since people are likely to modify `GITHUB_ENV` there too.

woodruffw:
For PowerShell, we need to detect patterns like `... | Out-File -FilePath $env:GITHUB_ENV -Append` and `... >> $env:GITHUB_ENV`.

woodruffw:
Another, more serious option here would be to use `tree-sitter` grammars as necessary:

* https://crates.io/crates/tree-sitter-bash
* https://github.com/airbus-cert/tree-sitter-powershell (no Rust crate yet)

woodruffw:
> @woodruffw It seems that [Python is also a shell option](https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#defaultsrunshell) 🙀 

Oh yeah, it gets way worse 🙂 -- there's no way we'll end up being fully general here. But I bet we can come up with some good patterns there as well. 

> I gave a try on tree-sitter-bash here. Hope it is in the right direction 

Thanks, that looks amazing! I'll do a more detailed review in a bit, but that approach of popping nodes from the stack is great.

woodruffw:
`shell: cmd` is also a possibility, and in that case the pattern will look something like:

```cmd
echo NAME=VAR >> %GITHUB_ENV%
```

PR #208, opened 2024-11-26
feat: even more precision for bash steps in github-env

ubiratansoares:
A follow-up for #197 

Here we leverage [tree-sitter-bash](https://crates.io/crates/tree-sitter-bash) rather than regexes to match usages of `$GITHUB_ENV` within a Bash script body. 

I took a simple approach, getting the full expansion for a `file_redirect` or `pipeline` and matching inner contents. Perhaps I'm missing some corner cases, but at least we have a starting point to iterate over 🙂

I also added some additional test cases (comments) and prepared the implementation to match against other types of shells in upcoming PRs.

### DIFF
--- a/src/audit/github_env.rs
+++ b/src/audit/github_env.rs
@@ -9,19 +9,16 @@ use std::ops::Deref;
 use tree_sitter::Parser;
 
 pub(crate) struct GitHubEnv {
-    // Note : interior mutability adopted since Parser::parse requires &mut self
-    bash_script_paster: RefCell<Parser>,
+    // NOTE: interior mutability used since Parser::parse requires &mut self
+    bash_parser: RefCell<Parser>,
 }
 
 audit_meta!(GitHubEnv, "github-env", "dangerous use of GITHUB_ENV");
 
 impl GitHubEnv {
-    fn evaluate_github_environment_within_bash_script(
-        &self,
-        script_body: &str,
-    ) -> anyhow::Result<bool> {
+    fn bash_runs_has_github_env_write(&self, script_body: &str) -> anyhow::Result<bool> {
         let tree = &self
-            .bash_script_paster
+            .bash_parser
             .borrow_mut()
             .parse(script_body, None)
             .context("failed to parse bash script body")?;
@@ -53,11 +50,9 @@ impl GitHubEnv {
     }
 
     fn uses_github_environment(&self, run_step_body: &str, shell: &str) -> anyhow::Result<bool> {
-        // Note : as an upcoming refinement, we should evaluate shell interpreters
-        // other than Bash
-
+        // TODO: handle `run:` bodies other than bash.
         match shell {
-            "bash" => self.evaluate_github_environment_within_bash_script(run_step_body),
+            "bash" => self.bash_runs_has_github_env_write(run_step_body),
             &_ => {
                 log::warn!(
                     "'{}' shell not supported when evaluating usage of GITHUB_ENV",
@@ -80,7 +75,7 @@ impl WorkflowAudit for GitHubEnv {
             .set_language(&bash.into())
             .context("failed to load bash parser")?;
         Ok(Self {
-            bash_script_paster: RefCell::new(parser),
+            bash_parser: RefCell::new(parser),
         })
     }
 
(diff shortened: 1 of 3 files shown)


src/audit/github_env.rs
```diff
@@ -2,28 +2,66 @@ use super::{audit_meta, WorkflowAudit};
...
 pub(crate) struct GitHubEnv;
 
 audit_meta!(GitHubEnv, "github-env", "dangerous use of GITHUB_ENV");
 
 impl GitHubEnv {
-    fn uses_github_environment(run_step_body: &str) -> bool {
-        GITHUB_ENV_WRITE_SHELL.is_match(run_step_body)
+    fn evaluate_github_environment_within_bash_script(script_body: &str) -> anyhow::Result<bool> {
+        let bash = tree_sitter_bash::LANGUAGE;
+        let mut parser = Parser::new();
```
woodruffw, line 17: Nit: creating the parser with each evaluation is probably slower than necessary; we could probably reuse `parser` by making it a property of `GitHubEnv` that gets loaded when the audit itself is loaded with `GitHubEnv::new` 🙂

src/audit/github_env.rs
```diff
@@ -47,8 +85,9 @@ impl WorkflowAudit for GitHubEnv {
             return Ok(findings);
         }
 
-        if let StepBody::Run { run, .. } = &step.deref().body {
-            if Self::uses_github_environment(run) {
+        if let StepBody::Run { run, shell, .. } = &step.deref().body {
+            let interpreter = shell.clone().unwrap_or("bash".into());
```
woodruffw, line 89: This is unfortunately not guaranteed to be the right default, since Windows runners default to `pwsh`. I think what we need to do is look at the surrounding `NormalJob::runs_on` and maybe add some kind of `RunsOn::default_shell()` helper. I'll look into that a bit.

(Nonblocker, since this is still improving on the precision of the current approach 🙂)

src/audit/github_env.rs
```diff
@@ -72,7 +111,7 @@ mod tests {
     use crate::audit::github_env::GitHubEnv;
 
     #[test]
-    fn test_shell_patterns() {
+    fn test_exploitable_bash_patterns() {
```
woodruffw, line 114: I think we can fold these two test functions into one, with `(case, expected: bool)` as the parametrization 🙂

src/audit/github_env.rs
woodruffw: Actually, it's a bit more annoying than even that: we need to determine the step's shell via the following precedence:

1. [defaults.run.shell](https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#defaultsrunshell)
2. [jobs.<job_id>.defaults.run.shell](https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#jobsjob_iddefaultsrunshell)
3. [jobs.<job_id>.steps[*].shell](https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsshell)
4. If none of the above are set, use the default shell implied by the runner from `runs-on` (which might be indeterminate in the self-hosted runner case)

ubiratansoares: @woodruffw I totally missed `pwsh` on Windows as the default, I never used Windows runners and I though that `bash` was the default (via WLS). Another day, another learning I guess 🙂

woodruffw: I forget about too 🙂 -- the changes in #213 should enable this, but I'll rebase off of your work so that it's not a blocker here.

ubiratansoares:
@woodruffw I lack the energy to work on anything else today, I'll resume the work on this PR tomorrow, addressing all your suggestions, and perhaps making the most of #213 🙂

Meanwhile, thanks again for the great review (as usual), it's a pleasure contributing to and learning from this project ❤️

woodruffw:
Thank you too! And there's no rush or urgency at all, it's all supposed to be open source fun 🙂

src/audit/github_env.rs
ubiratansoares: Done [here](https://github.com/woodruffw/zizmor/pull/208/commits/bc0e914233b4ac5209822e04f25faee2df6b573f), along with other changes!

woodruffw (approved):
LGTM, thanks a ton @ubiratansoares!

PR #227, opened 2024-12-03
feat: handle powershell in github-env audit

woodruffw:
~~This is purely experimental: I revived a copy of `tree-sitter-powershell` and got the Rust bindings working. I don't love the idea of having to maintain this, though.~~

~~Changes needed here: https://github.com/woodruffw-forks/tree-sitter-powershell/pull/1~~

Uses `tree-sitter-powershell` to match env writes, like we do with `tree-sitter-bash`.

See #197


woodruffw:
Blocked: https://github.com/airbus-cert/tree-sitter-powershell/issues/12

woodruffw:
Almost unblocked here, but not quite: https://github.com/airbus-cert/tree-sitter-powershell/issues/12

woodruffw:
Final blocker: https://github.com/airbus-cert/tree-sitter-powershell/pull/15

PR #230, opened 2024-12-04
feat: remote auditing

### LINKED ISSUE #50: Support remote auditing?
Right now, `zizmor` requires workflows to already be present on disk. But we could in theory fetch repos for users and audit from there:

```bash
# would audit example/example/.github/workflows/*.yml
zizmor gh:example/example

# would audit just `release.yml` in the specified repo
zizmor gh:example/example:release.yml
```

I have mixed feelings about doing this, since it's more convenient for users but requires me to think about temporary directories + tweaking the SARIF output.

### LINKED ISSUE #177: Refactor: switch to `camino` for path handling
`camino` is a UTF-8-safe path handling library, which would simplify things quite a bit for us.

Ref: <https://docs.rs/camino/latest/camino/>

woodruffw:
~~WIP.~~

Once finished, this will enable `zizmor foo/bar`, where `foo/bar` is a GitHub repo that `zizmor` will extract workflows to audit from.

~~(The `@`-prefix is not a firm design choice -- I might remove that.)~~

Closes #50

Closes #177

CC @miketheman as an interested party :slightly_smiling_face:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -189,6 +189,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "9ac0150caa2ae65ca5bd83f25c7de183dea78d4d366469f148435e2acfbad0da"
 
 [[package]]
+name = "camino"
+version = "1.1.9"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "8b96ec4966b5813e2c0507c1f86115c8c5abaadc3980879c3424042a02fd1ad3"
+dependencies = [
+ "serde",
+]
+
+[[package]]
 name = "cc"
 version = "1.2.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -2575,6 +2584,7 @@ dependencies = [
  "anstream",
  "anyhow",
  "assert_cmd",
+ "camino",
  "clap",
  "clap-verbosity-flag",
  "env_logger",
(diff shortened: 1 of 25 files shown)


miketheman:
Woohoo, thanks for taking this on!

> (The `@`-prefix is not a firm design choice -- I might remove that.)

I agree - especially considering that we cannot predict the future of Naming Things.

In my head, it's absolutely fine and fair to use the entire repo URL as the input - and this don't have to "teach" a new thing.

miketheman:
For completeness, an example:

```shell
zizmor https://github.com/woodruffw/zizmor
```

woodruffw:
> In my head, it's absolutely fine and fair to use the entire repo URL as the input - and this don't have to "teach" a new thing.

True -- the only reason why I'm shying away from this is because I don't want to have to teach `zizmor` how to validate URLs, versus being given a slug that it can make correct-by-construction into a GitHub URL.

For example, I don't want to get bug reports for `ssh://` URLs, or GHE deployments, or for `http://random-filehost.example.com/why-is-there-a-workflow-here.yml` :slightly_smiling_face: 

As a middle ground, I think I'll drop the `@` prefix but go with a convention that's already universal on GitHub, namely `org/repo(@ref)?`. That can be disambiguated easily from a local file by testing for presence, and leaves open changing to a full URL input in the future if/when that becomes pressing.

PR #231, opened 2024-12-05
feat: improves expression expansion when handling matrices

### LINKED ISSUE #114: Generalize matrix handling
#113 and #22 reveal the limitations of our current (primitive) approach to processing matrices and matrix expressions. We need a more general approach.

Some rough ideas:

1. We probably need a `models::Matrix` API similar to the ones for `Job`, `Step`, etc. to provide convenience APIs.
2. We probably need a one-shot API for "all possible expansions of this `matrix.foobar` expression". This could probably be done with [serde_json_path](https://docs.rs/serde_json_path/latest/serde_json_path/).

Needs further design thought; feedback & thoughts welcome!

ubiratansoares:
An attempt to #114 

There are a couple of improvements here, but at least we have a working implementation. I've added _snapshot_ tests to exercise different scenarios for `self-hosted` audit and `template-injection` as well.

See some comments below. Happy to get input here and interate over them 🙂

Closes #114


src/audit/self_hosted_runner.rs
```diff
@@ -106,21 +106,171 @@ impl WorkflowAudit for SelfHostedRunner {
...
+                  - run: echo \"hello from a self-hosted runner\"
+        "};
+
+        let auditable = parsed_workflow(workflow);
+
+        let sut = self_hosted_runner_audit();
+
+        let findings = sut.audit(&auditable).expect("cannot audit workflow");
+
+        assert_eq!(findings.len(), 1);
```
ubiratansoares, line 196: Because we are running only one audit, I'm just asserting that we find something. Happy to improve here, although constructing a `Finding` instance is not that simple. We can assert determinations with ease, though

src/models.rs
```diff
@@ -217,6 +219,146 @@ impl<'w> Iterator for Jobs<'w> {
...
+    fn expand_dimensions(
+        dimensions: &IndexMap<String, LoE<Vec<serde_yaml::Value>>>,
+    ) -> Vec<(String, String)> {
+        let normalized = dimensions
+            .iter()
+            .map(|(k, v)| (k.to_owned(), json!(v)))
+            .collect::<HashMap<_, _>>();
+
+        Matrix::expand(normalized)
+    }
```
ubiratansoares, line 320: I don't like this solution that much, but the issue here is that we have two different types to deal with (namely `IndexMap<String, serde_yaml::Value>` and `LoE<Vec<serde_yaml::Value>>>`) and my solution was just converting them to `serde_json:Value` to have a common ground when evaluating.

src/audit/template_injection.rs
```diff
@@ -220,12 +177,11 @@ impl TemplateInjection {
...
                     if let Some(Strategy { matrix, .. }) = &job.strategy {
                         let matrix_is_static = match matrix {
-                            // The matrix is statically defined, but one
-                            // or more keys might contain expressions.
-                            Some(LoE::Literal(matrix)) => self.matrix_is_static(context, matrix),
                             // The matrix is generated by an expression, meaning
                             // that it's trivially not static.
                             Some(LoE::Expr(_)) => false,
+
+                            Some(inner) => models::Matrix::new(inner).is_static(),
```
ubiratansoares, line 184: This now became straightforward, but there is an optimization opportunity here by caching the expansion somewhere, so we don't expand the same matrix twice (first when running `self-hosted`, second when running `template-injection`)

woodruffw:
Thanks for tackling this @ubiratansoares! I've been procrastinating on improving things here 😅 -- I'll do a quick pass now and then a more detailed on later today/tomorrow.

src/models.rs
```diff
@@ -217,6 +219,146 @@ impl<'w> Iterator for Jobs<'w> {
...
+                .iter()
+                .flat_map(|(key, value)| {
+                    let mut new_path = current_path.clone();
+                    new_path.push('.');
+                    new_path.push_str(key);
+                    Matrix::walk_path(value, new_path)
+                })
+                .collect(),
+        }
+    }
```
ubiratansoares, line 359: I went after the recursive implementation when I was interating, no strong opinion on recursive vs. iterative for this use case since we should not expect deeply nested matrix definitions

src/audit/template_injection.rs
woodruffw: Yeah, good callout -- one thing we could do here is have `models::Job` contain the expanded `Option<Matrix>` to cache things. But we can think about that as a follow-on 🙂

src/models.rs
```diff
@@ -217,6 +219,146 @@ impl<'w> Iterator for Jobs<'w> {
...
+                expansions
+            }
+        }
+    }
+
+    /// Checks whether some expanded path leads to an expression
+    pub(crate) fn is_static(&self) -> bool {
+        self.expand_values()
+            .iter()
+            .any(|(_, expansion)| expansion.starts_with("${{") && expansion.starts_with("}}"))
```
woodruffw, line 299: Should this be:

```suggestion
            .any(|(_, expansion)| expansion.starts_with("${{") && expansion.ends_with("}}"))
```

Otherwise I think this will always return `false`.

src/audit/self_hosted_runner.rs
woodruffw: Let's use snapshot tests for these! That'll be easier to maintain and doesn't require us to have inline workflows or construct `Finding` objects here 🙂 

(These could all be individual workflows under `test-data` (or `test-data/matrices` or similar), or they could be one big workflow. No preference from me.)

ubiratansoares: I'll do that, I had the current test written I got `insta` stuff only after syncing 🙂

src/models.rs
ubiratansoares: Fixed [here](https://github.com/woodruffw/zizmor/pull/231/commits/6a6932146096d7a686346e37be3ed2d429cba626)

src/audit/self_hosted_runner.rs
ubiratansoares: Done [here](https://github.com/woodruffw/zizmor/pull/231/commits/e28d18fc1b7c873c2079be946a87bc6f2fd16eaf), I added a sub-folder `self-hosted`, but happy to add more test cases covering `template-injection` 🙂

src/models.rs
```diff
@@ -217,6 +219,146 @@ impl<'w> Iterator for Jobs<'w> {
...
+            bail!("job is not a normal job")
+        };
+
+        let Some(strategy) = &job.strategy else {
+            bail!("job does not define a strategy")
+        };
+
+        let Some(inner) = &strategy.matrix else {
+            bail!("job does not define a matrix")
+        };
```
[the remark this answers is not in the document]
ubiratansoares: I gave a few tries on this one, tbh I could not figure out how to collapse it to a single `let` 🙈

Would `and_then` chaining be your take here?

ubiratansoares:
@woodruffw Fixed all the conflicts and addressed some comments! Impressed by the amount of stuff that happened since Friday, btw 😄

src/models.rs
```diff
@@ -224,6 +226,146 @@ impl<'w> Iterator for Jobs<'w> {
...
+                if let LoE::Literal(includes) = &matrix.include {
+                    let additional_expansions = includes
+                        .iter()
+                        .flat_map(Matrix::expand_inclusions)
+                        .collect::<Vec<_>>();
+
+                    expansions.extend(additional_expansions);
+                };
+
+                // Todo : exclude from expansions
```
ubiratansoares, line 296: @woodruffw Happy to learn about your thoughts on the general approach before adding this one!

src/audit/self_hosted_runner.rs
ubiratansoares: I went for those tests and caught a bug [here](https://github.com/woodruffw/zizmor/pull/231/commits/686b9665804331658b7315ffe8356238bbd8f841) 😄

src/models.rs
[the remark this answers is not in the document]
woodruffw: Ah yeah, I forgot this wouldn't work: `NormalJob` has a `Box<...>` in it, so we can't completely flatten these destructures (since `box_patterns` was never stabilized). Sorry for the wild goose chase here!

We can still reduce it slightly however; I'll push up what I'm thinking.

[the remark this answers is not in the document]
woodruffw: 7756e9e

src/audit/self_hosted_runner.rs
woodruffw: Nice!

woodruffw: Yeah, subfolders are totally fine by me.

src/models.rs
woodruffw: This approach looks good to me, go ahead!

ubiratansoares: Done [here](https://github.com/woodruffw/zizmor/pull/231/commits/4135ccb0a7b113b7dea90a5a0f96360fb282a7ca)

src/models.rs
woodruffw: I'm OK with recursion here -- I find this easier to follow than the imperative equivalent would be 🙂. If we end up with panics from stack overflows due to deeply nested paths we can revisit this later.

src/models.rs
```diff
@@ -224,6 +226,162 @@ impl<'w> Iterator for Jobs<'w> {
...
+    pub(crate) fn is_static(&self) -> bool {
+        let expands_to_expression = self
+            .expand_values()
+            .iter()
+            .any(|(_, expansion)| expansion.starts_with("${{") && expansion.ends_with("}}"));
+
+        !expands_to_expression
+    }
+
+    fn expand_inclusions_or_exclusions(
```
woodruffw, line 323: Nitpick: `expand_explicit_rows` would be just as descriptive here, but shorter 🙂 

```suggestion
    fn expand_explicit_rows(
```

woodruffw (approved):
Thanks @ubiratansoares, this LGTM! One nitpick but otherwise we're good to go.

src/models.rs
ubiratansoares: Done [here](https://github.com/woodruffw/zizmor/pull/231/commits/4c17106d81c7d4d4595b8c5cf61e1f9d8f1f4f02) 🚀

PR #232, opened 2024-12-06
refactor: experiment with tracing

### LINKED ISSUE #40: Track progress with `tracing` + `tracing_indicatif`
If we were to use `tracing` instead of the standard `log` APIs, we could use [`tracing-indicatif`](https://github.com/emersonford/tracing-indicatif) to produce a more responsive/meaningful progress render.

woodruffw:
~~Very WIP.~~

The basic idea here:

1. Remove `log` and `env_logger` entirely in favor of `tracing` and its crates
2. Work on making the `tracing_indicatif` spans as pretty as possible/as useful as possible
3. Evaluate if this is actually worth it?

Closes #40


woodruffw:
This is working pretty well:

demo

PR #259, opened 2024-12-09
docs: add template-injection tips

woodruffw:
See convo in https://github.com/python/cpython/pull/127749#discussion_r1875081281.

CC @sethmlarson @alexwaygood @hugovk

### DIFF
--- a/docs/audits.md
+++ b/docs/audits.md
@@ -462,6 +462,23 @@ from the expanded template.
 This avoids the vulnerability, since variable expansion is subject to normal
 shell quoting/expansion rules.
 
+!!! tip
+
+    To fully remediate the vulnerability, you **should not** use
+    `${{ env.VARNAME }}`, since that is still a template expansion.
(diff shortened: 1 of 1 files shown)


woodruffw:
> What are your thoughts on adding a linkable heading that you can put directly into Zizmor's error's, so users are able to see these tips right after the scanner runs and know what to do to remediate?

It's already there! The finding IDs should be clickable in your terminal, and will take you directly to each audit's docs.

(I'm open to ideas for how to better surface this -- OSC 8 links are great for hyperlinking but they don't make it obvious when a thing is a link 😅)

hugovk:
> It's already there! The finding IDs should be clickable in your terminal, and will take you directly to each audit's docs.

I found this by mistake, it was a nice surprise and subsequently useful :) 🚀

hugovk (approved):
Thank you!

AlexWaygood:
Thanks! :D

PR #274, opened 2024-12-10
feat: evaluates a matrix expansion only once

ubiratansoares:
A follow-up from https://github.com/woodruffw/zizmor/pull/231

Just realised that we can expand the Matrix when instantiating, hence storing the expanded values as a field and avoiding a new expansion every time we need to evaluate the same matrix . 😄


woodruffw (approved):
LGTM!

Issue #283, opened 2024-12-12
[BUG]: 0.9.0 crashes with "job does not define a strategy or interior matrix"


Nateowami:
### Pre-submission checks

- [X] I am **not** filing a feature request. These should be filed via the feature request form instead.
- [X] I have looked through the [open issues](https://github.com/woodruffw/zizmor/issues?q=is%3Aissue+is%3Aopen+) for a duplicate report.

### Expected behavior

Given this workflow file, zizmor 0.8.0 handles it fine:
``` yaml
name: Workflow

on:
  workflow_call:
    inputs:
      os:
        type: string

jobs:
  deploy:
    runs-on: ${{inputs.os}}
    steps:
      - run: echo "Hello world"
```

### Actual behavior

zizmor 0.9.0 crashes with this message:
>self-hosted-runner failed on release.yml
>
>Caused by:
>    job does not define a strategy or interior matrix

### Reproduction steps

1. Run `zizmor --version` to verify version is `0.9.0`
2. Save the YAML file above as `release.yml`
3. Run `zizmor release.yml`
4. Observe the error

ubiratansoares:
My fault here 🙈 

Thanks for catching that @Nateowami. I have a tentative fix [here](https://github.com/woodruffw/zizmor/pull/284), cc @woodruffw

woodruffw:
Thanks @Nateowami for the report and @ubiratansoares for the fast patch! Looking now.

woodruffw:
Fix LGTM. We'll have a patch release for this shortly.

woodruffw:
0.9.1 cut and should be live shortly. Thanks again both!

Nateowami:
Can confirm this fixes the issue. Thanks for the speedy fix!

PR #284, opened 2024-12-12
fix: dont crash when an expression does not expand a matrix

ubiratansoares:
A tentative fix for #283 

Added both the repro Workflow and the snapshot test to get us covered!


woodruffw (approved):

woodruffw:
LGTM, thanks for the fast fix @ubiratansoares! I'll get this out in `0.9.1` now.

Issue #261, opened 2024-12-09
New audit: cache poisoning


woodruffw:
Cache poisoning is difficult/impossible for us to detect in the general case, since it's a runtime aspect.

However, we *can* flag certain patterns that we know are likely to result in exploitable cache poisoning, e.g.:

1. The workflow has a release/tag trigger, indicating that it's a release process
2. The workflow has one or more steps that include caching, such as `setup-python` with the `cache:` key or `setup-node` with similar. There are probably many of these, and we need to figure them out/enumerate them.

woodruffw:
> @woodruffw If you don't have plans to take this one, I'd love to give it a try 🙂

Go for it! This one has some interesting edge cases (e.g. there might be a top-level release trigger but then an `if:` that limits it on a job with a cache), but we can figure those out as you iterate 🙂 

Off the top of my head, here's an initial set of actions we want to catch here:

* `actions/setup-python`: has `cache:`
* `actions/setup-node`: `has `cache:`
* `ruby/setup-ruby`: has `bundler-cache:`
* `actions-rust-lang/setup-rust-toolchain`: has `cache:` plus some others: <https://github.com/actions-rust-lang/setup-rust-toolchain#inputs>
    * `cache` appears to be on by default for this one!
* any use of `Swatinem/rust-cache` at all (in a release workflow)?
* any use of `actions/cache` or `actions/cache/restore` at all (in a release workflow)?

ubiratansoares:
Thanks for sharing your thoughts. Those considerations definitely help 🚀 

I'm particular interested having some upfront chat around the set of Actions we want to check. 

My first take would be having the Actions provided by Github as the hardcoded ones to go for. These are not going anywhere (hopefully 😂), and they look like a good place to start.

Those Actions would be : 

- [actions/setup-java](https://github.com/actions/setup-java/blob/main/action.yml#L56)
- [actions/setup-go](https://github.com/actions/setup-go/blob/main/action.yml#L15)
- [actions/setup-python](https://github.com/actions/setup-python/blob/main/action.yml#L10)
- [actions/setup-dotnet](https://github.com/actions/setup-dotnet/blob/main/action.yml#L20)
- [actions/setup-node](https://github.com/actions/setup-node/blob/main/action.yml#L24)
- [actions/cache/restore](https://github.com/actions/cache/blob/main/restore/action.yml)

However, I really like the idea of dynamically retrieve a list of Actions we want to inspect for cacheability, using a schema that exposes at least 

[...]

PR #294, opened 2024-12-13
feat: initial cache-poisoning audit

ubiratansoares:
A first iteration on #261 

There is a lot to unpack in this PR already, so I added several inline comments, and I'll go for docs as soon as we are good with the overal design 🙂

First consideration : for now, I'm matching as triggers for this audit only `release` and `push` (with tags filters).
I'm aware that more scenarios exist : I myself use another pattern (on top of `workflow_dispatch`), happy to get input and iterate, or eventually tackle them as a follow-up. I miss a check for the default branch of the repo, but since it requires querying the Github API, I've skipped it for now.

Second consideration: because of the multitude of caching-aware Actions that exist (and how they behave), and because we should aim to consume a static API at some point in the future, I've modelled a `CacheAwareAction` struct, which is the very core of this PR.

In general, I tried to capture with this data holder:

- that some actions will run caching behavior by default, both for saving and restoring (eg `actions/setup-go`)
- that some actions have an input that allows either with opt-in or opt-out from **_restoring cache_** (eg `actions/cache`)
- that some actions define opt-in not by specifying a boolean input, but actually a set of values (eg `actions/setup-node`)
- eventually the user opts-in for caching on top of an expression: there is not much to do, but I keep track of that

There is a static list of `CacheAwareAction`, covering all Actions provided by Github and a few others. Happy to add as many as we want to have 🔥

Third consideration: I defined a strategy for this audit [on top of a chat I had with Adnan](https://github.com/AdnaneKhan/ActionsCacheBlasting/issues/3), focusing on the unzip part of cache-poisoning attacks. This means I'm not considering `actions/cache/save` in the Actions we inspect for. 

Most of the Actions I've listed bring both saving/restoring out-of-the-box, but some of them define a `lookup-only` input (as per `actions/cache` and `actions/cache/restore` do), that only queries the Github API for cache entries but [do not download any caches](https://github.com/actions/cache/blob/main/restore/action.yml). Maybe this is incomplete/insufficient, and I'd love to get inputs here! 

Last but not least : I added several examples of vulnerable Workflows and snapshot test all of them.

Most likely I missed other things, but hopefully this PR sesa a place to get started 🙂


src/audit/cache_poisoning.rs
```diff
@@ -0,0 +1,292 @@
...
+        let (yaml_key, annotation) = match cache_usage {
+            CacheUsage::DefaultActionBehaviour => ("uses", "cache enabled by default here"),
+            CacheUsage::DirectOptIn => ("with", "opt-in for caching here"),
+            CacheUsage::ConditionalOptIn => ("with", "opt-in for caching might happen here"),
+        };
+
+        findings.push(
+            Self::finding()
+                .confidence(Confidence::Low)
+                .severity(Severity::High)
```
ubiratansoares, line 275: Not sure about these two!

woodruffw:
Thanks @ubiratansoares! I'll try and do an initial pass on this today or tomorrow.

src/audit/cache_poisoning.rs
woodruffw: These look good to me for now 🙂

src/audit/cache_poisoning.rs
```diff
@@ -0,0 +1,292 @@
...
+    /// The type of this toggle (Boolean, MultiValue)
+    cache_control_input_toggle: CacheControlToggle,
+    /// Whether this Action adopts caching as the default behavior
+    caching_by_default: bool,
+}
+
+/// The list of know cache-aware actions
+/// In the future we can easily retrieve this list from the static API,
+/// since it should be easily serializable
+static KNOWN_CACHE_AWARE_ACTIONS: LazyLock<Vec<CacheAwareAction>> = LazyLock::new(|| {
```
woodruffw, line 40: Nitpick: you _should_ be able to remove the `LazyLock` here and declare `const KNOWN_CACHE_AWARE_ACTIONS` instead, as a `&[CacheAwareAction]` rather than a `Vec<...>` 🙂

src/audit/cache_poisoning.rs
```diff
@@ -0,0 +1,292 @@
...
+#[derive(PartialEq)]
+enum CacheControlToggle {
+    Boolean,
+    MultiValue,
+}
+
+/// The general schema for a cache-aware actions
+struct CacheAwareAction {
+    /// The owner/repo part within the Action full coordinate
+    reference: &'static str,
```
woodruffw, line 26: Two additional considerations here:

* We should make sure to support `owner/repo/action/path`, i.e. subpaths 🙂 
* Action defaults will probably change over time, so in practice we probably also need the ability to indicate versions here

Put together, and I _think_ it might make sense to re-use the `RepositoryUses` model here, since that'll give us all the (optional) additional state we need. I've also added `RepositoryUses::matches` in #295, which should make fuzzy-matching of different `uses:` variants slightly more convenient.

src/audit/cache_poisoning.rs
```diff
@@ -0,0 +1,292 @@
...
+    /// The name of the input the toggle caching behavior
+    cache_control_input_name: &'static str,
+    /// Whether this input defines opt-in behavior
+    cache_control_input_defines_opt_in: bool,
```
woodruffw, line 30: Thoughts on collapsing these into an enumerated state? e.g.:

```
cache_control: CacheControl::OptIn("field-name")
cache_control: CacheControl::OptOut("field-name")
```

(and maybe `cache_control_input_toggle` too, although I haven't thought too hard on that one.)

src/audit/cache_poisoning.rs
```diff
@@ -0,0 +1,292 @@
...
+    {
+        Ok(Self)
+    }
+
+    fn audit_step<'w>(&self, step: &Step<'w>) -> anyhow::Result<Vec<Finding<'w>>> {
+        let mut findings = vec![];
+
+        let trigger = &step.workflow().on;
+
+        if !self.trigger_used_when_publishing_artifacts(trigger) {
```
woodruffw, line 254: This looks good for now, thanks!

Thinking out loud: another future proxy we could use for "is this a publishing workflow" is whether or not the workflow contains any well-known publishing actions, e.g. `gh-action-pypi-publish`. 

Not something to do with this iteration, but worth thinking about 🙂

src/audit/cache_poisoning.rs
ubiratansoares: There is definitely some room to escape the primitives obsession here 😄

src/audit/cache_poisoning.rs
ubiratansoares: Thanks for sharing this input!

The way I solved subpaths was by adding two different `CacheAwareAction` instances in the list, particularly in the case of `actions/cache` and `actions/cache/restore`. 

This was a simple take ofc, and the only advantage I see for sticking with it is the possibility of two Actions in two different subpaths actually defining different sets of inputs/outputs, meaning that we would need to handle them separatedly anyway.

That's is not case for `actions/cache` and `actions/cache/restore`, though. They share some semantics on their APIs (the same input controls opt-out from restoring in both of them), but who knows what wait us in the wild 😄

I'm happy to go for `RepositoryUses`, I don't have a strong opinion here. Let me know what would be your call 🙂

src/audit/cache_poisoning.rs
```diff
@@ -0,0 +1,292 @@
...
+    cache_control_input_toggle: CacheControlToggle,
+    /// Whether this Action adopts caching as the default behavior
+    caching_by_default: bool,
+}
+
+/// The list of know cache-aware actions
+/// In the future we can easily retrieve this list from the static API,
+/// since it should be easily serializable
+static KNOWN_CACHE_AWARE_ACTIONS: LazyLock<Vec<CacheAwareAction>> = LazyLock::new(|| {
+    vec![
```
woodruffw, line 41: It might be worth adding `PyO3/maturin-action` to this list as well, since `sccache: true` enables caching via `sccache` + GHA :slightly_smiling_face: 

(Our own `pypi.yml` uses this, which is not ideal :sweat_smile:)

src/audit/cache_poisoning.rs
ubiratansoares: Done [here](https://github.com/woodruffw/zizmor/pull/294/commits/85475ab269837f423dceab783fac81f838a20cee)

src/audit/cache_poisoning.rs
ubiratansoares: Also improved [here](https://github.com/woodruffw/zizmor/pull/294/commits/85475ab269837f423dceab783fac81f838a20cee). Reads much better now, btw!

src/audit/cache_poisoning.rs
ubiratansoares: Added [here](https://github.com/woodruffw/zizmor/pull/294/commits/accbdbdb447e0788b180e85f1deb3b5ff4a286c7)

ubiratansoares:
@woodruffw Addressed your suggestions, let me know whether I missed something! Docs to follow 🙂

woodruffw:
Thanks! I think that's everything -- just undraft and give me a ping when you're ready for a final review!

ubiratansoares:
@woodruffw Added some docs [here](https://github.com/woodruffw/zizmor/pull/294/commits/f7071625108b19b1665f4c96e0b61029c080063e), let me know what you think :) 

Some follow-ups from this PR, so we can keep it size manageable for the review: 

- Add more Actions in the list of well-know cache-aware Actions 😄
- Evaluate the presence of well-know publish-like Actions (e.g. `gh-action-pypi-publish`) in addition to the trigger-based evaluation
- Improve evaluation for `push` trigger with well-know branch paths (eg, `release-*` branches) ?

docs/audits.md
```diff
@@ -693,6 +693,52 @@ like `pull_request_target`.
 
 If you need to pass state between steps, consider using `GITHUB_OUTPUT` instead.
 
+## `cache-poisoning`
+
+| Type     | Examples                | Introduced in  | Works offline  | Enabled by default |
+|----------|-------------------------|----------------|----------------|--------------------|
+| Workflow  | [cache-poisoning.yml]   | v1.0.0         | ✅             | ✅                 |
```
ubiratansoares, line 700: Went for `1.0.0` since you added the milestone 😄

ubiratansoares:
Added [this additional commit](https://github.com/woodruffw/zizmor/pull/294/commits/6866b1891840246dc3712ac29df4d20585955973) to highlight another (potential) follow-up : we should evaluate `step.if` to avoid some false positives, since the user can avoid cache restoration by disabling the a cache-aware step

docs/audits.md
woodruffw: Good call, although I _think_ I'm going to do at least one more minor release before going to 1.0 😅 -- so let's do this as `v0.10.0`, which I'll plan to have out pretty soon:

```suggestion
| Workflow  | [cache-poisoning.yml]   | v0.10.0         | ✅             | ✅                 |
```

src/audit/cache_poisoning.rs
```diff
@@ -0,0 +1,282 @@
...
+                        None
+                    }
+                }
+                Some(_) => Some(CacheUsage::ConditionalOptIn),
+            },
+        }
+    }
+
+    fn evaluate_cache_usage(&self, uses: &str, env: &Env) -> Option<CacheUsage> {
+        let (reference, _) = uses.split_once("@")?;
```
woodruffw, line 185: Nitpick: I think ad-hoc parsing here is going to be brittle, e.g. if a user does `uses: ruby/setup-ruby` (with no `@`-pin), which we'd currently end up silently skipping here due to the `split_once` usage.

Instead, maybe we should parse with `Uses::from_common` here?

src/audit/cache_poisoning.rs
```diff
@@ -0,0 +1,282 @@
...
+
+        let known_action = KNOWN_CACHE_AWARE_ACTIONS.iter().find(|action| {
+            let Uses::Repository(uses) = action.uses else {
+                return false;
+            };
+
+            uses.matches(reference)
```
woodruffw, line 192: I might be mis-reading this (jetlag 😅), but I _think_ this is the wrong way around: the API is `Uses::matches(template)`, and in this case the known cache-aware action is our template (e.g. `ruby/setup-ruby` which should match `ruby/setup-ruby`, `ruby/setup-ruby@v1`, etc.).

src/audit/cache_poisoning.rs
ubiratansoares: Definetely yes! Check my comment below 🙂

src/audit/cache_poisoning.rs
ubiratansoares: You are totally correct! I missed the opportunity to make the most of `RepositoryUses::matches` here, and only because I did not test a case like `ruby/setup-ruby`  in the snapshots I did not spot the issue before. Awesome catch, even under jetlag 🙂

Fixed that [here](https://github.com/woodruffw/zizmor/pull/294/commits/c935683be64adc72377971026e6764e410bd460e), along with version that introduces this new audit

src/models.rs
```diff
@@ -589,6 +589,10 @@ impl RepositoryUses<'_> {
             _ => None,
         }
     }
+
+    pub(crate) fn as_template(&self) -> String {
+        format!("{}/{}", self.owner, self.repo)
+    }
```
ubiratansoares, line 595: Created this new method as a helper, since I need to get a template from what we have in well-know list.

woodruffw:
Thanks! I'll give this a final pass tomorrow, but this is looking great to me!

Let's track the follow-ups in https://github.com/woodruffw/zizmor/pull/294#issuecomment-2547938847 with a new issue 🙂

src/models.rs
woodruffw: I don't love that we have to re-build the slug here, but that's my fault for building a bad abstraction here 😅 -- I'll try and clean this up as part of the final review.

woodruffw:
[`ee7d227` (#294)](https://github.com/woodruffw/zizmor/pull/294/commits/ee7d227ae7b7cf225e4dd7bdca410084935df5d4) tweaks the audit slightly so that it catches our own problematic `pypi.yml` workflow. I need to think about how best to fix that, plus how best to ensure this audit sees our fx.

woodruffw (approved):
LGTM, great work @ubiratansoares!

PR #297, opened 2024-12-14
fix: template-injection: consider runner.tool_cache safe

woodruffw:
Adds another always-safe context to the safe list.

### DIFF
--- a/src/audit/template_injection.rs
+++ b/src/audit/template_injection.rs
@@ -70,6 +70,8 @@ const SAFE_CONTEXTS: &[&str] = &[
     "runner.os",
     // GitHub Actions temporary directory, value controlled by the runner itself.
     "runner.temp",
+    // GitHub Actions cached tool directory, value controlled by the runner itself.
+    "runner.tool_cache",
 ];
 
 impl TemplateInjection {


GaryJones:
Is this still working as intended? When I run zizmor with `--pedantic`, I get:

```
help[template-injection]: code injection via template expansion
  --> ./.github/workflows/integration.yml:68:39
   |
68 |         run: echo "::add-matcher::${{ runner.tool_cache }}/php.json"
   |         --- this run block            ^^^^^^^^^^^^^^^^^ may expand into attacker-controllable code
   |
   ```
   
 Without `--pendatic`, there are no findings.

Either `runner.tool_cache` is safe, or it's not.

woodruffw:
> Is this still working as intended? When I run zizmor with --pedantic, I get:

Yes, please see the persona docs: https://docs.zizmor.sh/usage/#using-personas

Specifically, the pedantic persona means "include code smells, even if not likely exploitable." zizmor considers all template expansions to be code smells, and puts likely non-exploitable ones under the pedantic persona so that they don't show up by default.

More information on that is documented under the audit as well:

https://docs.zizmor.sh/audits/#template-injection

(See the first tip callout in that section for context)

PR #298, opened 2024-12-14
fix: uses context when evaluating static values from matrix expansions

ubiratansoares:
A tentative fix for the issue describe [here](https://github.com/woodruffw/zizmor/issues/22#issuecomment-2543128489)

We change the existing `models::Matrix` API for checking static expansions (and used know only by the `template-injection` audit) to actually take a given context (i.e, an expansion path) when evaluating.


woodruffw:
Thanks @ubiratansoares! I'll review this tonight.

woodruffw (approved):
LGTM! I think it'd be good to minimize the reproduction test case here slightly (to keep things small) but I'll look at that tomorrow.

ubiratansoares:
@woodruffw Done [here](https://github.com/woodruffw/zizmor/pull/298/commits/114a4ab63826cfc1af2117a25ab3d936aed9d926). Also moved the reproducer under `test-data/template-injection` (I missed that somehow 🙈)

woodruffw (approved):
Thanks!

PR #319, opened 2024-12-17
feat: add zizmor Docker image

### LINKED ISSUE #135: Feature: Add a docker image
### Pre-submission checks

- [X] I am **not** a bug (crash, false positive/negative, etc). These must be filed via the bug report template.
- [X] I have looked through the open issues for a duplicate request.

Usability

I'd like a docker image containing `zizmor` to be built and deployed regularly, and available on docker hub or a similar registry.

### Additional context

_No response_

trallard:
Closes #135

I am opening a WIP PR since I have some outstanding questions.

## Notes 📝

- I had originally suggested basing this on the `uv` Docker image. However, I think that might be overkill for a "vanilla" image for Zizmor, so I used the Rust image instead (`bookworm-slim,` which also helps keep the image quite small). That way, we provide the minimal required amount and leave it to users to build off as they need.

## Questions/needs input 💬 

- I set the entry point as `ENTRYPOINT ["/bin/bash"]`, or would it be best to point it to Zizmor directly? ✅ 
- For building and pushing the images:
  - I generally favour pushing images as part of my release processes, so I plan to add a reusable workflow that can be called from the `release.yml` workflow. ✅ 
  - I think it would be suitable to build both `linux/amd64` and `linux/arm64` images. 
- Also, it seems suitable to add a Docker entry to the zizmor docs (similar to https://docs.astral.sh/uv/guides/integration/docker/)

TODO for @woodruffw:

- Create a `docker` environment for the build and publish Docker job
- Add relevant credentials to authenticate to GHCR


woodruffw:
Thanks @trallard, I appreciate you opening this!

> * However, I think that might be overkill for a "vanilla" image for Zizmor, so I used the Rust image instead (`bookworm-slim,` which also helps keep the image quite small). That way, we provide the minimal required amount and leave it to users to build off as they need.

Good call IMO 🙂 -- however, I wonder if we could keep the PyPI installation (i.e. via `uv tool install` or `pip install` rather than rebuilding with `cargo install`. I think this would keep our CI fast *and* simplify the arm64/amd64 image builds (since we already have native builds for both). WDYT?

> * I set the entry point as `ENTRYPOINT ["/bin/bash"]`, or would it be best to point it to Zizmor directly?

Pointing to `zizmor` directly makes sense to me!

> * I generally favour pushing images as part of my release processes, so I plan to add a reusable workflow that can be called from the `release.yml` workflow.
> * I think it would be suitable to build both `linux/amd64` and `linux/arm64` images.

SGTM -- either a reusable workflow or a new `docker` job within the existing workflow would be fine by me. Dealer's choice 🙂 

> * Also, it seems suitable to add a Docker entry to the zizmor docs (similar to [docs.astral.sh/uv/guides/integration/docker](https://docs.astral.sh/uv/guides/integration/docker/))

Yep! Ideally this PR would include those doc changes, although they could also be a follow-up.

trallard:
Yay! Thanks for the quick follow up. 

These all seem good so I will work the rest of the things and can add some docs in this PR, then I can iterate if needed. 

> Good call IMO 🙂 -- however, I wonder if we could keep the PyPI installation (i.e. via uv tool install or pip install rather than rebuilding with cargo install. I think this would keep our CI fast and simplify the arm64/amd64 image builds (since we already have native builds for both). WDYT?

I can certainly do that. Should be easy enough to swap.

woodruffw:
Hey @trallard! Do you intend to pick this up sometime soon? No problem at all if not, but if you think it's gonna be >1 month I'm happy to build on top of what you've got here so I can land this soon-ish 🙂

trallard:
Hey there I took some time off but should be able to pick this in the next week or so.

tyler36:
Will this be an official docker image? Are you also planning to add this image Docker Hub?

woodruffw:
> Will this be an official docker image?

It'll be official in the sense that it comes from this repo, so as official as the `cargo install` and `pip install` sources. Like the other sources, you can expect it to receive updates on the same cadence/schedule as the main repository releases.

> Are you also planning to add this image Docker Hub?

It'll be available on GHCR most likely, since GHCR is easy to publish to with GitHub Actions.

trallard:
Now that I have finally caught up with my massive backlog, I can wrap this up so I am taking this out of `Draft` mode.

To test locally (i.e., build the Docker image), one can run the following command with a valid zizmor version:

```
docker build \
    --build-arg ZIZMOR_VERSION=1.3.0 \
    -t zizmor:latest .
```

Since I ended up using `uv` to install zizmor it seemed that the easiest way to add a GH workflow was to add it at the end of the `pypi` workflow to ensure the release exists and we can install zizmor properly.

.github/workflows/pypi.yml
```diff
@@ -176,3 +176,50 @@ jobs:
...
+        uses: docker/build-push-action@ca877d9245402d1537745e0e356eab47c3520991 # v6
+        with:
+          context: .
+          push: true
+          tags: ${{ steps.docker-metadata.outputs.tags }}
+          build-args: |
+            ZIZMOR_VERSION=${{ github.ref_name }}
+          platforms: linux/amd64,linux/arm64
+          # Build provenance attestations
+          provenance: true
```
trallard, line 225: I opted to add `provenance` attestations, but we can also add `sbom` ones or none (your choice).

woodruffw: Thanks! I'm good with `provenance` ones to start; if someone wants `sbom` they can file an issue and let us know 🙂

.github/workflows/pypi.yml
```diff
@@ -176,3 +176,50 @@ jobs:
...
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@6524bf65af31da8d45b59e8c27de4bd072b392f5 # v3
+      - name: Extract Docker metadata
+        id: docker-metadata
+        uses: docker/metadata-action@369eb591f429131d6889c46b94e711f089e6ca96 # v5
+        with:
+          images: "${{ env.ZIZMOR_IMAGE }}"
+          tags: |
+            # use the release tag for the image tag too
+            type=ref, event=tag
```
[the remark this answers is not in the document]
woodruffw: Same release tag makes sense to me!

.github/workflows/pypi.yml
```diff
@@ -176,3 +176,50 @@ jobs:
...
+  docker-build:
+    name: Build and publish Docker image
+    runs-on: ubuntu-latest
+    env:
+      ZIZMOR_IMAGE: ghcr.io/woodruffw/zizmor
+    environment:
+      name: docker
+    if: ${{ startsWith(github.ref, 'refs/tags/') || github.event_name == 'workflow_dispatch' }}
+    # Only run if the release job was successfully completed
+    needs: [release]
```
woodruffw, line 189: There might be a bit of a race condition with this, sadly: PyPI's propagation can take a few minutes IME, so running immediately after a `release` might fail due to the index not listing the release yet.

I think there are two (easy-ish) options here: 

1. Add some kind of blocking "wait for PyPI" job so that the flow goes `release -> wait-for-pypi -> docker-build` instead
2. Move this to a separate `docker-publish.yml` or similar that can be triggered separately, e.g. by a `workflow_dispatch` with a `tag:` input

I'd be OK with either of these but a slight preference for (2), especially since we can give it a `cron` trigger that'll do the update automatically 🙂

woodruffw (commented):
Thanks @trallard! I _think_ I'd prefer having the Docker build be in its own workflow, for reasons in the review comments below. But I'm curious to hear what you think 🙂

Dockerfile
woodruffw, line 1: This looks great to me, thanks!

Optional for the initial version, but if I'm thinking about this right: I think we could squish these images even further by using [`uv`'s own Docker image](https://docs.astral.sh/uv/guides/integration/docker/#using-installed-tools). But I'm completely fine to kick that down the road a bit 🙂

trallard: We could, maybe, but as is, the current image is only `163MB` and the build does not take a long time. I can make some comparisons as it is barely any difference if you prefer.

.github/workflows/pypi.yml
trallard: I was initially going for 2-ish, but then I realized the trigger would need to be other than a release tag so that this did not end up running _before_ the PyPI release was ready, so I moved this inside the PyPI workflow, though I completely forgot about the potential delays on the PyPI side.

Let me move this out into a separate workflow as in 2) and we can sort out the trigger event (`cron` or something else).

trallard: Another alternative could be using the built wheels (already uploaded to the artefacts) to install zizmor (thus avoiding any potential delays on the PyPI side of things) so all the workflows would trigger from the same release tag event.

Dockerfile
woodruffw: Nope, I'm OK with leaving it as is -- that was mostly just me thinking out loud. The current size seems pretty reasonable!

tyler36:
Thank you for working on this. I'm excited to try it out!

.github/workflows/pypi.yml
woodruffw: That could work, although in that case we'd need to reimplement the wheel/tag matching, right? I think a separate workflow + trigger is good for now 🙂

PR #321, opened 2024-12-17
feat: add support for .zizmor.yml files

### LINKED ISSUE #320: Feature: More zizmor configuration locations
### Pre-submission checks

- [X] I am **not** reporting a bug (crash, false positive/negative, etc). These must be filed via the bug report template.
- [X] I have looked through the open issues for a duplicate request.

I just stumbled on this package the other day and really appreciate the work on this! The main problem I have is that the configuration location `zizmor.yml` is a bit too present for my liking. 

I would like to see a dotfile `.zizmor.yml` adopted by default (following other linting/checkers like `.flake8`, `.pre-commit-config.yaml`, `.yamllint.yml`, etc.) to make it easy to ignore this file in my build system (`ignore= ".*"`).

### Additional context

Given that this checker looks specifically at GitHub Workflows, I don't see any value in adding the configurations to a language-specific build file (namely `pyproject.toml`). Being able to drop this information when building a package is important.

Zeitsperre:
Fixes #320

Adds support for configuration files named `.zizmor.yml` as well as slightly simplifies the file-searching logic.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -27,8 +27,9 @@ configuration files in the following order of precedence:
    explicitly, the config file does *not* need to be named `zizmor.yml`.
 1. `${CWD}/.github/zizmor.yml`
 1. `${CWD}/zizmor.yml`
+1. `${CWD}/.zizmor.yml`
 
-For the last two discovery methods, `${CWD}` is the current working directory,
+For the folder-tree-based discovery methods, `${CWD}` is the current working directory,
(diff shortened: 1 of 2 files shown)


woodruffw:
Closing for now per #320/#322.


(For anybody else who hits this: the recommended way to use a non-default config file is to pass it with `--config`, e.g. `--config `.zizmor.yml`.)

PR #330, opened 2024-12-19
feat: Fix Sarif schema and add rules to Sarif files 

fcasal:
This PR

- fixes the schema used in the sarif files
- adds a rules section with the documentation URL 
- replaces the finding identifier with the finding description in the result's message 

When viewed with the Sarif Explorer VSCode extension,
Before
image

After 
image


woodruffw (approved):
Thanks @fcasal!

Issue #334, opened 2024-12-19
cache-poisoning: follow-up improvements


woodruffw:
- [x] Evaluate the presence of well-know publish-like Actions (e.g. `gh-action-pypi-publish`) in addition to the trigger-based evaluation #338 
- [ ] Improve evaluation for `push` trigger with well-know branch paths (eg, `release-*` branches) ?
- [x] Add more Actions in the list of well-know cache-aware Actions 😄

_Originally posted by @ubiratansoares in https://github.com/woodruffw/zizmor/issues/294#issuecomment-2547938847_

ubiratansoares:
Investigating some cache-aware Actions I learned that [Mozilla-Actions/sccache-action](https://github.com/Mozilla-Actions/sccache-action) is an example of an Action that has no opt-in/opt-out controls and always cache/restores by default.

We may have more of these, which means that some tweaking on `CacheAwareAction` is needed to reflect the absence of control input / control values

ubiratansoares:
@woodruffw When implementating #345 I found (or remembered?) an interesting thing : it seems that Github URLs actually are not case-senstive. 👀

There are plenty issues / discussions / SO posts [like this one](https://github.com/github/docs/issues/32838), and eventually it's quite easy to check that too: for example, these HTTP GETs download the same tarball ([actions/upload-artifact](https:/github.com/repos/actions/upload-artifact)):

- https://api.github.com/repos/actions/upload-artifact/tarball/6f51ac03b9356f520e9adb1b1b7802705f340c2b
- https://api.github.com/repos/Actions/upload-artifact/tarball/6f51ac03b9356f520e9adb1b1b7802705f340c2b
- https://api.github.com/repos/actions/Upload-Artifact/tarball/6f51ac03b9356f520e9adb1b1b7802705f340c2b

and the same is also true for repositories : 

- https://github.com/PyO3/maturin-action
- https://github.com/pyo3/maturin-action

The issue that follows is that docs like the ones we have for [Mozilla-Actions/sccache-action](https://github.com/Mozilla-Actions/sccache-action) may promote the lowercase template of an Action as provided examples; those surely will be copied and pasted to Workflows in the wild, and the current implementation of `RepositoryUses::matches`  won't be able to match those templates.
 
Ofc this affects mostly `cache-poisoning` for now, but I _think_ the issue happens to all use-cases in which we need to match an Action template. 

[...]

PR #338, opened 2024-12-20
feat: detects well-known publisher actions in cache-audit

ubiratansoares:
A first follow-up from #294. also contributing to #334 

We now have a list of well-know publisher Actions (2 of them so far 😄) and the audit inspects all the steps at once alongside with the previous trigger evaluation to determine whether the Workflow/Job actually builds artifacts for publishing.

I also introduced an `PublishingArtifactsScenario` enum to easilty match the situation we are dealing with and signal properly the non-primary locations in `FindingBuilders`.

### DIFF
--- a/src/audit/cache_poisoning.rs
+++ b/src/audit/cache_poisoning.rs
@@ -2,7 +2,6 @@ use crate::audit::{audit_meta, WorkflowAudit};
 use crate::finding::{Confidence, Finding, Severity};
 use crate::models::{Job, Step, Steps, Uses};
 use crate::state::AuditState;
-use anyhow::{anyhow, bail};
 use github_actions_models::common::expr::ExplicitExpr;
 use github_actions_models::common::Env;
 use github_actions_models::workflow::event::{BareEvent, OptionalBody};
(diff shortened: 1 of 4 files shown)


src/audit/cache_poisoning.rs
```diff
@@ -110,13 +111,27 @@ static KNOWN_CACHE_AWARE_ACTIONS: LazyLock<Vec<CacheAwareAction>> = LazyLock::ne
     ]
 });
 
+/// A list of well-know publisher actions
+/// In the future we can retrieve this list from the static API
+static KNOWN_PUBLISHER_ACTIONS: LazyLock<Vec<Uses>> = LazyLock::new(|| {
+    vec![
+        Uses::from_step("pypa/gh-action-pypi-publish").unwrap(),
+        Uses::from_step("softprops/action-gh-release").unwrap(),
```
woodruffw, line 119: Let's add the Ruby one too:

```suggestion
        Uses::from_step("softprops/action-gh-release").unwrap(),
        Uses::from_step("rubygems/release-gem").unwrap(),
```

src/audit/cache_poisoning.rs
```diff
@@ -218,31 +265,18 @@ impl CachePoisoning {
...
-
-        if !self.trigger_used_when_publishing_artifacts(trigger) {
-            return Ok(findings);
-        }
 
+    fn uses_cache_aware_step<'w>(
+        &self,
+        step: &Step<'w>,
+        scenario: &PublishingArtifactsScenario<'w>,
+    ) -> anyhow::Result<Finding<'w>> {
```
woodruffw, line 273: It looks like we never use the `Err(...)` variants from this, so maybe this could be `Option<Finding<'w>>` instead?

woodruffw (commented):
Nice, thanks @ubiratansoares! Two small comments.

src/audit/cache_poisoning.rs
ubiratansoares: Totally agree, let me fix that!

ubiratansoares: Done [here](https://github.com/woodruffw/zizmor/pull/338/commits/7c203c8974696f834f800f2a8c7f146df1bf3cde)

woodruffw (approved):
LGTM, nice work!

PR #341, opened 2024-12-20
feat: adds more publish-aware actions to cache-audit

ubiratansoares:
Another contribution to #334 

I investigated the Github Marketplace along with public Github orgs out there to improve a a bit the list of publisher-like Actions. Added Actions from Docker, GCP, AWS, Azure and a few popular ones used to publish binaries to Github releases.

Should we document this in the official docs, maybe?


woodruffw:
> Should we document this in the official docs, maybe?

Makes sense to me, although I'm not sure where the best place for it to live is -- any thoughts?

woodruffw (approved):
LGTM, thanks @ubiratansoares!

PR #342, opened 2024-12-21
Updates astral-sh/setup-uv action to v5

colindean:
[Caught by dependabot](https://github.com/colindean/transmission-blocklist/pull/9) moments after I set it up in one of my repos.

### DIFF
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -333,7 +333,7 @@ jobs:
           persist-credentials: false
 
       - name: Install the latest version of uv
-        uses: astral-sh/setup-uv@v4
+        uses: astral-sh/setup-uv@v5
 
       - name: Run zizmor 🌈
         run: uvx zizmor --format sarif . > results.sarif # (2)!


woodruffw (approved):
Thanks @colindean! FYI we have Dependabot configured here so this will be bumped automatically as well in the future.

woodruffw (approved):

woodruffw:
CC @ubiratansoares for viz: it looks like the latest version of `setup-uv` switches to "caching enabled by default," so now we need to think about that version handling behavior in the cache poisoning audit 🙂

colindean:
It seemed to me like dependabot may not update the example embedded inside of the documentation. It would be surprising to me if it did do so inside of fenced blocks inside of markdown files, but certainly welcome!

woodruffw:
Oh, my bad -- I misread the source file here. I thought this was being generated from a snippet that Dependabot was tracking, but it's still embedded in the markdown itself!

colindean:
It would be really cool if the docs pulled it in from a live example.

PR #343, opened 2024-12-21
fix: setup-go's cache-control is OptOut

jmelahman:
`setup-go`'s caching is on by default and opt-out with `cache=false`, https://github.com/actions/setup-go?tab=readme-ov-file#v4

Fixes [this](https://github.com/jmelahman/connections/blob/master/.github/workflows/release.yml) false positive,
screenshot-2024-12-21-09-34-42

### DIFF
--- a/src/audit/cache_poisoning.rs
+++ b/src/audit/cache_poisoning.rs
@@ -54,8 +54,8 @@ static KNOWN_CACHE_AWARE_ACTIONS: LazyLock<Vec<CacheAwareAction>> = LazyLock::ne
         // https://github.com/actions/setup-go/blob/main/action.yml
         CacheAwareAction {
             uses: Uses::from_step("actions/setup-go").unwrap(),
-            cache_control: CacheControl::OptOut("cache"),
-            control_value: ControlValue::String,
+            cache_control: CacheControl::OptIn("cache"),
+            control_value: ControlValue::Boolean,
             caching_by_default: true,
         },
         // https://github.com/actions/setup-node/blob/main/action.yml
--- a/src/audit/cache_poisoning.rs
+++ b/src/audit/cache_poisoning.rs
@@ -54,7 +54,7 @@ static KNOWN_CACHE_AWARE_ACTIONS: LazyLock<Vec<CacheAwareAction>> = LazyLock::ne
         // https://github.com/actions/setup-go/blob/main/action.yml
         CacheAwareAction {
             uses: Uses::from_step("actions/setup-go").unwrap(),
-            cache_control: CacheControl::OptIn("cache"),
+            cache_control: CacheControl::OptOut("cache"),
             control_value: ControlValue::String,
             caching_by_default: true,
         },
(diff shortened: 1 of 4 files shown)


woodruffw:
Thanks @jmelahman! I'll review this later today.

ubiratansoares:
My bad here, @jmelahman, thank you so much for catching this one! 

Too easy got lost in the booleans when modelling the behaviour of those cache-aware Actions, I guess 🙈 

If I got everything correctly, the issue is `setup-go`  [actually defines opt-in via the `cache` input;](https://github.com/actions/setup-go/blob/main/action.yml#L16C1-L16C106) however the control value is actually a `boolean`, not a string-as-enum in `setup-node`, `setup-java` and others.

I gave a try on your Workflow locally, setting the control value to `ControlValue::Boolean` and it fixes the issue

```bash
➜  zizmor git:(ufs/issue-343) ✗ git diff && cargo run -- tests/test-data/cache-poisoning/issue-343-repro.yml

Δ src/audit/cache_poisoning.rs
──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────

───────────────────────────────────────────────────────────────────────────────────────┐
• 55: static KNOWN_CACHE_AWARE_ACTIONS: LazyLock<Vec<CacheAwareAction>> = LazyLock::ne │
───────────────────────────────────────────────────────────────────────────────────────┘
        CacheAwareAction {
            uses: Uses::from_step("actions/setup-go").unwrap(),
            cache_control: CacheControl::OptIn("cache"),
            control_value: ControlValue::String,
            control_value: ControlValue::Boolean,
            caching_by_default: true,
        },
        // https://github.com/actions/setup-node/blob/main/action.yml
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.09s
     Running `target/debug/zizmor tests/test-data/cache-poisoning/issue-343-repro.yml`
2024-12-21T16:36:37.656862Z  WARN zizmor: skipping impostor-commit: can't run without a GitHub API token
2024-12-21T16:36:37.656919Z  WARN zizmor: skipping ref-confusion: can't run without a GitHub API token
2024-12-21T16:36:37.656941Z  WARN zizmor: skipping known-vulnerable-actions: can't run without a GitHub API token
2024-12-21T16:36:37.660063Z  INFO audit: zizmor: 🌈 completed /Users/lonewolf/RustroverProjects/zizmor/tests/test-data/cache-poisoning/issue-343-repro.yml
No findings to report. Good job! (6 suppressed)

```

I _think_ the way to go here is moving the control value to `ControlValue:Boolean`, since the description of the `cache` input in `setup-go` makes it clear that true == opt-in. 

Does that make sense?

woodruffw:
Unless I'm misunderstanding, I think `setup-go` is indeed opt-out:

> The action will try to enable caching unless the cache input is explicitly set to false.

(From the README)

However, I agree that this should be `ControlValue::Boolean`.

woodruffw:
Oh, this was my misunderstanding -- I forgot about the distinction between `CacheControl` and `caching_by_default`. @ubiratansoares's explanation makes sense to me!

woodruffw (approved):

woodruffw:
Thanks for raising this @jmelahman!

PR #345, opened 2024-12-22
feat: adds support to non-configurable caching on cache-audit

ubiratansoares:
Another follow-up for #334 

Here we tweak a bit the models for `CacheAwareAction`, moving into to an enum an introducing `Configurable` and `NotConfigurable` variants. 

The not-configurable variant covers Actions like [Mozilla-Actions/sccache-action](https://github.com/Mozilla-Actions/sccache-action), which always cache/restore and does not provide control inputs / control values to disable caching.

I added a snapshot test convering this new use case. Added some docs on the semantics of `CacheControlInput:OptIn` and `CacheControlInput:OptOut` as well.

### DIFF
--- a/src/audit/cache_poisoning.rs
+++ b/src/audit/cache_poisoning.rs
@@ -19,9 +19,9 @@ enum CacheControlValue {
 
 /// The input that controls the behaviour of a configurable caching Action
 enum CacheControlInput {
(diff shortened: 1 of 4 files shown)


woodruffw (approved):
Thanks @ubiratansoares!

PR #352, opened 2024-12-23
feat: detects push to release branches in cache-poisoning

ubiratansoares:
One more follow-up for #334

We now detect at least one common pattern for Workflows triggered with `push/branches`. Deploying when pushing to a convention release branch is quite common, and this change sets the stage to detect other similar branch filters in the future.

There is at least one additional PR to before finishing #334, raising it asap!


woodruffw (approved):
LGTM, nice work @ubiratansoares!

PR #353, opened 2024-12-24
feat: adds case-insensitive template matching for uses actions

### LINKED ISSUE #334: cache-poisoning: follow-up improvements

ubiratansoares:
Closes #334 

Improves the implementation of `RepositoryUses::matches` so it becomes case-insensitive regarding a provided template.

I opted for case-insensitive comparisons rather than normalizing to a lowercased template because I see this approach introducing less changes, and all of them live within a single function.

### DIFF
--- a/src/models.rs
+++ b/src/models.rs
@@ -627,8 +627,8 @@ impl RepositoryUses<'_> {
             return false;
         };
 
-        self.owner.to_lowercase() == other.owner.to_lowercase()
-            && self.repo.to_lowercase() == other.repo.to_lowercase()
+        self.owner.eq_ignore_ascii_case(other.owner)
+            && self.repo.eq_ignore_ascii_case(other.repo)
             && self.subpath.map(|s| s.to_lowercase()) == other.subpath.map(|s| s.to_lowercase())
             && other.git_ref.map_or(true, |git_ref| {
                 Some(git_ref.to_lowercase()) == self.git_ref.map(|r| r.to_lowercase())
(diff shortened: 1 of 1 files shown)


src/models.rs
```diff
@@ -617,17 +617,22 @@ impl RepositoryUses<'_> {
...
         };
 
-        self.owner == other.owner
-            && self.repo == other.repo
-            && self.subpath == other.subpath
-            && other
-                .git_ref
-                .map_or(true, |git_ref| Some(git_ref) == self.git_ref)
+        self.owner.to_lowercase() == other.owner.to_lowercase()
+            && self.repo.to_lowercase() == other.repo.to_lowercase()
```
woodruffw, line 631: The `owner` and `repo` could probably be compared with [eq_ignore_ascii_case](https://doc.rust-lang.org/std/primitive.str.html#method.eq_ignore_ascii_case) instead, to avoid a few allocations (since GH requires both to be ASCII).

The subpath and git_ref can be anything OTOH, so those look good to me!

ubiratansoares: Awesome input, justed learned about another small gem in Rust std 🙂

Done [here](https://github.com/woodruffw/zizmor/pull/353/commits/6e76cbfbc2aa2831754b1b9c633ee3de43d1bc7d)

woodruffw (approved):
Thanks @ubiratansoares, nice work!

Issue #234, opened 2024-12-06
[BUG]: False positive: github-env on trivial echos


woodruffw:
### Pre-submission checks

- [X] I am **not** filing a feature request. These should be filed via the feature request form instead.
- [X] I have looked through the [open issues](https://github.com/woodruffw/zizmor/issues?q=is%3Aissue+is%3Aopen+) for a duplicate report.

### Expected behavior

The following should not trigger the github-env audit, since it's trivially safe: the LHS has only one envvar and no newlines or other injection potential.

```
echo "PYPI_URL=https://upload.pypi.org/legacy/" >> $GITHUB_ENV
```

### Actual behavior

```
error[github-env]: dangerous use of GITHUB_ENV
  --> /Users/alex_gaynor/projects/cryptography/.github/workflows/pypi-publish.yml:39:9
   |
39 |         - run: |
   |  _________^
40 | |           echo "PYPI_URL=https://upload.pypi.org/legacy/" >> $GITHUB_ENV
   | |________________________________________________________________________^ GITHUB_ENV write may allow code execution
   |
```

### Reproduction steps

N/A, h/t @alex

woodruffw:
For context, this is what the tree-sitter tree for the above looks like:

```
program [0, 0] - [1, 0]
  redirected_statement [0, 0] - [0, 62]
    body: command [0, 0] - [0, 47]
      name: command_name [0, 0] - [0, 4]
        word [0, 0] - [0, 4]
      argument: string [0, 5] - [0, 47]
        string_content [0, 6] - [0, 46]
    redirect: file_redirect [0, 48] - [0, 62]
      destination: simple_expansion [0, 51] - [0, 62]
        variable_name [0, 52] - [0, 62]
```

So for trivial cases, at least, we could detect when the `body` has `name = echo` and has a single `argument`, which is a single `string_content` with no interior expansions.

woodruffw:
Fixed by #354.

PR #354, opened 2024-12-25
feat: github-env: use tree-sitter queries

woodruffw:
~~Very WIP.~~

This experimentally replaces the stack-of-nodes approach with hand-written TS queries, which can be more precise about matches and their sub-captures (e.g. allowing us to match only when the RHS is actually a variable expansion).

It also adds some examples of false positives that the previous approach flagged. These FPs aren't likely to occur in the wild, however, so the direct benefit for actual users is pretty small.

The larger impact here is indirect:

* By switching to TS queries, we get full access to match and capture-level spanning information. We had this indirectly with the stack-of-nodes approach, but the query language makes it even simpler by giving us capture names 🙂 
* This unblocks a bugfix for #234 which would be challenging to generalize with the previous approach -- we could do it one-off, but doing things at the TS query level gives us more generality.

At the same time, there are some downsides to this approach:

* It's not as elegant (IMO) as the previous approach.
* It's probably slower both at startup and runtime, since each TS query needs to be parsed, evaluated, etc.
* Writing TS queries is a less-ideal DX, since they're a new DSL that isn't highlighted and aren't super easy to debug (I mostly use the tree-sitter playground).

CC @ubiratansoares for thoughts on this approach -- I think it makes things somewhat easier/more general overall, but it has downsides. I'm curious what you think about these tradeoffs 🙂 

(I also still need to do the same for the PowerShell codepath.)

### DIFF
--- a/src/audit/github_env.rs
+++ b/src/audit/github_env.rs
@@ -9,7 +9,7 @@ use std::cell::RefCell;
 use std::ops::{Deref, Range};
 use std::sync::LazyLock;
 use streaming_iterator::StreamingIterator;
-use tree_sitter::{Parser, Query, QueryCapture, QueryCursor, QueryMatches, Tree};
+use tree_sitter::{Language, Parser, Query, QueryCapture, QueryCursor, QueryMatches, Tree};
 
 static GITHUB_ENV_WRITE_CMD: LazyLock<Regex> =
     LazyLock::new(|| Regex::new(r#"(?mi)^.+\s*>>?\s*"?%GITHUB_ENV%"?.*$"#).unwrap());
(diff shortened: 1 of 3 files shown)


src/audit/github_env.rs
```diff
@@ -143,9 +192,21 @@ impl WorkflowAudit for GitHubEnv {
...
             .set_language(&pwsh)
             .context("failed to load powershell parser")?;
+
+        let bash_redirect_query = Query::new(&bash.into(), BASH_REDIRECT_QUERY)?;
+        let bash_pipeline_query = Query::new(&bash.into(), BASH_PIPELINE_QUERY)?;
+
         Ok(Self {
             bash_parser: RefCell::new(bash_parser),
             pwsh_parser: RefCell::new(pwsh_parser),
+            bash_redirect_query_span_idx: bash_redirect_query
```
woodruffw, line 202: NB: I cache these capture indices at the struct level because they vary between individual queries, and because extracting them is technically fallible (although infallible in practice, as long as there's no typo). But it's definitely a bit of an eyesore.

woodruffw:
NB https://github.com/tree-sitter/tree-sitter/pull/4034 -- that'll allow us to remove the `streaming-iterator` direct dep.

woodruffw:
> * This unblocks a bugfix for [[BUG]: False positive: github-env on trivial echos #234](https://github.com/woodruffw/zizmor/issues/234) which would be challenging to generalize with the previous approach -- we could do it one-off, but doing things at the TS query level gives us more generality.

[`1a1eac7` (#354)](https://github.com/woodruffw/zizmor/pull/354/commits/1a1eac784f4a0b4e690b58dbbc2b4c69b0ee8e0a) does an initial version of this -- it's not _quite_ as clean as I would have liked, but it's probably shorter than the equivalent would be if we didn't have the query doing pre-filtering/shape checking for us.

ubiratansoares (commented):
Hey @woodruffw! Thanks for tagging me for input!

I was not aware of TS query capabilities, always learning I guess 😄 

Imho, this approach looks amazing. I'm sold tbh, regardless the downsides!

I do agree that TS queries might be not as intituitive to augment or troubleshoot compared to walking nodes directly, but they look like a solid take for tackling more false positives we may have in the future.

Perf-wise, I _think_ we should not bother that much for now, afterall `zizmor` is already pretty fast elsewhere.

Since https://github.com/tree-sitter/tree-sitter/pull/4034 was already merged, perhaps you want to wait until the next tree-sitter release to merge this. Otherwise, LGTM 🚀

Issue #362, opened 2024-12-26
Feature: add argument --ghe-hostname for GHE Servers


runjivu:
### Pre-submission checks

- [X] I am **not** reporting a bug (crash, false positive/negative, etc). These must be filed via the bug report template.
- [X] I have looked through the open issues for a duplicate request.

### What's the problem this feature will solve?

I'm trying zizmor out with GHE Server repos, but for now github_api client's api_base is fixed to "https://api.github.com".

### Describe the solution you'd like

by adding additional parameter --ghe-hostname, i can run something like
`zizmor --gh-token {token} --ghe-hostname {github.example.com} example/repos`,
which would be nice

### Additional context

I am new to GitHub Actions security hardening.
In a GHE Server / self-hosted runner environments, it seems/feels inherently safer against many attacks..
Are they really?
I’d love to hear some advice on what specific things I should focus on to keep things secure in these environments! 😃

woodruffw:
Hi @runjivu, thanks for the feature request!

Overall, this sounds like a good idea to me -- supporting both hosted and on-prem GHE is something we should do.

A couple of thoughts:

1. How does this interact across multiple instances/hosts? For example, if a repo on a GHE host references `actions/checkout`, how do we ensure that the user's `GH_TOKEN` works for both the GHE host and the normal GH host, e.g. for the `impostor-commit` audit?
2. How does GitHub's own `gh` CLI handle these cases? IMO we should closely mirror their own flags/behavior here.

> In a GHE Server / self-hosted runner environments, it seems/feels inherently safer against many attacks..
> Are they really?

It depends 🙂: self-hosted runners come in a large variety of configurations, some of which are less secure than GitHub-hosted runners. In general, GitHub recommends only using self-hosted runners on private repos, regardless of whether the repo is on a GH or GHE host.

In general however, my understanding is that the "best practice" for self-hosted runners is to *never* use persistent runners (*only* use ephemeral runners), and to manage your runners at whatever organizational level makes sense for your use case (e.g. repo/org/enterprise level) and *not* allow deviations from that level.

Some more docs here: <https://docs.github.com/en/actions/hosting-your-own-runners/managing-self-hosted-runners/about-self-hosted-runners#self-hosted-runner-security>

runjivu:
Hi @woodruffw , thank you for the quick response!

my initial thoughts on your points : 
1. > How does GitHub's own gh CLI handle these cases?

when logging in with `gh auth login`,
[...]

PR #363, opened 2024-12-26
feat: add argument --ghe-hostname for GHE Servers

runjivu:
adds #362 feature !

+ cargo test passed
+ auditing GHE Server's remote repo works
+ --offline --ghe-hostname fails as expected

### DIFF
--- a/src/github_api.rs
+++ b/src/github_api.rs
@@ -29,7 +29,7 @@ pub(crate) struct Client {
 }
 
 impl Client {
-    pub(crate) fn new(token: &str, cache_dir: &Path) -> Self {
+    pub(crate) fn new(token: &str, cache_dir: &Path, hostname: Option<&str>) -> Self {
         let mut headers = HeaderMap::new();
         headers.insert(USER_AGENT, "zizmor".parse().unwrap());
         headers.insert(
@@ -66,8 +66,13 @@ impl Client {
         }))
         .build();
 
+        let api_base = match hostname {
+            Some(hostname) => format!("https://{hostname}/api/v3"),
+            None => "https://api.github.com".to_string(),
+        };
+
         Self {
-            api_base: "https://api.github.com",
+            api_base: Box::leak(api_base.into_boxed_str()),
             http,
         }
     }
--- a/src/github_api.rs
+++ b/src/github_api.rs
@@ -23,13 +23,51 @@ use crate::{
     utils::PipeSelf,
 };
 
+/// Represents different types of GitHub hosts
+#[derive(Debug, PartialEq)]
+pub(crate) enum GitHubHost {
+    Garage,
+    Enterprise,
+    Localhost,
+    Standard,
+}
+
+impl GitHubHost {
+    fn from_hostname(hostname: &str) -> Self {
+        let normalized = hostname.to_lowercase();
+
+        if normalized.eq_ignore_ascii_case("garage.github.com") {
+            Self::Garage
+        } else if normalized.eq_ignore_ascii_case("github.localhost") {
+            Self::Localhost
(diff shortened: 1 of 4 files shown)


src/main.rs
```diff
@@ -48,13 +48,18 @@ struct App {
...
     offline: bool,
 
     /// The GitHub API token to use.
-    #[arg(long, env, group = "_offline")]
+    #[arg(long, env)]
     gh_token: Option<String>,
 
+    /// The GitHub API Server Hostname, if you are running GHE.
+    #[arg(long, env)]
+    gh_hostname: Option<String>,
```
woodruffw, line 61: Rather than a `None` state, perhaps we can default to `https://api.github.com/` here? That way there's no need for the subsequent `Option<...>` handling 🙂

src/github_api.rs
```diff
@@ -66,8 +66,13 @@ impl Client {
...
         .build();
 
+        let api_base = match hostname {
+            Some(hostname) => format!("https://{hostname}/api/v3"),
+            None => "https://api.github.com".to_string(),
+        };
+
         Self {
-            api_base: "https://api.github.com",
+            api_base: Box::leak(api_base.into_boxed_str()),
```
woodruffw, line 75: No `Box::leak`, please! If this needs to be a `String` instead of a `&'static str`, that's perfectly fine.

runjivu: Will change that!

src/main.rs
```diff
@@ -48,13 +48,18 @@ struct App {
...
+    #[arg(short, long, env = "ZIZMOR_OFFLINE", 
+        conflicts_with_all = ["gh_token", "gh_hostname"])]
     offline: bool,
 
     /// The GitHub API token to use.
-    #[arg(long, env, group = "_offline")]
+    #[arg(long, env)]
     gh_token: Option<String>,
 
+    /// The GitHub API Server Hostname, if you are running GHE.
```
woodruffw, line 59: Nitpick: slightly shorter and more generic (since this will also allow people to proxy the GH API, in addition to working with GHE 🙂)

```suggestion
    /// The GitHub API host to use.
```

src/main.rs
runjivu: We could, but there should be handing at github_api.rs making api_base,
because we have to add /api/v3 if it is on GHE server..

Should i change it to String?

woodruffw: > because we have to add /api/v3 if it is on GHE server..

I might be missing something, but is this always true? The GHE docs don't seem to consistently mention an `/api/v3` prefix for API routes. I'm wondering if it's a redundant prefix, i.e. whether `/{route} and `/api/v3/{route}` both work.

runjivu: https://docs.github.com/en/enterprise-server@3.15/rest/quickstart?apiVersion=2022-11-28&tool=curl

Api docs for ghe server mentions /api/v3,
but I'll test it again to make sure!

woodruffw: Ah yep, looks like there's no redirect, at least for `github.com`.



> (Maybe gh handles this internally..?)

Yep, it looks like they do:

https://github.com/cli/cli/blob/6c5145166003ac6fb952c5c591a6f3bdeea10465/internal/ghinstance/host.go#L59-L70

That uses `ghauth.IsEnterprise`, which seems to come from here:

https://github.com/cli/go-gh/blob/65bd8d766abd062846dd5cffe21defab7c0fe4c5/pkg/auth/auth.go#L159-L166

So IMO we should copy that logic :slightly_smiling_face:

woodruffw:
Thanks @runjivu! I've folded this into #371.

Issue #350, opened 2024-12-23
Composite actions: support for existing audits


woodruffw:
#331 will add composite action support. From there, the current audits need to be extended to work with composite actions, where appropriate.

- [x] artipacked #896 
- [x] dangerous-triggers (N/A)
- [x] excessive-permissions (N/A)
- [x] hardcoded-container-credentials (N/A?)
- [x] impostor-commit: #331
- [x] known-vulnerable-actions @ubiratansoares #367
- [x] ref-confusion @ubiratansoares #365
- [x] self-hosted-runner (N/A?)
- [x] template-injection: #331
- [x] use-trusted-publishing #899 
- [x] unpinned-uses @ubiratansoares #364
- [x] insecure-commands #359
- [x] github-env #358 
- [x] cache-poisoning (N/A?)

woodruffw:
I'm going to keep burning this down, but I would appreciate help as well! If anybody is interested in working on one of these, leave a comment here and I'll assign it to you 🙂

ubiratansoares:
@woodruffw Happy to help here!

Let me give a try on `known-vulnerable-actions`, `unpinned-uses, and `ref-confusion`, they don't look too hard and I can learn a bit more about their inner details along the way. I'll raise individual PRs for each one, for the sake of easier reviews 🙂

ubiratansoares:
@woodruffw Happy new year 🎉🎉🎉

If ok for you, I'm happy take the 2 remaining audits, so we get the all audits supported in Composite Actions 🙂

woodruffw:
That would be awesome, thank you @ubiratansoares! And happy new year!

woodruffw:
#896 and #899 wrap this up 🙂

PR #364, opened 2024-12-26
feat: adds unpinned-uses support to composite actions

ubiratansoares:
Contributes to #350 

I tried to dedup a bit the inner logic for this audit by extracting at least one common function. I do think there is room for more deduplication, but I did not want to push hard on that in the first take. 

Let me know your thoughts 🙂

### DIFF
--- a/docs/audits.md
+++ b/docs/audits.md
@@ -547,9 +547,9 @@ or @rubygems/release-gem for canonical examples of using it.
 
 ## `unpinned-uses`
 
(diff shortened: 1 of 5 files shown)


woodruffw:
Looks good to me! Feel free to dedupe here or in a follow-up, whatever you prefer 🙂

ubiratansoares:
Follow-up then, so I have room to get more used to the new CompositeStep APIs 🙂. 

I'll take a self-note to revisit this one!

woodruffw (approved):
Thanks @ubiratansoares!

PR #367, opened 2024-12-26
feat: adds composite actions support to known-vulnerable-actions

ubiratansoares:
Another follow-up for #350. Very similar to #365.  🙂


woodruffw (approved):
Thanks!

PR #379, opened 2025-01-03
docs: ditches beta warning from static website

ubiratansoares:
A follow-up for #377 . 

Just checked `zizmor` website and there is one beta warning  left at `docs/index.md`. Removed just in case 😄


woodruffw (approved):
Thanks!

Issue #378, opened 2025-01-02
[BUG]: How to address "opt-in for caching here"?


ashishb:
### Pre-submission checks

- [X] I am **not** filing a feature request. These should be filed via the feature request form instead.
- [X] I have looked through the [open issues](https://github.com/woodruffw/zizmor/issues?q=is%3Aissue+is%3Aopen+) for a duplicate report.

### Expected behavior

When Docker images are built (but not pushed), cached artifacts are harmless.
How do I disable https://woodruffw.github.io/zizmor/audits/#cache-poisoning for those cases?

### Actual behavior

```
error[cache-poisoning]: runtime artifacts potentially vulnerable to a cache poisoning attack
  --> ...
   |
34 | /         with:
35 | |           cache-binary: true
   | |____________________________^ opt-in for caching here
36 |
37 |         - name: Build docker
38 |           uses: docker/build-push-action@v6
   |           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ runtime artifacts usually published here
   |
```

### Reproduction steps

Create a new action

```
      - name: Set up Docker Buildx
        uses: docker/setup-buildx-action@v3
        with:
          cache-binary: true

      - name: Build docker
        uses: docker/build-push-action@v6
        with:
          cache-from: type=gha
          cache-to: type=gha,mode=max
          push: false
```

woodruffw:
Thanks for the report @ashish!

To make sure I understand: is the false positive happening because `push: false` disables the publish, but we don't detect that?

If that sounds right to you, I think what we need to do here is better filter the publish actions in a manner similar to the cache-aware actions. CC @ubiratansoares for thoughts 🙂

ashishb:
> To make sure I understand: is the false positive happening because push: false disables the publish, but we don't detect that?

Yes, I am not an expert on caching but I do believe that `push: false` does kill this attack.

ubiratansoares:
@ashishb Thanks for reporting! 

I agree with @woodruffw, this looks like a false positive and to proper handle it we need to enhance what we have for publisher-like Actions (like `docker/build-push-action`) so we model whether publishing can be disabled or ran in some sort of dry-run mode.

For now, I think the easiest way to disable this check is using zizmor's support to [ignoring results](https://woodruffw.github.io/zizmor/usage/#ignoring-results).

danweller18:
[...]

PR #383, opened 2025-01-05
refactor: CacheAwareAction -> UsesCoordinate

woodruffw:
WIP.

This is the first part of #378: it generalizes `CacheAwareAction` into `UsesCoordinate`, which we can then reuse for conditions on the publish step as well.


woodruffw:
This is not yet fully general, since the matching logic still lives in `cache-poisoning`. But I'll do that in a follow-up.

PR #415, opened 2025-01-10
docs: Nextcloud also secured it's actions with zizmor

nickvergessen:
Thanks a lot for the great tool :)

### DIFF
--- a/docs/snippets/trophies.md
+++ b/docs/snippets/trophies.md
@@ -337,6 +337,14 @@
         - NetApp/harvest#3247
 
 
+-   ![](https://github.com/nextcloud.png?size=40){ width="40" loading=lazy align=left } nextcloud
+
+    ---
+
+    ??? example "Examples"
+        - nextcloud/.github#477
+
+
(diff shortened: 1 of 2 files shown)


woodruffw (approved):
Thanks @nickvergessen!

nickvergessen:
Hmm the `.` seems to cause weird things:
> grafik

It also had an error at the beginning when I tried to run `make trophies` Maybe you need to adjust a regex somewhere :see_no_evil:

woodruffw:
Hah. Yeah, let me see about tweaking that. Thanks regardless for updating the trophy, I appreciate it!

woodruffw:
Huh, this looks like it's probably a bug in MagicLink: `{user}/{repo}#num` should always work: https://facelessuser.github.io/pymdown-extensions/extensions/magiclink/#issues-pull-requests-and-discussions

I'll file an upstream issue.

woodruffw:
Fixed upstream: https://github.com/facelessuser/pymdown-extensions/pull/2563, so this should be fixed for us whenever the dep rolls over.

PR #419, opened 2025-01-10
fix: update to use static link for PyPI display

chenrui333:
currently the demo link on pypi is not working, updating to use a static link

image


woodruffw:
Thanks @chenrui333!

PR #422, opened 2025-01-10
docs: fix formatting

userdocs:
fix code block formatting


woodruffw (approved):
thanks!

woodruffw:
Yep, it was correct but I prefer the changes you've suggested -- IME most YAML documents prefer the latter and auto-formatters prefer it as well.

PR #453, opened 2025-01-17
bugfix: sarif: use absolute physical locations only

### LINKED ISSUE #452: [BUG]: Incorrect physical location for composite actions in subdirectories
### Pre-submission checks

- [x] I have looked through the [open issues](https://github.com/woodruffw/zizmor/issues?q=is%3Aissue+is%3Aopen+) for a duplicate report.

### Expected behavior

The physical location contains the full path from the root of the repository.

```json
"logicalLocations": [
  {
    "properties": {
      "symbolic": {
        "key": {
          "Local": {
            "path": "/home/runner/work/repository/repository/subdirectory/action.yml"
          }

[...]

"physicalLocation": {
  "artifactLocation": {
    "uri": "subdirectory/action.yml",
    "uriBaseId": "%SRCROOT%"
  },
```

### Actual behavior

The physical location is always `action.yml` regardless of how nested it is.

```json
"logicalLocations": [
  {
    "properties": {
      "symbolic": {
        "key": {
          "Local": {
            "path": "/home/runner/work/repository/repository/subdirectory/action.yml"
          }

[...]

"physicalLocation": {
  "artifactLocation": {
    "uri": "action.yml",
    "uriBaseId": "%SRCROOT%"
  },
```

The result of this is that GitHub code scanning will not show where the problem is.

### Reproduction steps

1. Create a composite action that fails an audit.
2. Make sure that action.yml is in a subdirectory, e.g. `subaction/action.yml`
3. Run `zizmor --format sarif`.
4. Note the physical location does not contain the subdirectory.

### Logs

I can DM you a real-world example where this happened.

### Additional context

_No response_

woodruffw:
This changes the SARIF generation so that `physicalLocations` now always contains an absolute path to the input. This is a slight over-correction of the previous behavior, which would use a repo-relative path when possible and fall back on the filename component when it couldn't infer one.

In practice, the trick used by the previous behavior worked well for workflow path (which are always under `.github/workflows` but doesn't generalize for action paths, which can be anywhere. As a result there's no "anchor" to distinguish the repository prefix from the rest of the path.

This should have no negative effect on most users: local users will have their full paths in the SARIF instead, as will remote (GitHub Actions) users. In the latter case, GitHub Advanced Security should be able to "rebase" the paths correctly.

Fixes #452.

### DIFF
--- a/docs/release-notes.md
+++ b/docs/release-notes.md
@@ -15,6 +15,9 @@ of `zizmor`.
   as unpinned/unhashed (#439)
 * The [excessive-permissions] audit has been refactored, and better captures
   both true positive and true negative cases (#441)
+* The SARIF output mode (`--format=sarif`) now always returns absolute paths
+  in its location information, rather than attempting to infer a (sometimes
+  incorrect) repository-relative path (#453)
 
 ### Fixed
 
--- a/src/main.rs
+++ b/src/main.rs
@@ -394,7 +394,7 @@ fn run() -> Result<ExitCode> {
         OutputFormat::Plain => render::render_findings(&registry, &results),
         OutputFormat::Json => serde_json::to_writer_pretty(stdout(), &results.findings())?,
         OutputFormat::Sarif => {
-            serde_json::to_writer_pretty(stdout(), &sarif::build(&registry, results.findings()))?
+            serde_json::to_writer_pretty(stdout(), &sarif::build(results.findings()))?
         }
     };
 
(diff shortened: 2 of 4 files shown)


woodruffw:
Did some more looking: GitHub recommends relative paths (reasonable), but it looks like they'll handle absolute paths in SARIFs just fine: `upload-sarif` will relativize using `checkout_path`, which defaults to `${{ github.workspace }}`. That will be correct for 99.9% of users; certainly more correct than the previous behavior.

https://github.com/github/codeql-action/blob/c2c0a2908e95769d01b907f9930050ecb5cf050d/analyze/action.yml#L44-L47

Issue #400, opened 2025-01-06
New audit: secret leakage


woodruffw:
By default, secrets in the `secrets.*` context are redacted in logs, etc. However, if a secret is JSON, this kind of pattern can accidentally disclose it:

```
${{ fromJSON(secrets.oopsie).foo.bar }}
```

This happens because `secrets.oopsie` is redacted, but substrings/arbitrary decodings of `secrets.oopsie` are not.

As such, we should probably detect any flow of `fromJSON(secrets).attr.access` or `fromJSON(secrets.abc).attr.access` and flag it.

Doing this fully generally is a little tricky, since expressions can appear *literally almost anywhere*. So the existing audit infrastructure might be insufficient for this; instead we probably need an `audit_raw(body: &str)` or similar.

PR #485, opened 2025-01-23
feat: "raw" audit support + `overprovisioned-secrets`

### LINKED ISSUE #401: New audit: excessive secret provisioning
By default, the GitHub Actions runner will only include/inject explicitly referenced secrets. For example, if a repo has `secrets.{foo,bar,baz}` but `workflow.yml` only mentions `secrets.{foo,bar}`, then `secrets.baz` will not be included at runtime.

However, *if* the workflow references the entire `secrets` context, the runner can't filter the selected secrets and will instead include the entire context. For example:

```
# good: no other secrets added to runner
secrets.foo

# bad: adds all secrets to runner despite only accessing foo
fromJSON(secrets).foo
```

We should flag on any "bare" use of `secrets`.

Like #400, this is going to require some internal refactoring/rethinking to enable us to scan all expressions in the input, regardless of the context they appear in.

woodruffw:
~~WIP.~~

This will enable audits like #400 and #401.

Some open design questions:

~~1. How should we shoehorn this into `SymbolicLocation`/`ConcreteLocation`? In particular, how should we ensure that ignore comments are captured correctly?~~

Closes #401.


woodruffw:
The `overprovisioned-secrets` audit now works, although it doesn't support comment ignores yet. That's the last blocking piece here.

PR #506, opened 2025-01-31
ci: make site.yml reusable

woodruffw:


woodruffw:
This doesn't work without excessively broadening the environment, which defeats the point.

PR #519, opened 2025-02-05
bugfix: fix has_tag lookup

### LINKED ISSUE #518: ref-confusion in setup-task
### Discussed in https://github.com/woodruffw/zizmor/discussions/517

<sup>Originally posted by **philpennock** February  5, 2025</sup>
This is probably me being stupid, but I'm confused by a `ref-confusion` complaint I get.
(`zizmor --version` says `zizmor 1.3.0`, was installed via `cargo install --locked zizmor`)

If we have a GitHub Actions step:
```
      - name: Install Task
        uses: arduino/setup-task@v2
```

then with GH_TOKEN defined, we get:
```
warning[ref-confusion]: git ref for action with ambiguous ref type
  --> .github/workflows/export-dashboards.yml:25:9
   |
25 |         uses: arduino/setup-task@v2
   |         --------------------------- uses a ref that's provided by both the branch and tag namespaces
   |
```

(And no, we're not pinning actions to specific checksums, we're not set up to be tracking and getting security updates if we do that, so do not _yet_ intend to make that switch.)

When I check in that `arduino/setup-task` action repo, there are _branches_ `v1` and `v2` and `git tag -l` has:
```
1.0.2
v1.0.0
v1.0.1
v1.0.3
v2.0.0
```

So I'm confused about where the ref confusion is coming from.  AFAICT, `@v2` must be a branch in `refs/heads/` and can't be in `refs/tags/`

What am I missing, please?</div>

woodruffw:
This caused false positives in the `ref-confusion`
audit by using the wrong endpoint. The correct
endpoint only returns the exact matching tag or
404, rather than a list of one or more
partial matches.

See #517. Fixes #518.


woodruffw:
WIP, since `test_with` doesn't do exactly what I need: it's still a build-time rather than a runtime check, so running `cargo insta` multiple times against the same compilation product but with different env states causes problems.

PR #574, opened 2025-02-26
Add `forbidden-uses` audit

### LINKED ISSUE #558: Feature: Manual allowlist/denylist for actions (`forbidden-uses`)
### Pre-submission checks

For organizations that are worried about supply-chain security, it can be important to only allow actions that have been reviewed and approved.

A new check that outputs a warning if an actions is used that is not on an allowlist.

By default, the check should only allow:
- Official GitHub Actions (`actions/*`)
- First-party actions (hosted on the same user/organization as the repo itself.

Additional actions could be added to an allowlist (via settings). Inversely, actions that have should not be used could be put into a denylist (and thus always result in high-severity finding).
Since most users don't need this, findings for non-allowlisted actions should probably be suppressed by default.

### Additional context

Some ideas for settings:

```yaml
rules:
  forbidden-uses:
    allow:  # Matching actions will not result in a finding
      - "actions/*"
      - "@self/*"  # Something like this might be used to allowlist first-party actions (?)
      - "some-user/trusted-action"
      - "trusted-org/*"
      - "another-user/action-where-only-a-specific-version-is-trusted@e046bf4c08d98a867ad0fce4a27d3c0d06976598"
    deny:  # Matching actions will result in finding of level error (or warning?)
     - "untrusted-user/*"
     - "some-user/untrusted-action"
     - "another-user/action-where-only-a-specific-version-is-untrusted@5790ee9005aa9399722753077ba64e92947a5727"
     - "*"  # Add this if you want to deny everything that is listed under allow
```

Holzhaus:
Fixes #558.

Not finished yet, but if you have any comments on the configuration part, let me know.


src/config.rs
```diff
@@ -158,6 +163,19 @@ impl Config {
 
         false
     }
+
+    pub(crate) fn rule_config<'a, T>(&'a self, ident: &str) -> Option<T>
```
woodruffw, line 167: I _think_ this should be `Result<T>`, with the following states:

* If `policy` is `Value::Null`, we should return `Ok(T::default())`
* If `policy` loads correctly, we should return `Ok(policy)`
* If `policy` fails to deserialize, we should return `Err(...)`

That way we present an appropriate hard error to the user when their configuration is invalid, rather than using the default configuration (which is probably not that they intended) 🙂

woodruffw: (This also requires a `T: Default` constraint, but IMO that's fine!)

src/audit/forbidden_uses.rs
```diff
@@ -0,0 +1,123 @@
...
+impl ForbiddenUses {
+    pub fn evaluate_allowlist<'u>(&self, uses: &Uses) -> Option<(&'u str, Severity, Persona)> {
+        let Uses::Repository(repo_uses) = uses else {
+            return None;
+        };
+
+        self.config.is_uses_denied(repo_uses).then_some((
+            "action is not on the allowlist",
+            Severity::Medium,
+            Persona::default(),
```
woodruffw, line 54: I think this needs to be the "auditor" persona when the user doesn't supply a configuration, since otherwise this will be _way_ too noisy for typical users.

woodruffw:
Thanks @Holzhaus! This looks great as a start; I left some initial comments.

src/audit/mod.rs
```diff
@@ -180,7 +182,7 @@ pub(crate) use audit_meta;
 /// **only** [`Audit::audit`] and not [`Audit::audit_normal_job`] or
 /// [`Audit::audit_step`].
 pub(crate) trait Audit: AuditCore {
-    fn new(state: AuditState) -> Result<Self>
+    fn new(state: AuditState, config: &Config) -> Result<Self>
```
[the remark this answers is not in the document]
Holzhaus: There are reasons why I didn't do that `AuditState` is cloned for every audit. I didn't want to do that for every audit since it's quite expensive.
Of course we could store a reference to the config in the audit state if you prefer that.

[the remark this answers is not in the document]
woodruffw: > Of course we could store a reference to the config in the audit state if you prefer that.

Yep, that's what I was thinking! I think we could also make `AuditState` into a reference for audits as well; IIRC the only reason it was owned instead of borrowed is because of an earlier refactor that should no longer be relevant 🙁

woodruffw:
Just flagging as a note for myself: when documenting this new audit, we'll need to make sure users understand what it can and can't do. In particular, if a user is concerned about [impostor commits](https://www.chainguard.dev/unchained/what-the-fork-imposter-commits-in-github-actions-and-ci-cd), then this audit doesn't protect against those directly since `foo/bar@hash...` can be a `hash...` from a different `org/repo` in `foo/bar`'s network.

This is OK IMO, since we have the `impostor-commit` audit for exactly this purpose. But it should be clear in the docs to users that they need to run both, if impostor commits are inside their threat model.

woodruffw:
I thought about this some more, and I think I want to merge this with the existing `unpinned-uses` audit -- the two overlap closely, and would conflict in some ways if left unmerged.

PR #575, opened 2025-02-27
feat: respect .gitignore files when collecting inputs

### LINKED ISSUE #556: Feature: Do not scan directories that are gitignore'd
### Pre-submission checks

For repos with large dep and artifact trees (Cargo's `./target/` dir, `node_modules`, `vendor` directories), zizmor spends a lot of time recursing into them and scanning for workflows and actions I do not want it to scan, even if they were found.

Respect `.gitignore` (in all levels of the directory tree) to exclude directories from scanning automatically.

I'd prefer to use `.gitignore` (or something like `pyproject.toml`) for this to avoid yet another config file in the repo root.

### Additional context

_No response_

woodruffw:
This uses the excellent `ignore` crate to honor `.gitignore` while collecting inputs. 

For the moment, we only honor `.gitignore` and its family of files (including the global `.gitignore` and `.git/info/exclude`). We don't honor `.ignore` or similar files, although we could in the future.

Additionally, in the future we could easily honor our own `.zizmorignore` or similar file, although I don't love the idea of a separate config file for this. Instead, that could perhaps go in `zizmor.yml` and be configured via `ignore`'s support for overrides.

Closes #556.


woodruffw:
This works well for me locally, but I'm realizing that I don't have a great regression testcase for ensuring this doesn't accidentally break things. So I'm going to work on that first before merging here.

PR #586, opened 2025-03-06
feat(cli): fine-grained color control

woodruffw:
~~WIP.~~

~~This doesn't quite work yet, since tracing_indicatif and anstream::AutoStream don't compose cleanly. I'm not 100% sure if this is a fundamental incompatibility or API misuse on my part, yet.~~


woodruffw:
I thought about this some more, and I think this approach will never compose 100% to my liking: `anstream` is designed to strip all ANSI escapes, including the escapes that make progress bars render nicely. In other words, there's no clean composition between `tracing_indicatif` and `anstream` at the tracing writer level, since the former always wants some degree of ANSI escaping and the latter always wants to filter ANSI codes (when requested).

Instead, the better approach here is probably to avoid composing the "no color" and progress bar use cases: forcefully disabling colors should probably also disable the progress bar, even though it's technically its own thing.

The alternative to the above would be do so something more like what `uv` does, i.e. maintain a `Printer` abstraction that cooperates with a separate progress bar manager to ensure everything remains synchronized without having to cross underlying I/O streams.

woodruffw:
The above is no longer an immediate problem, since `--no-progress` (#589) means that we can disable the progress bar in the troublesome case.

As such, this is now close enough to what I'm thinking. But it still needs tests.

PR #599, opened 2025-03-11
Add google/zx to trophies

antonmedv:


woodruffw:
Thanks @antonmedv! FYI this is already present, but only for the first PR in your list. Could you move `google/zx#1126` next to the existing entry and run `make trophies`? Thanks!

antonmedv:
Done!

PR #603, opened 2025-03-14
ci: pypi: use manylinux_2_28 consistently

woodruffw:
Using auto causes us to default to exceptionally
old versions of glibc, along with containers that
struggle badly to build things like tree-sitter.

See: https://github.com/tree-sitter/tree-sitter/issues/4186

See: https://github.com/tree-sitter/tree-sitter/issues/4271

### DIFF
--- a/.github/workflows/pypi.yml
+++ b/.github/workflows/pypi.yml
@@ -22,19 +22,19 @@ jobs:
             manylinux: "2_28"
           - runner: ubuntu-24.04
             target: x86
-            manylinux: "auto"
+            manylinux: "2_28"
           - runner: ubuntu-24.04
             target: aarch64
             manylinux: "2_24"
           - runner: ubuntu-24.04
             target: armv7
-            manylinux: "auto"
+            manylinux: "2_28"
           - runner: ubuntu-24.04
             target: s390x
-            manylinux: "auto"
+            manylinux: "2_28"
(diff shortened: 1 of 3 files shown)


woodruffw:
We're in a pickle here:

1. We can't switch with `auto`, since it selects an exceptionally old container with a version of GCC that doesn't play nicely by default with `tree-sitter`'s build. I could probably hack around this, but it's not worth it just to support a small number of EOL manylinux tags.
2. We can't upgrade all the way to `2_28`, since `maturin-action` appears to have trouble bootstrapping its toolchain with the system `pip` there: https://github.com/PyO3/maturin-action/issues/249. I could work around that by disabling `sccache`, but I'd really prefer not to.

woodruffw:
I'm declaring bankruptcy on making this work consistently, and limiting my scope of interest to the following platforms going forwards for wheel builds:

* macOS: ARM64 and AMD64
* Linux: musl with ARM64/Arm7/AMD64, glibc 2.28+ for ARM64/Arm7/AMD64. Some wheels may also be glibc 2.24+, but this will be dropped since 2.24 is pretty consistently EOL on distros.
* Windows: AMD64 and x86

Anything else _might_ be supported again in the future, but will probably be contingent on a more reliable build setup + fewer pain points with our non-Rust dependencies (namely, `ring` and `tree-sitter`).

PR #615, opened 2025-03-18
Clearly state that `actions: read` is only required for private repos

### LINKED ISSUE #608: Possible Error in Documentation: `actions: read` Permission May Not Be Required for Using `zizmor` in GitHub Actions
[The documentation](https://woodruffw.github.io/zizmor/usage/#use-in-github-actions) states that `zizmor` requires `actions: read` permission when used in GitHub Actions. However, testing suggests that this permission is not actually needed. Additionally, the [GitHub API documentation](https://docs.github.com/en/rest/authentication/permissions-required-for-github-apps?apiVersion=2022-11-28#repository-permissions-for-actions) does not indicate that this permission is necessary.  

Could you confirm whether this requirement is indeed needed, and update the documentation accordingly if it's incorrect?

vivodi:
fix #608

### DIFF
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -404,9 +404,8 @@ jobs:
     runs-on: ubuntu-latest
     permissions:
       security-events: write
-      # required for workflows in private repositories
-      contents: read
-      actions: read
+      contents: read # only required for workflows in private repositories
+      actions: read # only required for workflows in private repositories
     steps:
       - name: Checkout repository
         uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


docs/usage.md
```diff
@@ -404,9 +404,8 @@ jobs:
     runs-on: ubuntu-latest
     permissions:
       security-events: write
-      # required for workflows in private repositories
-      contents: read
-      actions: read
+      contents: read # only required for workflows in private repositories
+      actions: read # only required for workflows in private repositories
```
woodruffw, line 408:
```suggestion
      contents: read # only needed for private repos
      actions: read # only needed for private repos
```

woodruffw (approved):

PR #616, opened 2025-03-18
tests: add repro case for #612

### LINKED ISSUE #612: [BUG]: Ignoring results with inline comments on composite actions
### Pre-submission checks

- [x] I have looked through the [open issues](https://github.com/woodruffw/zizmor/issues?q=is%3Aissue+is%3Aopen+) for a duplicate report.

### Expected behavior

Ignore results within composite actions using inline comments.

### Actual behavior

Results are not ignored when inline comments are used within composite actions.

### Reproduction steps

On a composite action that raises `error[template-injection]` for example, adding `# zizmor: ignore[template-injection]` has no effect.

### Logs

```plain text

```

### Additional context

Kudos for the great work being done @woodruffw.

zizmor is being used to tighten up our actions here: https://github.com/ansys/actions/pull/725
>[!NOTE]
>I am opening this as a bug and not as a feature request because the documentation implies that inline comments should be supported within composite actions. [Although on this page](https://woodruffw.github.io/zizmor/configuration/), it is stated that a `zizmor.yml` config cannot be used. I stand to be corrected though, maybe I missed something. Thanks!

woodruffw:
~~WIP.~~

Things I need to do here:

* Add the concept of an "invisible" span, which is used for evaluating ignore comments but not presented in outputs like SARIF or the terminal render.
* Add invisible spans to audits where appropriate.

Fixes #612.

### DIFF
--- a/docs/development.md
+++ b/docs/development.md
@@ -302,12 +302,12 @@ different situations:
 * "Related" locations are additional locations that are related to the
   finding, but not as important as the primary location. A finding can
   have multiple related locations.
-* "Invisible" locations are used to mark a span as relevant to a finding,
+* "Hidden" locations are used to mark a span as relevant to a finding,
   but are not shown in outputs like SARIF or the cargo-style "plain"
   output. These are useful marking spans as included in a finding e.g.
   so that `# zizmor: ignore` works in intuitive places.
 
-In general, audit authors shouldn't need invisible locations at all.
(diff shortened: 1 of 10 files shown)


moe-ad:
Thanks for this @woodruffw.
👏

woodruffw:
No problem! Thanks for reporting!

PR #634, opened 2025-03-27
cli: add a "GitHub" output format

### LINKED ISSUE #633: Feature: format output with GitHub Actions workflow commands
### Pre-submission checks

GitHub supports annotating PRs with comments attached to relevant files/lines - for example, https://github.com/pytest-dev/pytest-github-actions-annotate-failures is a plugin that does this for pytest failures.

This would be neat to have for zizmor when someone runs it inside a GitHub action/workflow since it would visually point at the exact file and line that is causing issues.

It is pretty straightforward to implement - you can see how it's done in the above pytest plugin. At a glance, you already have all relevant information (filename, line number, comment). It can be turned on either with an optional --argument, or automatically detect that it's running within a workflow by inspecting os.environ as is done by pytest plugin: https://github.com/pytest-dev/pytest-github-actions-annotate-failures/blob/v0.2.0/pytest_github_actions_annotate_failures/plugin.py#L36

### Additional context

Official documentation for this output format is here: https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#example-setting-a-notice-message

woodruffw:
Closes #633.

### DIFF
--- a/.github/workflows/test-output.yml
+++ b/.github/workflows/test-output.yml
@@ -1,4 +1,4 @@
-name: Test SARIF Presentation
+name: Test output formats
 
 on:
   pull_request:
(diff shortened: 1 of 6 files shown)


woodruffw:
This works, but comes with some significant limitations that make me hesitant to merge it: GitHub Actions currently limits check annotations to 10 per step, or 50 per job (if multiple steps are producing annotations). As a result, any `zizmor` run that produces a nontrivial number of annotations will have some annotations dropped, probably in a first-come-first serve basis.

In practice, I think this makes check annotations unworkable/unusable in the general case, since users have to dive into the action logs to see the findings that aren't given a check annotation due to the limit.

More context here: https://github.com/orgs/community/discussions/26680, https://github.com/orgs/community/discussions/68471

woodruffw:
I'm not fully sure why, but GitHub's annotations also seem to be slightly buggy with spans that end at the end of input: 



The above should have three annotations (one at the very end of the file), but that last one doesn't render.

woodruffw:
This now works as expected:

Screenshot 2025-04-07 at 5 27 04 PM

The last thing to do here is document the limitations.

Issue #617, opened 2025-03-18
Feature: custom personas


davidmreed:
### Pre-submission checks

- [x] I am **not** reporting a bug (crash, false positive/negative, etc). These must be filed via the bug report template.
- [x] I have looked through the open issues for a duplicate request.

### What's the problem this feature will solve?

I'm interested in defining a custom persona. I believe this might supersede #435 in a way that's a little less uncomfortable than allowing generic ignores. It could also simplify or subsume #396.

An example driver would be the desire to include un-hashed references in the default persona, but not other pedantic findings, while reducing the severity of injection findings. This could represent the threat model of a private repository that is sensitive to supply-chain attacks but not to attacker-controlled workflow events.

### Describe the solution you'd like

Expand zizmor config to include a `personas:` section. Each persona should be able to identify which findings are "of interest" to that persona, such that they're emitted when that persona is selected but not otherwise.

It might also be valuable to allow the Persona to set the severity of an issue. Referencing the use case in #396, a "trusted" persona might omit some findings entirely, and reduce the severity of other findings while still including them.

davidmreed:
If this feature is of interest to the maintainers I would be up for contributing to the implementation.

woodruffw:
Hey @davidmreed, thanks for opening a feature request!

I think this is an interesting idea, and I agree with your appraisal about this being a potentially more generic, more palatable solution to both #396 and #435.

At the same time, there are some architectural challenges here:

* Personas are currently very static: there's a fixed enum, and each finding explicitly encodes its persona. This would need to become customizable, while still expressing a "default" persona for each finding.
* zizmor's config currently serves two purposes: it's both an in-repo config and a general config (shared across multiple repos), and people use it in both ways. This is causing some impedance mismatches, e.g. people want to audit multiple repos in a single `zizmor` invocation and are expecting one config to be loaded per repo, but `zizmor` really only loads the one it discovers in `$CWD`. Custom personas probably fit into the "general config" category (since having a custom persona doesn't seem super useful in just a single repo!), so a blocker for doing them in the config would be figuring out a "v2" config that splits these different use cases appropriately.

I'm curious if you have thoughts about these! One thing I've been thinking about is supporting "profiles" in addition to "personas," which would allow these kinds of fine-grained configurations/remappings while preserving the persona-driven defaults. But I haven't thought extensively about that yet (as well as its maintenance/complexity implications).

Issue #240, opened 2024-12-06
Figure out sub-field spanning


woodruffw:
#233 highlights a piece of non-ideal user confusion: zizmor operates at the layer of YAML element spans at the moment, meaning that many findings have "identical" spans because their actual findings are really substrings, etc within a YAML string.

For example, these two findings _look_ like duplicates, but are really slightly different spans within the same parent span:

```
info[template-injection]: code injection via template expansion
  --> /Users/wtan/Projects/zizmor/docs.yml:55:9
   |
55 |         - name: Update Docs Reference Section and Push Changes
   |           ---------------------------------------------------- info: this step
56 |           continue-on-error: true
57 |           run: |
   |  _________-
58 | |           python docs/build_reference.py
...  |
66 | |             echo "No changes to commit"
67 | |           fi
   | |____________- info: github.head_ref may expand into attacker-controllable code
   |
   = note: audit confidence → Low

info[template-injection]: code injection via template expansion
  --> /Users/wtan/Projects/zizmor/docs.yml:55:9
   |
55 |         - name: Update Docs Reference Section and Push Changes
   |           ---------------------------------------------------- info: this step
56 |           continue-on-error: true
57 |           run: |
   |  _________-
58 | |           python docs/build_reference.py
...  |
66 | |             echo "No changes to commit"
67 | |           fi
   | |____________- info: github.ref may expand into attacker-controllable code
   |
   = note: audit confidence → Low
```

I need to figure out a good way to represent and highlight these.

h/t @Ninja3047

woodruffw:
#945 does the first chunk of this, by making the expression APIs span-aware.

woodruffw:
#949 will lay out all of the internals for this, including using `bot-conditions` as a proof of concept. I'll roll it out to more audits with subsequent PRs.

PR #637, opened 2025-03-31
feat: Audit secrets outside an environment

aldur:


woodruffw:
Hey @aldur, thanks for taking a stab at this! I'll do a high-level review later today or tomorrow.

src/audit/secrets_outside_environment.rs
```diff
@@ -0,0 +1,80 @@
...
+
+impl Audit for SecretsOutsideEnvironment {
+    fn new(_state: super::AuditState) -> anyhow::Result<Self>
+    where
+        Self: Sized,
+    {
+        Ok(Self)
+    }
+
+    fn audit_raw<'w>(
```
woodruffw, line 25: Unless I'm missing something, I don't think you want `audit_raw` here -- that's only for low-level audits like `unredacted-secrets`.

Since you're operating purely on the YAML here and are focused on step-level findings, I think you want `audit_step` here.

tests/integration/test-data/secrets-outside-environment.yml
woodruffw, line 1: Could you add a snapshot test for this? See `snapshot.rs` for how it's done for other audits :slightly_smiling_face:

src/audit/secrets_outside_environment.rs
woodruffw: Or rather, `audit_normal_job`, since you're testing the job's `environment`.

src/audit/secrets_outside_environment.rs
```diff
@@ -0,0 +1,80 @@
...
+                            }
+                            StepBody::Run {
+                                run: _,
+                                shell: _,
+                                env,
+                                working_directory: _,
+                            } => match env {
+                                LoE::Expr(_) => {
+                                    // TODO: Implement this.
+                                    panic!("We don't handle Expr yet!")
```
woodruffw, line 54: Just noting this TODO :slightly_smiling_face: -- for reference, the `template-injection` or `unredacted-secrets` audits might be a good source for using the expression parsing APIs.

aldur: Thank you! Addressed in `1d5b2df`.

src/audit/secrets_outside_environment.rs
aldur: Thank you! I went for the `audit_step` and access the `environment` by going up the hierarchy. Addressed in `1d5b2df`.

aldur:
Thanks for the comments! I have addressed them, added a snapshot and had to modify `use-trusted-publishing.yml` in `6b963b6` for `cargo insta test` to pass. 

I have also some `.snap.new` files in my cwd, should I add them to this PR?

woodruffw:
> I have also some `.snap.new` files in my cwd, should I add them to this PR?

You should run `cargo insta test --review` and then accept them, and then check _those_ files in 🙂

aldur:
> > I have also some `.snap.new` files in my cwd, should I add them to this PR?
> 
> You should run `cargo insta test --review` and then accept them, and then check _those_ files in 🙂

Done, thank you!

woodruffw:
Thanks @aldur, this is looking really good! 

One thing I'm realizing, based on the changes to the other tests: this might be slightly too conservative as a default audit, at least for now, since a lot of people have relatively simple CI/CD setups with 1-2 workflows where an environment wouldn't provide much isolation.

Given that, I'm thinking that it might make sense to mark this as a "pedantic" audit for now, i.e. by calling `.persona(Persona::Pedantic)` on each finding produced. How would you feel about that? IMO that would allow people to opt into the behavior here while not being too disruptive, and we could make it a "regular" audit in the future (or make it enableable in the regular persona once #617 is fleshed out more).

aldur:
> Thanks @aldur, this is looking really good!
> 
> One thing I'm realizing, based on the changes to the other tests: this might be slightly too conservative as a default audit, at least for now, since a lot of people have relatively simple CI/CD setups with 1-2 workflows where an environment wouldn't provide much isolation.
> 
> Given that, I'm thinking that it might make sense to mark this as a "pedantic" audit for now, i.e. by calling `.persona(Persona::Pedantic)` on each finding produced. How would you feel about that? IMO that would allow people to opt into the behavior here while not being too disruptive, and we could make it a "regular" audit in the future (or make it enableable in the regular persona once #617 is fleshed out more).

Done! Sorry it took me some time to get back to this :)

woodruffw:
> Done! Sorry it took me some time to get back to this :)

No problem at all! I'll do another review tonight.

src/audit/secrets_outside_environment.rs
```diff
@@ -0,0 +1,83 @@
...
+};
+
+use super::{Audit, audit_meta};
+use crate::{finding::Confidence, models::StepCommon, AuditState, AuditLoadError, Persona};
+
+pub(crate) struct SecretsOutsideEnvironment;
+
+audit_meta!(
+    SecretsOutsideEnvironment,
+    "secrets-outside-environment",
```
woodruffw, line 13: I'm trying to make new audit names a bit shorter, so some ideas here:

1. `secret-outside-env`
2. `secret-without-env`
3. `secret-missing-env`
4. `secret-no-env` (shortest, but maybe not easy to understand)

Curious if you have any other naming thoughts as well :slightly_smiling_face:

src/audit/secrets_outside_environment.rs
```diff
@@ -0,0 +1,83 @@
...
+    }
+}
+
+impl SecretsOutsideEnvironment {
+    fn check_secrets_access<'w>(
+        s: &str,
+        step: &crate::models::Step<'w>,
+        findings: &mut Vec<crate::finding::Finding<'w>>,
+    ) -> anyhow::Result<()> {
+        if s.contains("secrets") {
```
woodruffw, line 70: I *think* we need to exclude `secrets.github_token` here, since that's a "latent" secret that's present regardless of any environment.

src/audit/secrets_outside_environment.rs
```diff
@@ -0,0 +1,83 @@
...
+                eenv = with;
+            }
+            StepBody::Run {
+                run: _,
+                shell: _,
+                env,
+                working_directory: _,
+            } => match env {
+                LoE::Expr(e) => {
+                    Self::check_secrets_access(e.as_bare(), step, &mut findings)?;
```
woodruffw, line 47: I think this check needs to be a little more discerning: `e` needs to be parsed into an `Expr`, which then needs to be walked recursively to visit each `Expr::Context` and match _that_ on the `secrets` namespace. 

Without that, we'll end up with (admittedly goofy) false positives like:

```yaml
with:
  yolo: ${{ 'secrets' }}
```

Writing the AST walker is currently a bit annoying/manual; if you want to wait until I add a `Visitor` API that might make your life a bit easier (I plan on doing that pretty soon) :slightly_smiling_face:

src/audit/secrets_outside_environment.rs
```diff
@@ -0,0 +1,83 @@
...
+            return Ok(findings);
+        }
+
+        let eenv: &Env;
+        match &step.body {
+            StepBody::Uses { uses: _, with } => {
+                eenv = with;
+            }
+            StepBody::Run {
+                run: _,
```
woodruffw, line 41: Note: I'm okay with skipping `run:` blocks for an MVP here, but we _will_ need to extract expressions from those and scan for secrets as well at some point.

src/audit/secrets_outside_environment.rs
aldur: Picked number 2 and fixed in `96c90ac`.

src/audit/secrets_outside_environment.rs
aldur: Yep that makes sense! Fixed it [here](b5dab62).

src/audit/secrets_outside_environment.rs
aldur: I like that! It will also increase the audit confidence. I am ok with waiting for the `Visitor` API, just give me a ping once ready and I'll put it to work :)

src/audit/secrets_outside_environment.rs
aldur: Makes sense! I naively thought that wouldn't be required, would this be an example?

```
- name: Clean install dependencies and build
  run: |
    npm ci
    PASSWORD=${{ secrets.password}} npm run build
```

aldur:
Hi there! Any news on the `Visitor` API? I can give this another shot and get it merged.

woodruffw:
Hey @aldur, sorry for the delay! 

The `Visitor` API is high on my stack -- it got deferred a bit because of some larger expression API refactoring that needed to be done to enable #240, but now that that's done I'll be working on it. So you should be unblocked here shortly.

(P.S. Sorry for the huge conflict set here -- with any luck most of it can probably be discarded and the snapshot tests can be re-done on top of just the audit changes.)

src/audit/secrets_outside_environment.rs
woodruffw: Yep, exactly 🙂

woodruffw:
I'm done with the expression API refactoring, but I wasn't able to reach a `Visitor` API design that made a lot of sense -- all of the patterns I tried didn't map super well to our existing use patterns, or could be expressed more easily/cleanly with the existing "match and recurse" style.

Given that, I think we can move forward without blocking on any larger visitor API changes here; the existing APIs should work well enough. Sorry for such a delay on my end.

aldur:
Sorry I could never get to this, and thank you! :)

PR #644, opened 2025-04-03
bugfix: cache-poisoning: fix false positive for docker/setup-buildx-action

### LINKED ISSUE #642: [BUG]: setup-buildx-action false positive for cache poisoning audit
Hi, maintainer of Docker GitHub Actions here.

The following check is not quite right:

https://github.com/woodruffw/zizmor/blob/ccd4178eb5276bd0bd17f2e814a921959ff67517/src/audit/cache_poisoning.rs#L109-L114

Even if `cache-binary` is `true`, it will not write or read from cache if `version` is not set. By default it will use buildx binary available on GitHub runner: https://github.com/actions/runner-images/blame/main/images/ubuntu/Ubuntu2404-Readme.md#L74

I think this should check for both `version` input being set and `cache-binary` set to `true`.

woodruffw:
This is an initial stab at exposing greater generality when evaluating whether a set of controls are "satisfied" for the purpose of considering an action enabled.

Doing this while also preserving the concept of "default enablement" results in some very wonky 4-valued logic, since there are 4 different states a control can evaluate to:

1. Satisfied by default
2. Satisfied through evaluation
3. Not satisfied through evaluation
4. "Conditional," i.e. we don't know because it's driven by an expression or similar

This is a little hard to follow, so I need to think about this more before actually merging it. However, the tests _do_ pass and I've also added a successful test for the original bug case that spurred this.

Fixes #642.

### DIFF
--- a/tests/integration/snapshots/integration__snapshot__cache_poisoning-15.snap
+++ b/tests/integration/snapshots/integration__snapshot__cache_poisoning-15.snap
@@ -1,0 +1,20 @@
+---
+source: tests/integration/snapshot.rs
+expression: "zizmor().input(input_under_test(\"cache-poisoning/issue-642-repro.yml\")).run()?"
+snapshot_kind: text
+---
+error[cache-poisoning]: runtime artifacts potentially vulnerable to a cache poisoning attack
+  --> @@INPUT@@:20:9
+   |
+15 | /         with:
+16 | |           cache-binary: true
+17 | |           version: latest
+   | |_________________________^ opt-in for caching here
+18 |
+19 |         - name: Build docker
+20 |           uses: docker/build-push-action@48aba3b46d1b1fec4febb7c5d0c644b249a11355
+   |           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ runtime artifacts usually published here
+   |
+   = note: audit confidence → Low
+
+1 finding: 0 unknown, 0 informational, 0 low, 0 medium, 1 high
--- a/docs/release-notes.md
+++ b/docs/release-notes.md
@@ -17,6 +17,8 @@ of `zizmor`.
   to determine whether contexts actually expand in an unsafe manner,
   making it significantly more accurate (#640)
 * The [cache-poisoning] audit is now aware of @jdx/mise-action (#645)
+* The [cache-poisoning] audit is now significantly more accurate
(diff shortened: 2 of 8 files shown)


tests/integration/snapshots/integration__snapshot__cache_poisoning-15.snap
```diff
@@ -0,0 +1,20 @@
...
+   |
+15 | /         with:
+16 | |           cache-binary: true
+17 | |           version: latest
+   | |_________________________^ opt-in for caching here
+18 |
+19 |         - name: Build docker
+20 |           uses: docker/build-push-action@48aba3b46d1b1fec4febb7c5d0c644b249a11355
+   |           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ runtime artifacts usually published here
+   |
```
crazy-max, line 17: Indeed we are publishing the build record as GitHub artifact in post step of the `build-push-action`: https://github.com/docker/build-push-action/blob/84ad562665bb303b549fec655d1b64f9945f3f91/src/main.ts#L208-L214

But there could false positives. If env vars `DOCKER_BUILD_SUMMARY=false` or `DOCKER_BUILD_RECORD_UPLOAD=false` are set, then no artifact is uploaded: https://github.com/docker/build-push-action/tree/master?tab=readme-ov-file#environment-variables

There are also other tricky cases where build summary would be disabled and therefore no artifact will be uploaded: https://github.com/docker/build-push-action/blob/84ad562665bb303b549fec655d1b64f9945f3f91/src/main.ts#L170-L180 but I think these are edge cases.

Env var could also be tricky to detect in some cases like internal script using `core.exportVariable` or append to `GITHUB_ENV`.

crazy-max (commented):
Nice work thanks!

Left a comment about runtime artifacts

tests/integration/snapshots/integration__snapshot__cache_poisoning-15.snap
woodruffw: Thanks for calling this out! Yeah, I don't have a fantastic way to model implicit controls like environment variables quite yet -- it should be doable in the same basic way, but I need to think more through the complexity implications 🙂

In the mean time, this is one of the reasons this finding is marked as "low confidence" -- we only issue high-confidence cache poisoning findings when the workflow trigger itself is unambiguously a publish trigger, i.e. tag or release.

PR #657, opened 2025-04-08
feat: add JSON format versioning

woodruffw:
This makes `--format=json` an alias for `json-v1`, giving me the flexibility to build a v2 JSON format at some point in the future. Also documents the compat policy for JSON format versions.


woodruffw:
xref https://github.com/clap-rs/clap/pull/5480 for why aliasing is done with a variant instead of with an alias value.

Issue #646, opened 2025-04-05
Feature: informative error message


rost-andreev:
### Pre-submission checks

- [x] I am **not** reporting a bug (crash, false positive/negative, etc). These must be filed via the bug report template.
- [x] I have looked through the open issues for a duplicate request.

### What's the problem this feature will solve?

When running zizmor on a badly formatted workflow file, the program exits with an error code without giving much explanation regarding what went wrong and how to fix the issue.

For instance, consider the following workflow file:
```yaml
name: Test

on:
  workflow_call:
    inputs:
      my-input:
        descripton: My input
    secrets:
      my-secret:
        description: My secret

jobs:
  job:
    runs-on: ubuntu-latest
    steps:
      - run: echo ${{ inputs.my-input }}
```
This workflow presents 3 issues:
1. `my-input.descripton` has a typo, i.e, invalid property
2. `my-input.type` property is missing
3. `my-secret.required` property is missing

When running zizmor on this workflow, the program prints the following output:
```
failed to register input: ./.github/workflows/workflow.yaml

Caused by:
    0: invalid GitHub Actions workflow: file://./.github/workflows/workflow.yaml
       
       Caused by:
           data did not match any variant of untagged enum Trigger
    1: invalid GitHub Actions definition: file://./.github/workflows/workflow.yaml
       
       Caused by:
           missing field `runs`
    2: failed to register input as workflow or action
```
Zizmor attempted to unmarshall the file into a predefined Workflow structure which expected the file to follow a certain pattern and contain some fields that were missing, which in turn caused the program to exit with an error message without explicitly stating the issue itself.

### Describe the solution you'd like

When running into such errors, zizmor could attempt to provide more information about the encountered formatting issues. One solution could be validating the workflow file with the [GitHub Workflow schema](https://json.schemastore.org/github-workflow.json). For example, take the following code:
```rust
fn main() {
    let workflow = include_str!("../workflow.yaml");
    let workflow: serde_json::Value = serde_yaml::from_str(workflow).unwrap();

    let schema = include_str!("../github-workflow.json");
    let schema: serde_json::Value = serde_json::from_str(schema).unwrap();

    let validator = jsonschema::validator_for(&schema).unwrap();
    let output = validator.apply(&workflow).basic();
    match output {
        jsonschema::BasicOutput::Valid(_) => println!("ok"),
        jsonschema::BasicOutput::Invalid(errors) => {
            for error in errors {
                println!(
                    "Error: {}: {}",
[...]

PR #662, opened 2025-04-13
feat: more informative error message

reandreev:
Addresses #646

This PR adds a workflow validation step after a failed unmarshaling attempt in order to find what the concrete issues with the workflows are. Running zizmor on the following workflow:
```yaml
name: Invalid

boom:

on:
  workflow_call:
    inputs:
      input:
        description: Input

permissions: {}

jobs:
  valid:
    runs-on: ubuntu-latest
    steps:
      - run: echo 'invalid'
```

will produce the following output:
```
failed to register input: .github/workflows/invalid.yaml

Caused by:
    0: invalid GitHub Actions workflow: file://.github/workflows/invalid.yaml
       
       Caused by:
           on.workflow_call.inputs.input: "type" is a required property
           Additional properties are not allowed ('boom' was unexpected)
           
    1: invalid GitHub Actions definition: file://.github/workflows/invalid.yaml
       
       Caused by:
           missing field `runs`
    2: failed to register input as workflow or action
```

An interesting thing I've noticed is that running zizmor on the following workflow:
```yaml
name: Invalid

boom:

on:
  workflow_call:

permissions: {}

jobs:
  valid:
    runs-on: ubuntu-latest
    steps:
      - run: echo 'invalid'
```

will succeed. Even though it has that additional `boom` property at the top level, zizmor still managed to unmarshal all of the expected properties correctly since those were properly defined and simply ignored the invalid one. Not sure if this is an expected behavior, but thought I'd mention it.

Regarding the validator instantiation, another approach would be to keep the GitHub Workflow schema as a dependency of the tool and instead do something like:
```rust
static WORKFLOW_VALIDATOR: LazyLock<Option<Validator>> = LazyLock::new(||
    match serde_json::from_str(include_str!("../github-workflow.json")) {
        Ok(schema) => validator_for(&schema).ok(),
        Err(_) => None,
    }
); 
```
Considering how seldom the schema itself gets updated that could also be an approach to consider. Pros: the tool would be more self-contained, cons: the tool would need to be recompiled each time the schema gets updated.

Some other things to consider:
1. In the example above the error messages could be sorted by their depth/nesting and/or the order of appearance within the file itself
2. More line numbers could be added
3. The schema could be cached somewhere instead of retrieving it on every run


woodruffw:
Hi @reandreev, thanks for looking into this! I'll do a full review in the coming days.

> Even though it has that additional `boom` property at the top level, zizmor still managed to unmarshal all of the expected properties correctly since those were properly defined and simply ignored the invalid one. Not sure if this is an expected behavior, but thought I'd mention it.

Yep, this is semi-unintentional behavior -- see https://github.com/woodruffw/github-actions-models/issues/33 for some context :slightly_smiling_face:

> Regarding the validator instantiation, another approach would be to keep the GitHub Workflow schema as a dependency of the tool and instead do something like:
> 
> ```rust
> static WORKFLOW_VALIDATOR: LazyLock<Option<Validator>> = LazyLock::new(||
>     match serde_json::from_str(include_str!("../github-workflow.json")) {
>         Ok(schema) => validator_for(&schema).ok(),
>         Err(_) => None,
>     }
> ); 
> ```

Yes, let's do this! IMO a network request in an error path like this isn't ideal; it'd be a lot faster and more reliable to bake the schema into the binary (and I agree that it won't change frequently).

src/utils.rs
```diff
@@ -1,15 +1,62 @@
...
+};
+use std::{collections::VecDeque, ops::Range};
+use std::{fmt::Write, sync::LazyLock};
 
 use crate::audit::AuditInput;
 
+static WORKFLOW_VALIDATOR: LazyLock<Option<Validator>> = LazyLock::new(
+    || match serde_json::from_str(include_str!("../github-workflow.json")) {
+        Ok(schema) => validator_for(&schema).ok(),
+        Err(_) => None,
```
woodruffw, line 23: Perhaps we could make this externally infallible? IMO it's okay if we unwrap or panic here, since we'll assume that the workflow schema gets loaded correctly. That way we don't need to propagate the `Option` through `validate_workflow` as well :slightly_smiling_face:

src/models.rs
```diff
@@ -94,8 +94,19 @@ impl Deref for Workflow {
...
+                Some(validation_errors) => {
+                    return Err(validation_errors)
+                        .with_context(|| format!("invalid GitHub Actions workflow: {key}"));
+                }
+                None => {
+                    return Err(error)
+                        .with_context(|| format!("invalid GitHub Actions workflow: {key}"));
+                }
+            },
+        };
```
woodruffw, line 109: Let's do the same for `Action::from_string` too, with <https://json.schemastore.org/github-action.json> :slightly_smiling_face:

github-workflow.json
woodruffw, line 1: No need to do it with this PR, but a thought: we'll probably want some kind of scheduled workflow that checks periodically for updates to this file.

(That concerns the CI though, so I'll do that in a followup.)

src/models.rs
reandreev: Added 🙂

src/utils.rs
reandreev: Done

src/utils.rs
```diff
@@ -1,15 +1,70 @@
...
+});
+
+static WORKFLOW_VALIDATOR: LazyLock<Validator> = LazyLock::new(|| {
+    validator_for(&serde_json::from_str(include_str!("../github-workflow.json")).unwrap()).unwrap()
+});
+
+pub(crate) fn validate_action(contents: String) -> Error {
+    match serde_yaml::from_str(&contents) {
+        Ok(workflow) => match ACTION_VALIDATOR.apply(&workflow).basic() {
+            Valid(_) => anyhow!("valid action but failed unmarshaling"),
```
reandreev, line 31: Not really sure how to handle this case since in theory this should be unreachable (?), maybe you have some suggestion on what to do or what error message to print out here?

src/utils.rs
```diff
@@ -1,15 +1,70 @@
...
+            Invalid(errors) => parse_validation_errors(errors),
+        },
+        Err(_) => anyhow!("invalid yaml"),
+    }
+}
+
+pub(crate) fn validate_workflow(contents: String) -> Error {
+    match serde_yaml::from_str(&contents) {
+        Ok(workflow) => match WORKFLOW_VALIDATOR.apply(&workflow).basic() {
+            Valid(_) => anyhow!("valid workflow but failed unmarshaling"),
```
reandreev, line 41: Same here

src/utils.rs
woodruffw: `unreachable` makes sense to me here 🙂 

```suggestion
            Valid(_) => unreachable!("valid action but invalid YAML"),
```

src/utils.rs
woodruffw:
```suggestion
            Valid(_) => unreachab;e!("valid workflow but invalid YAML"),
```

src/utils.rs
woodruffw: This has the side effect of making this a panic, however, so the tests below should probably become integration snapshots instead -- see <https://woodruffw.github.io/zizmor/development/#writing-snapshot-tests> and <https://github.com/woodruffw/zizmor/blob/ee6f160a304b96fbde2b6fe346e71218c1f07bf0/tests/integration/e2e.rs#L184-L195> for some docs/examples 🙂

reandreev: I think maybe the tests can remain as unit tests by attempting to capture the panic instead, what do you think? For example:
https://github.com/woodruffw/zizmor/blob/f642eaf19a95534c42bf68169156eaf995a19261/src/utils.rs#L387-L389
Either way I did have to update a snapshot there so thanks for reminding me 🙂
Also, regarding your suggestion `"valid action but invalid YAML"`, this unreachable snippet is only reached when there is a discrepancy between the schema validation and the unmarshaling logic (which in theory shouldn't happen), I'm not sure if the YAML is guaranteed to be invalid in those cases? 🤔

woodruffw: > Also, regarding your suggestion "valid action but invalid YAML", this unreachable snippet is only reached when there is a discrepancy between the schema validation and the unmarshaling logic (which in theory shouldn't happen), I'm not sure if the YAML is guaranteed to be invalid in those cases? 🤔

Oh yeah, I see what you mean (and I misunderstood the context for this error) -- yeah, this should probably instead indicate to the user that the bug is on `zizmor`'s modeling side, rather than an invalid workflow/action definition per the schema. Sorry for the confusion!

(This also undoes the need for the panic handling, so sorry for the wild goose chase there.)

woodruffw:
Thanks @reandreev! I went ahead and landed this with #719, and credited you in the release notes.

PR #663, opened 2025-04-13
feat: rewrite unpinned-uses, fold in forbidden-uses

woodruffw:
This rewrites `unpinned-uses` on top of the core idea of #574.

TODOs:

- [x] Snapshot tests, including config files
- [x] Documentation

Closes #574. See #558.


woodruffw:
This is good to go functionally, although I just realized it's missing a decent chunk of #558: it allows for various pinning policies, but doesn't provide a basic denylist capability.

Given that, I suppose `zizmor` does need some variant of a dedicated `forbidden-uses` audit. I'm going to merge here since this adds a lot of the needed scaffolding, then rebuild on top of #574.

PR #670, opened 2025-04-15
feat: generalize RepositoryUsesPattern

### LINKED ISSUE #669: Rethink `uses` patterns
Filing for myself so I don't forget.

Starting with #663, configuration files expose "patterns" for matching `uses` clauses. These patterns are of the following forms:

* `*`: match anything (matches `foo/bar` and `abc/def`)
* `owner/*`: match anything under `@owner` on GitHub  (matches `owner/foo` and `owner/bar` but not `joe/schmoe`)
* `owner/repo`: match any path of `owner/repo` (matches `owner/repo` and `owner/repo/action` but not `owner/other` or `owner/other/action`).

That last one is a little wonky: it behaves *like* `owner/repo/*` in terms of matching subpaths, but it *looks* like it only matches `owner/repo` exactly.

I'm not sure whether this is the right behavior or not:

1. Having to write both `owner/repo` and `owner/repo/*` when specifying rules/policies is very confusing! It's much more intuitive to just have one `owner/repo` pattern that matches both.
2. On the other hand, the difference between `owner/repo/foo` and `owner/repo/bar` *does* matter when hash-pinning and forbidding actions: `owner/repo/foo@hash` might be safe, while `owner/repo/bar@hash` might be compromised or insecure. Users should be able to express this.

The happy middle ground here might be:

1. `owner/repo` matches exactly `owner/repo`
2. `owner/repo/*` matches any subpath of `owner/repo`, including the empty subpath. In other words, `owner/repo/*` matches `owner/repo`, so users who don't care about subpaths at all can always use it and get the "intuitive" behavior.

refs: #664 #663

woodruffw:
Having done this, I now feel less confident that this makes sense vs. just exposing a `fnmatch` compatible pattern syntax. However that would probably be too flexible.

TODO: 

- [x] Update docs
- [x] More snapshot tests

Fixes #669.

### DIFF
--- a/docs/audits.md
+++ b/docs/audits.md
@@ -153,80 +153,80 @@ In practice, this means that workflows should almost always set
 `#!yaml permissions: {}` at the workflow level to disable all permissions
 by default, and then set specific job-level permissions as needed.
 
-For example:
+!!! example
 
-=== "Before :warning:"
+    === "Before :warning:"
 
-    ```yaml title="excessive-permissions.yml" hl_lines="8-9"
-    on:
-      release:
-        types:
-          - published
+        ```yaml title="excessive-permissions.yml" hl_lines="8-9"
+        on:
+          release:
+            types:
+              - published
 
-    name: release
+        name: release
 
-    permissions:
-      id-token: write # trusted publishing + attestations
-
-    jobs:
-      build:
-        name: Build distributions 📦
-        runs-on: ubuntu-latest
-        steps:
-          - # omitted for brevity
-
-      publish:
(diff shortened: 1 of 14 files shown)


woodruffw:
It turns out `uses:` clauses are shockingly malleable: things like `uses: foo/bar/./../bar/./@v1` are perfectly valid and resolve "correctly," probably because it's all just system paths under the hood.

This makes it hard to write a pattern scheme that'll catch all the different possible ways to write equivalent `uses:` clauses. Rather than trying to do that, I'll probably add an `obfuscated-uses` or similar audit that blanket-catches these.

PR #687, opened 2025-04-22
chore(docs): clarify default rule in explicit config

woodruffw:
h/t @Bo98


Bo98 (approved):
Thanks!

Maybe also worth a callout around line 598 of something like "specifying a configuration overrides the above default policy".

PR #688, opened 2025-04-22
docs: recommend GitHubSecurityLab/actions-permissions

Marcono1234:
Recommends https://github.com/GitHubSecurityLab/actions-permissions to make it easier for users to find the minimally required permissions. That might be easier for users than pure trial and error, or digging through the GitHub API documentation to find the needed permissions.

If you are aware of other tools which can do this, maybe it would be worth mentioning them as well.

:warning: That action is currently still in beta and has some [limitations](https://github.com/GitHubSecurityLab/actions-permissions/tree/main/monitor#known-limitations), but maybe it is useful nonetheless.

### DIFF
--- a/docs/audits.md
+++ b/docs/audits.md
@@ -153,8 +153,10 @@ In practice, this means that workflows should almost always set
 `#!yaml permissions: {}` at the workflow level to disable all permissions
 by default, and then set specific job-level permissions as needed.
 
-Tools such as [GitHubSecurityLab/actions-permissions](https://github.com/GitHubSecurityLab/actions-permissions)
-can help finding the minimally required permissions.
+!!! tip
+
+    @GitHubSecurityLab/actions-permissions can help find the minimally required
+    permissions.
 
 !!! example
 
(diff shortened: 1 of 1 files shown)


woodruffw:
Nice, TIL about this action! Reviewing now.

docs/audits.md
```diff
@@ -153,6 +153,9 @@ In practice, this means that workflows should almost always set
 `#!yaml permissions: {}` at the workflow level to disable all permissions
 by default, and then set specific job-level permissions as needed.
 
+Tools such as [GitHubSecurityLab/actions-permissions](https://github.com/GitHubSecurityLab/actions-permissions)
```
woodruffw, line 156: This should work 🙂 

```suggestion
Tools such as @GitHubSecurityLab/actions-permissions
```

woodruffw, line 156: Nitpick, nonblocking: if you wrap this in a `!!! tip` it might be a bit more prominent/easy to read. But I'll leave it at your discretion.

woodruffw (commented):
One comment, otherwise looks great! Thanks @Marcono1234!

woodruffw (approved):

docs/audits.md
Marcono1234: Thanks for making these documentation changes!

PR #692, opened 2025-04-24
docs: extend remediation for dangerous-triggers

### LINKED ISSUE #684: Feature: Audit for `pull_request_target` used for Dependabot & Renovate auto approve / merge
### Pre-submission checks

Relates to #170

Users sometimes write workflows to automatically approve or merge Dependabot and Renovate PRs. To my understanding `pull_request_target` is not actually needed for these cases because Dependabot and Renovate create branches within the same repository and therefore `on: pull_request` should suffice to get a privileged token (unlike malicious PRs from forks which don't get a privileged token).

Unfortunately it seems some webpages, including [older versions of the GitHub documentation](https://github.com/github/docs/blob/9ea839afd536cd9dc2304a31fe175a1261fa8d5f/content/code-security/supply-chain-security/keeping-your-dependencies-updated-automatically/automating-dependabot-with-github-actions.md#approve-a-pull-request) (but not the [latest one](https://docs.github.com/en/code-security/dependabot/working-with-dependabot/automating-dependabot-with-github-actions#automatically-approving-a-pull-request)) recommended using `pull_request_target`.

Maybe either add a new audit rule, or extend the existing `dangerous-triggers` one, to report workflows which check if the author is Dependabot or Renovate, and where it is therefore likely that `pull_request_target` is not needed and `pull_request` suffices (if users explicitly specify `permissions`, see [GitHub documentation](https://docs.github.com/en/code-security/dependabot/troubleshooting-dependabot/troubleshooting-dependabot-on-github-actions#changing-github_token-permissions)).

(This would act as additional defense in combination with #170)

### Additional context

_No response_

Marcono1234:
Resolves #684
See also https://github.com/woodruffw/zizmor/issues/680#issuecomment-2818926648

Maybe the proposed documentation change here is a bit verbose. Feel free to adjust it in any way you like, or let me know if I should change something.

### DIFF
--- a/docs/audits.md
+++ b/docs/audits.md
@@ -115,12 +115,29 @@ Some general pointers:
   re-tooling the workflow to be a [reusable workflow].
 * Replace `pull_request_target` with `pull_request`, unless you *absolutely*
   need repository write permissions (e.g. to leave a comment or make
-  other changes to the upstream repo).
+  other changes to the upstream repo).\
+  `pull_request_target` is only needed to perform privileged actions on
+  pull requests on external forks. If you only expect pull requests from
+  branches within the same project, or if you are fine with some functionality
+  not working for external pull requests, prefer `pull_request`.
+* Automation for Dependabot pull requests can be implemented using `pull_request`,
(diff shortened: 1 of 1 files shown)


woodruffw (approved):
Thanks @Marcono1234! I did a bit of reflowing, but this looks great.

PR #696, opened 2025-04-25
bugfix: fix e2e test failures on Windows

Marcono1234:
- `e2e::invalid_config_file`
Was previously failing because `/dev/null` does not exist on Windows. Have used `NUL` now for Windows.
    ```
    error: failed to load config: The system cannot find the path specified. (os error 3)
    ```
- `e2e::menagerie`
Was previously failing because the expected snapshot output contains the file path separator `/`, but on Windows it is `\`. Have added replacement logic.

Not sure if these changes are clean enough, or if they could be simplified. Feedback is appreciated!

`e2e::color_control_tty` and `e2e::progress_bar_tty` still fail for me because I don't have `unbuffer`. It seems there are [workarounds for Windows](https://stackoverflow.com/q/11516258) but I haven't tried this yet.

### DIFF
--- a/tests/integration/common.rs
+++ b/tests/integration/common.rs
@@ -1,4 +1,5 @@
 use anyhow::{Context as _, Result};
+use regex::{Captures, Regex};
 use std::{env::current_dir, io::ErrorKind};
 
 use assert_cmd::Command;
@@ -149,8 +150,16 @@ impl Zizmor {
             OutputMode::Both => [output.stderr, output.stdout].concat(),
         })?;
 
+        let input_placeholder = "@@INPUT@@";
+        let input_path_regex = Regex::new(&format!(r"{input_placeholder}[\\/\w.-]+"))?;
         for input in &self.inputs {
-            raw = raw.replace(input, "@@INPUT@@");
+            raw = raw.replace(input, input_placeholder);
+            // Normalize Windows '\' file paths to using '/'
+            raw = input_path_regex
+                .replace_all(&raw, |captures: &Captures| {
+                    captures.get(0).unwrap().as_str().replace("\\", "/")
+                })
+                .into_owned();
         }
 
         Ok(raw)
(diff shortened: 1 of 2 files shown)


tests/integration/e2e.rs
```diff
@@ -186,7 +186,7 @@ fn invalid_config_file() -> Result<()> {
     insta::assert_snapshot!(
         zizmor()
             .output(OutputMode::Stderr)
-            .config("/dev/null")
+            .config(if cfg!(windows) { "NUL" } else { "/dev/null" })
```
woodruffw, line 189: LGTM, thanks!

tests/integration/common.rs
```diff
@@ -149,8 +150,16 @@ impl Zizmor {
...
+        let input_path_regex = Regex::new(&format!(r"{input_placeholder}[\\/\w.-]+"))?;
         for input in &self.inputs {
-            raw = raw.replace(input, "@@INPUT@@");
+            raw = raw.replace(input, input_placeholder);
+            // Normalize Windows '\' file paths to using '/'
+            raw = input_path_regex
+                .replace_all(&raw, |captures: &Captures| {
+                    captures.get(0).unwrap().as_str().replace("\\", "/")
+                })
+                .into_owned();
```
woodruffw, line 162: Out of curiosity, which of the `e2e` tests is revealing this behavior?

Marcono1234: This one if I recall correctly:
https://github.com/woodruffw/zizmor/blob/d3b7eb86c15d17533d9776fd2a5e485459268f81/tests/integration/snapshots/integration__e2e__menagerie.snap#L10-L13

On Windows you will get something like (with `\`):
```
INFO audit: zizmor: 🌈 completed @@INPUT@@\.github\dummy-action-2\action.yml 
...
```

woodruffw: Got it, thanks!

I thought about it a bit, and I think we should put this in a a `cfg!` block that matches only Windows -- it probably doesn't affect anything anywhere else, but doing that will make it clearer to future contributors (including myself, who will forget) what's going on here 🙂

woodruffw (approved):
Thanks, looks great!

PR #697, opened 2025-04-25
feat: reduce code duplication for workflow step handling

Marcono1234:
I noticed that multiple audit rules were handling `audit_step` and `audit_composite_step`, but previously their code was duplicated.

Therefore I tried to deduplicate it now. As part of that I also tried to make the `use` declarations a bit more consistent.

Hopefully I did not overlook something. Feedback is appreciated!

### DIFF
--- a/src/audit/forbidden_uses.rs
+++ b/src/audit/forbidden_uses.rs
@@ -1,10 +1,10 @@
 use anyhow::{Context, anyhow};
 use github_actions_models::common::Uses;
 
-use super::{Audit, AuditLoadError, AuditState, Finding, Step, audit_meta};
-use crate::finding::{Confidence, Persona, Severity};
-use crate::models::CompositeStep;
+use super::{Audit, AuditLoadError, AuditState, audit_meta};
+use crate::finding::{Confidence, Finding, FindingBuilder, Persona, Severity};
 use crate::models::uses::RepositoryUsesPattern;
+use crate::models::{CompositeStep, Step, StepCommon};
 use serde::Deserialize;
 
 pub(crate) struct ForbiddenUses {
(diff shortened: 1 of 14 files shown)


src/audit/forbidden_uses.rs
```diff
@@ -72,41 +57,46 @@ impl Audit for ForbiddenUses {
...
+            return Err(AuditLoadError::Skip(anyhow!("audit not configured")));
+        };
+
+        Ok(Self { config })
+    }
+
+    fn audit_step<'w>(&self, step: &Step<'w>) -> anyhow::Result<Vec<Finding<'w>>> {
+        self.process_step(step)?
+            .into_iter()
+            .map(|f| f.build(step.workflow()))
```
Marcono1234, line 85: Ideally the `build` calls would also be done in the `process_step` method.

I tried to solve this by adding a `fn document(&self) -> &'s impl AsRef<yamlpath::Document>;` to the `StepCommon` trait (whose implementations then return `self.workflow()` respectively `self.action()`). But unfortunately I was running into lifetime issues then when trying to use that `document()` method.

If you want, try it yourself as well; maybe I am just missing Rust knowledge there.

woodruffw: Ah yeah, good thought. I'll review these changes as-is, but agreed that `process_step` should be doing the `FindingBuilder -> Finding` transform internally.

woodruffw:
Thanks @Marcono1234, these are great changes! One optional thing that could be done in this PR if you'd like, but otherwise this LGTM.

src/models.rs
```diff
@@ -564,6 +571,10 @@ impl<'s> StepCommon<'s> for Step<'s> {
         utils::env_is_static(name, &envs)
     }
 
+    fn uses(&self) -> Option<&common::Uses> {
+        self.uses()
+    }
```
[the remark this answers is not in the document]
Marcono1234: Have removed the methods from the struct impls now.

woodruffw (approved):
LGTM, thanks!

PR #703, opened 2025-04-28
feat: further reduce step handling code duplication

Marcono1234:
Follow-up for #697

I was able to implement this now by having the new `document()` method return `&'s yamlpath::Document`[^1] instead of `&'s impl AsRef<yamlpath::Document>`, since for the latter the Rust compiler wanted explicit lifetimes at the callsites of `document()`, and I could not get that to work.

(Though if you know a way to make this work with `impl AsRef<...>` or similar, that would probably be even better.)

[^1]: Actually using  `&'s yamlpath::Document` was suggested by Copilot, though it still need some manual work to make it compile. Maybe it just chose the more straight forward approach without really understanding the effects, and I was just so fixated on using `impl AsRef<...>`.

### DIFF
--- a/src/finding/mod.rs
+++ b/src/finding/mod.rs
@@ -251,12 +251,7 @@ impl<'w> SymbolicLocation<'w> {
     }
 
     /// Concretize this `SymbolicLocation`, consuming it in the process.
-    pub(crate) fn concretize(
-        self,
-        document: &'w impl AsRef<yamlpath::Document>,
-    ) -> Result<Location<'w>> {
-        let document = document.as_ref();
-
+    pub(crate) fn concretize(self, document: &'w yamlpath::Document) -> Result<Location<'w>> {
         // If we don't have a path into the workflow, all
         // we have is the workflow itself.
         let feature = if self.route.components.is_empty() {
@@ -524,6 +519,13 @@ impl<'w> FindingBuilder<'w> {
     }
 
     pub(crate) fn build(self, document: &'w impl AsRef<yamlpath::Document>) -> Result<Finding<'w>> {
+        self.build_with_document(document.as_ref())
+    }
+
+    pub(crate) fn build_with_document(
+        self,
+        document: &'w yamlpath::Document,
+    ) -> Result<Finding<'w>> {
         let mut locations = self
             .locations
             .iter()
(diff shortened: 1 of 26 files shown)


src/finding/mod.rs
```diff
@@ -524,6 +519,13 @@ impl<'w> FindingBuilder<'w> {
     }
 
     pub(crate) fn build(self, document: &'w impl AsRef<yamlpath::Document>) -> Result<Finding<'w>> {
+        self.build_with_document(document.as_ref())
+    }
+
+    pub(crate) fn build_with_document(
```
Marcono1234, line 525: Unfortunately there is this separate method needed now because apparently `&T` does not implement `AsRef<T>`. Not sure about this `build_with_document` document; suggestions are welcome.

woodruffw:
Thanks @Marcono1234! I'll take a look at this tonight.

src/finding/mod.rs
woodruffw: I thought about this a bit and added a new `AsDocument` trait -- this allows us to do what `AsRef` can't, since Rust won't let us do `impl<T: StepCommon> AsRef<str> for T` for coherence reasons.

AFAICT this gives us the best of both worlds: we don't need a `build`/`build_with_document` split _and_ we can have most things implement `AsDocument` to give us further flexibility.

src/models.rs
```diff
@@ -62,6 +62,19 @@ pub(crate) trait StepCommon<'s> {
...
+
+pub(crate) trait AsDocument<'a, 'doc> {
+    fn as_document(&'a self) -> &'doc yamlpath::Document;
+}
+
+impl<'a, 'doc, T: StepCommon<'doc>> AsDocument<'a, 'doc> for T {
+    fn as_document(&'a self) -> &'doc yamlpath::Document {
+        self.document()
+    }
 }
```
woodruffw, line 78: This PR's diff is now a mess thanks to me renaming things, but this is the core of the change: we have an `AsDocument` trait that `build(...)` now takes.

(`AsDocument` has a slightly wonky lifetime set, to allow us to express both pre-existing lifetime patterns: things like `Step` where the document outlives its container, and things like `Workflow` where the document's lifetime is coextensive with its container.)

woodruffw (approved):

woodruffw:
Thanks @Marcono1234!

PR #713, opened 2025-04-29
feat: new audit: stale-action-refs

Marcono1234:
Resolves #689

### DIFF
--- a/docs/audits.md
+++ b/docs/audits.md
@@ -1317,6 +1317,37 @@ Address the source of obfuscation by simplifying the expression,
                   repository: octocat/hello-world
         ```
 
+## `stale-action-refs`
+
+| Type     | Examples                | Introduced in | Works offline  | Enabled by default | Configurable |
+|----------|-------------------------|---------------|----------------|--------------------|--------------|
+| Workflow, Action  | N/A            | v1.7.0        | ❌            | ✅                | ❌          |
+
+Checks for `#!yaml uses:` clauses which pin an action using a SHA reference,
+but where that reference does not point to a Git tag.
+
+When using an action commit which is not a Git tag / release version, that commit
+might contain bugs or vulnerabilities which have not been publicly documented
+because they might have been fixed before the subsequent release. Additionally,
+because changelogs are usually only published for releases, it is difficult to
+tell which changes of the subsequent release the pinned commit includes.
+
+!!! note
+
+    This is a `--pedantic` only audit because the detected situation is not
+    a vulnerability per se. But it might be worth investigating which commit
+    the SHA reference points to, and why not a SHA reference pointing to a
+    Git tag is used.
+
+    Some action repositories use a "rolling release branch" strategy where
+    all commits on a certain branch are considered releases. In such a case
+    findings of this audit can likely be ignored.
+
+### Remediation
+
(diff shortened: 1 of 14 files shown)


tests/integration/common.rs
```diff
@@ -25,6 +25,7 @@ pub fn input_under_test(name: &str) -> String {
 
 pub enum OutputMode {
     Stdout,
+    #[allow(dead_code, reason = "currently not used by any integration test")]
     Stderr,
```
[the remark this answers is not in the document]
woodruffw: No worries, let's keep it here.

woodruffw:
Nice, thank you @Marcono1234! This looks great from a very quick initial pass; I'll do a more detailed review tonight.

docs/audits.md
```diff
@@ -1317,6 +1317,37 @@ Address the source of obfuscation by simplifying the expression,
...
+!!! note
+
+    This is a `--pedantic` only audit because the detected situation is not
+    a vulnerability per se. But it might be worth investigating which commit
+    the SHA reference points to, and why not a SHA reference pointing to a
+    Git tag is used.
+
+    Some action repositories use a "rolling release branch" strategy where
+    all commits on a certain branch are considered releases. In such a case
+    findings of this audit can likely be ignored.
```
woodruffw, line 1344: I was thinking a bit, and I think we _could_ detect this scenario by looking up the SHA ref, and then seeing whether it belongs to a branch other that the default one. But that's likely to expensive in terms of API calls and slow, so I'm OK with not doing it for now.

src/audit/stale_action_refs.rs
```diff
@@ -0,0 +1,92 @@
...
+            Some(commit_ref) => self
+                .client
+                .longest_tag_for_commit(&uses.owner, &uses.repo, commit_ref)
+                .with_context(|| {
+                    format!(
+                        "couldn't retrieve tag for {owner}/{repo}@{commit_ref}",
+                        owner = uses.owner,
+                        repo = uses.repo
+                    )
+                })?,
```
woodruffw, line 37: Nitpick: this is a great context message to add, but it's so good that it makes me realize that it should be in `longest_tag_for_commit` itself 🙂 -- could you move it there?

woodruffw (commented):
Thanks @Marcono1234, this is excellent! Only one tiny nitpick.

(Thinking out loud: I think we'll want to put the "mismatching comment" check under this same audit name, but that's not something that needs to be tackled in this PR at all.)

woodruffw (approved):

docs/audits.md
Marcono1234: It could also lead to false negatives when that branch is not a rolling release branch, but instead some work-in-progress branch containing bugs.

woodruffw: True. There's probably no good general solution here, then.

src/audit/stale_action_refs.rs
Marcono1234: Thanks for the compliment, but I just copied the message from `src/audit/known_vulnerable_actions.rs`, and _you_ were the one who added it there originally 😅

The `with_context` call there is now probably redundant; have created #729 to remove it.

woodruffw: 🤦

PR #716, opened 2025-04-29
test: update snapshot expressions

Marcono1234:
The data was outdated, but is apparently not causing a test failure. See also https://github.com/mitsuhiko/insta/issues/766

Also removes `snapshot_kind: text` which is apparently not emitted anymore because it is the default anyway (https://github.com/mitsuhiko/insta/pull/690).

Follow-up for #696, #698, and some earlier pull requests which had also changed the expressions


woodruffw:
Thanks! Yeah, I believe this doesn't cause any failures since `insta` doesn't check this in the normal path; only when updating an existing test.

woodruffw (approved):

Marcono1234:
insta has a `--force-update-snapshots` option; not sure if it would make sense to let the GitHub CI here run `cargo insta test --force-update-snapshots --all-features` (or similar) and then do a `git diff`, to make sure the snapshot outputs are all up to date.

PR #719, opened 2025-04-30
feat: more informative error message

### LINKED ISSUE #646: Feature: informative error message

woodruffw:
Recreation of #662.

Supersedes #662.
Closes #646.


src/utils.rs
```diff
@@ -1,15 +1,80 @@
...
+    let mut message = String::new();
+
+    for error in errors {
+        let description = error.error_description().to_string();
+        // HACK: error descriptions are sometimes a long rats' nest
+        // of JSON objects. We should render this in a palatable way
+        // but doing so is nontrivial, so we just skip them for now.
+        // NOTE: Experimentally, this seems to mostly happen when
+        // the error for an unmatched "oneOf", so these errors are
+        // typically less useful anyways.
```
Stranger6667, line 60: I'd be happy to provide a more structured output from the `jsonschema` side to improve the situation here - better handling for `oneOf` was [previously requested](https://github.com/Stranger6667/jsonschema/issues/649), but I didn't work on it yet.

Feel free to tag me on anything related to `jsonschema` crate changes, I am keen to make it more convenient to use :)

The current idea is to store the inner `oneOf` errors and make them accessible through the root `oneOf` error (at least users can have at least some level of introspection)

woodruffw: Thanks @Stranger6667, a more structured output here would be amazing! I can file an issue upstream in a bit demonstrating the current behavior (instead of just a `HACK` comment 😅).

(And thank you for making `jsonschema`!)

Stranger6667: Awesome! A concrete example would be very helpful as a part of an upstream issue :)

And thank you for your kind words

woodruffw: Opened https://github.com/Stranger6667/jsonschema/issues/739, let me know if there's any other details that would be helpful!

PR #728, opened 2025-04-30
doc: mention insta `--force-update-snapshots`

Marcono1234:
See https://github.com/woodruffw/zizmor/pull/716#issuecomment-2839617650


woodruffw (approved):
Thanks!

Issue #725, opened 2025-04-30
[BUG] `zizmor` aborts on empty workflow files


woodruffw:
### Discussed in https://github.com/woodruffw/zizmor/discussions/724

<div type='discussions-op-text'>

<sup>Originally posted by **jamesc-grafana** April 30, 2025</sup>
Hey, we're trying to use zizmor across our estate, and are finding that when it encounters an empty file the application panics, and the same is true when it encounters a private action. Should this be something that is reported as a finding rather than panicking and failing the whole run?</div>

---

See https://github.com/woodruffw/zizmor/discussions/724#discussioncomment-12994374 for full context.

woodruffw:
CC @jamesc-grafana for viz 🙂

jamesc-grafana:
@woodruffw I think I might have a fix for this specific one, I'll raise a PR

PR #732, opened 2025-05-01
bugfix: warn on parse errors instead of panic

jamesc-grafana:
Matches on the branch to add files to the input lists, and sends out a warning where the error branch is hit rather than panicking.

Works towards #725

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -281,9 +281,15 @@ fn collect_from_dir(
                 .parent()
                 .is_some_and(|dir| dir.ends_with(".github/workflows"))
         {
-            registry
+            match registry
                 .register_by_path(entry, Some(input_path))
-                .with_context(|| format!("failed to register input: {entry}"))?;
+                .with_context(|| format!("failed to register input: {entry}"))
+            {
+                Ok(_) => (),
+                Err(_) => {
+                    tracing::warn!("failed to register workflow: {entry}.")
+                }
+            };
         }
 
         if mode.actions()
@@ -291,7 +297,12 @@ fn collect_from_dir(
             && matches!(entry.file_name(), Some("action.yml" | "action.yaml"))
         {
             let action = Action::from_file(entry, Some(input_path))?;
-            registry.register_input(action.into())?;
+            match registry.register_input(action.into()) {
+                Ok(_) => (),
+                Err(_) => {
+                    tracing::warn!("failed to register action: {entry}")
+                }
+            };
         }
     }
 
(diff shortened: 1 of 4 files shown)


woodruffw:
Thanks for opening a PR @jamesc-grafana! This is a great start, but we need to handle two distinct cases here:

1. Invalid YAML -- these are okay to turn from errors into warnings, since they're malformed and GitHub itself will reject them. This covers your case, e.g. empty files 🙂 
2. Valid YAML but invalid workflow/action definitions -- these should remain errors, since they're sometimes bugs in `github-actions-models` itself (and I want users to notice and report these).

In effect I think this means the handling here needs to go down a layer: `Workflow::from_string` and `Action::from_string` probably need to do this internally so we can distinguish between these states. LMK if you're up to looking into that and I can give you some more pointers; otherwise I'm happy to attempt it later today or this weekend.

woodruffw:
> Thanks for the feedback, I see what you mean, this won't work for the github_api or explicit file spec right?

Yep, exactly! For consistency this needs to happen at `{Workflow,Action}::from_string`, since that's the common logic shared by all of the collection techniques.

> Unless I've missed something about how serde is deserializing the workflows/actions? It's possible I've missed something (Rust is not my daily language).

Yeah, there's a bit of nuance here: the YAML can be syntactically valid, but deserialization can still fail if the type (`T`) being deserialized into doesn't match the expected structure.

It's a little bit implicit in the current code because Rust figures out the types on its own, but this line:

https://github.com/woodruffw/zizmor/blob/4a9211a79ccef944803bbbf55f8b937dbdec728a/src/models.rs#L117

...actually returns a [`Workflow`](https://docs.rs/github-actions-models/latest/github_actions_models/workflow/struct.Workflow.html) type, which means that it fails if the YAML is valid but doesn't actually match that type. So we need to split the two kinds of errors apart to make sure that we still fail on the latter case.

Doing this is a little annoying, since `serde` doesn't always cleanly split between syntax and semantic errors. I think the right way to do this is to do `serde_yaml::from_str` into a `serde_yaml::Value` (which would catch all the syntax errors), followed by `serde_yaml::from_value` to convert the `Value` into a `Workflow` (which would catch the semantic errors).

In terms of returns, we could probably model that as `Result<Option<Self>>`, i.e. `Result` for a hard error and `Option` for a soft error. Or with a custom error type; I'd be okay either way 🙂

woodruffw:
No worries at all! I appreciate you looking into it; I'll see if I have time this even to make progress as well.

woodruffw:
I've opened #734 to experiment with a general solution to this; the TL;DR is that it adds an `InputError` type that distinguishes between the states here, which I'll be able to use to filter the different errors into fail vs warn groups.

Issue #704, opened 2025-04-28
New audit: `unpinned-images`


woodruffw:
Saving this as an idea.

This would be like `unpinned-uses`, but for `container:` blocks.

For example, this `semgrep` job should be flagged since the container (`semgrep/semgrep`) isn't pinned in any way:

```
jobs:
  semgrep:
    name: Scan
    runs-on: ubuntu-latest
    env:
      SEMGREP_APP_TOKEN: ${{ secrets.SEMGREP_APP_TOKEN }}
    container:
      image: semgrep/semgrep
    steps:
    - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
      with:
        persist-credentials: false
    - run: semgrep ci
```

woodruffw:
Possible names: `unpinned-containers`, `unpinned-images`, others?

trumant:
Will raise a PR for this one later today

woodruffw:
Thanks! FWIW there may be some `uses:` parsing logic that could be reused here (the Docker `uses:` parser, in particular.)

woodruffw:
Done with #733. Thanks a ton @trumant!

Follow-on ideas: #736, #737, #738

PR #733, opened 2025-05-01
feat: add unpinned-container-images check

trumant:
This check emits high confidence, high severity findings when jobs have container.image references that either fail to specify an image tag altogether or use a tag that is likely not static `latest`.

This updates #704 

## Feedback desired

- Should this be user configurable such that by default it issues finding when it sees `image: foo/bar` or `image: foo/bar:latest`, but support configuration that enforces SHA256 image refs only.
- A more sophisticated configuration could be to suppress findings for no tag and latest tag for certain specified docker registry hosts

Curious to hear thoughts from the community on the most reasonable functionality for the first version of the check.

I'd also be interested in feedback on the testing coverage here. I'm not entirely sure that I understand the best way to ensure this is fully and usefully integrated into the acceptance test suite.


trumant:
Leaving this in draft mode until I get some feedback on the desired version 1 scope/requirements.

woodruffw:
Thanks @trumant! I'll be able to review this tonight or tomorrow.

docs/audits.md
```diff
@@ -1348,6 +1348,31 @@ tell which changes of the subsequent release the pinned commit includes.
...
+
+```yaml
+container:
+  image: foo/bar:latest
+```
+
+### Remediation
+
+Change the `#!yaml container.image:` value to pin the container image to a specific, static tag
+or even better, to a SHA256 reference of the image.
```
trumant, line 1373: I fully intend to flesh this out with more detailed content once we reach consensus on the desirable behavior

src/audit/unpinned_container_images.rs
```diff
@@ -0,0 +1,93 @@
...
+    "unpinned-container-images",
+    "unpinned container image references"
+);
+
+impl Audit for UnpinnedContainerImages {
+    fn new(_state: &AuditState<'_>) -> Result<Self, AuditLoadError> {
+        Ok(Self)
+    }
+
+    fn audit_workflow<'doc>(&self, workflow: &'doc Workflow) -> Result<Vec<Finding<'doc>>> {
```
woodruffw, line 51: FYI: you can use `audit_normal_job` instead of `audit_workflow` to have `zizmor`'s machinery do the job loop + filter for you 🙂 -- see `bot_conditions.rs` for an example of that.

(That'll make the implementation here a bit shorter.)

src/audit/unpinned_container_images.rs
```diff
@@ -0,0 +1,93 @@
...
+        let mut findings = vec![];
+
+        for job in workflow.jobs() {
+            let Job::NormalJob(job) = &job else {
+                continue;
+            };
+
+            if let Some(Container::Container {
+                image,
```
woodruffw, line 60: Nit: here and below I'd suggest using [`DockerUses`](https://docs.rs/github-actions-models/latest/github_actions_models/common/struct.DockerUses.html) to parse the image into a model with a nice API (specifically, it has `image.tag()` and follows the OCI spec).

trumant: Thanks for the direct pointer. That was my initial inclination, but I should have read more of the existing codebase before rushing my contrib.

src/audit/unpinned_container_images.rs
```diff
@@ -0,0 +1,93 @@
...
+                        findings.push(
+                            self.build_finding(image, job.location().primary().with_keys(&["services".into(), service.as_str().into(), "image".into()]), workflow)?);
+                    }
+                    if self.container_image_ref_is_pinned_to_latest(image) {
+                        findings.push(
+                            self.build_finding(image, job.location().primary().with_keys(&["services".into(), service.as_str().into(), "image".into()]), workflow)?);
+                    }
+                }
+            }
+        }
```
woodruffw, line 89: Nit: you could probably do a collect-and-process here to make this a bit shorter, i.e. something like:

```rust
let mut images_with_locations = vec![];

if let Some(Container::Container { image, .. }) = &job.container {
	images.push((DockerUses::from_str(&image)?, job.location().primary().with_keys(...)));
}

for (service, config) in job.services.iter() {
  // push into images as well
}

for (image, loc) in images_with_locations {
  match image.tag() {
    None => todo!("untagged finding"),
    Some("latest") => todo!("latest tag finding"),
    Some(_) => continue,
  } 
}
```

(That's pseudocode, hopefully it gets the idea across.)

woodruffw:
Thanks @trumant! The overall approach here looks great to me and makes a lot of sense 🙂.



> * Should this be user configurable such that by default it issues finding when it sees `image: foo/bar` or `image: foo/bar:latest`, but support configuration that enforces SHA256 image refs only.

I think yes, but maybe not for this MVP -- long term it should probably be configurable in a manner similar to `unpinned-uses` (where there's a default set of policies but the user can override them), but for the MVP I think we can start with "flag anything untagged or with the `latest` tag.

> * A more sophisticated configuration could be to suppress findings for no tag and latest tag for certain specified docker registry hosts

Same thought -- I think this would be great for a follow on!

(I think ideally this would use the same configuration mechanism as above -- users could set both patterns and policies, e.g. `myregistry.com: any` for untagged/unhashed and `spooky.com: hash-pin` for hash-pin enforced.)

woodruffw:
> I'd also be interested in feedback on the testing coverage here. I'm not entirely sure that I understand the best way to ensure this is fully and usefully integrated into the acceptance test suite.

Take a look at <https://woodruffw.github.io/zizmor/development/#writing-snapshot-tests> (if you haven't yet) -- the best and easiest way to do this is to add tests to `snapshot.rs` along with test files and then use `cargo insta` to review/commit the snapshots.

Doing it that way is a lot faster than manually writing acceptance tests; the existing snapshot test layout (and test asset layout) should hopefully be a pointer to how it's done, but please let me know if there's anything confusing or insufficient about the docs as well!

src/audit/unpinned_container_images.rs
woodruffw: No problem at all! It's definitely not the best documented at the moment 🙂

src/audit/unpinned_container_images.rs
trumant: Yep, I get it. Thanks for taking the time to teach!

trumant: I hope I've captured the spirit of your feedback in my latest changes.

woodruffw (approved):
Thanks @trumant, this looks great to me!

I'm going to do a small amount of cleanup and rename this to `unpinned-images` (I'm trying to make the audit IDs shorter in general, to minimize reflows). Thanks a ton for building this!

PR #734, opened 2025-05-01
refactor: begin splitting out syntax/sema error handling

### LINKED ISSUE #725: [BUG] `zizmor` aborts on empty workflow files

woodruffw:
~~WIP.~~

~~Starting with the workflow path; I need to handle the action path as well.~~

TODO:

- [x] Actually plumb the variants through and skip + warn on syntax errors instead of failing
- [x] Rethink the nested `anyhow::Error`s here (these are causing pain when nested/chained)
- [x] More tests

Closes #732.

Fixes #725.


woodruffw:
The machinery of this is done, I just need to think more about how I want to do the skipping such that the current mess of collection code become DRYer.

PR #740, opened 2025-05-02
feat: emit pedantic finding for tagged OCI images

### LINKED ISSUE #736: Feature: `unpinned-uses` should emit pedantic findings for tagged-but-not-hash-pinned
Follow-on to #733: right now we emit a finding if an image is untagged or has the `:latest` tag, but we could/should probably do something similar to `unpinned-uses`:

1. If the image is hash-pinned, no finding at all
2. If the image is tag-pinned without a hash, emit a pedantic finding
3. If the image is tag-pinned to `:latest`, emit a normal finding (we do this already)
4. If the image is unpinned, emit a normal finding (we do this already)

trumant:
Pedantic findings are emitted by the unpinned-images audit when it encounters OCI image references that use a tag other than `latest`.

The pedantic finding is resolved, will not be produced when the image reference uses a SHA256 hash rather than a tag.

This closes #736

### DIFF
--- a/tests/integration/snapshots/integration__snapshot__unpinned_images.snap
+++ b/tests/integration/snapshots/integration__snapshot__unpinned_images.snap
@@ -1,6 +1,6 @@
 ---
 source: tests/integration/snapshot.rs
-expression: "zizmor().input(input_under_test(\"unpinned-images.yml\")).run()?"
+expression: "zizmor().input(input_under_test(\"unpinned-images.yml\")).args([\"--persona=pedantic\"]).run()?"
 ---
 error[unpinned-images]: unpinned image references
   --> @@INPUT@@:16:7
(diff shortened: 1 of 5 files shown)


woodruffw:
Awesome, thanks @trumant!

tests/integration/snapshots/integration__snapshot__unpinned_images.snap
```diff
@@ -1,6 +1,6 @@
 ---
 source: tests/integration/snapshot.rs
-expression: "zizmor().input(input_under_test(\"unpinned-images.yml\")).run()?"
+expression: "zizmor().input(input_under_test(\"unpinned-images.yml\")).args([\"--persona=pedantic\"]).run()?"
```
trumant, line 3: 🤦 - thank you!

PR #743, opened 2025-05-04
feat: extend template injection audit & use CodeQL models

### LINKED ISSUE #417: template-injection: investigate other sinks
Right now we support `run:` and `actions/github-script` as potential "sinks" where template injection can occur, but there are others:

- [x] `azure/powershell`: runs powershell scripts
- [x] `azure/cli`: runs bash scripts

xref https://github.blog/security/application-security/how-to-secure-your-github-actions-workflows-with-codeql/

Marcono1234:
Resolves #417

This is a proof-of-concept for refactoring the `template-injection` audit to make it easier to add new injection sinks in the future, and to use CodeQL's models.

A new GitHub workflow is added which runs on schedule and on demand which pull the https://github.com/github/codeql repository, processes their injection sink models and generates a file in custom zizmor-specific format, and creates a pull request for updating the file.
That file is loaded by zizmor's `template-injection` audit at compile time.

Here is an example run from my fork:
- https://github.com/Marcono1234/zizmor/actions/runs/14815681649/job/41595747538
- Generated pull request: https://github.com/Marcono1234/zizmor/pull/2

The idea for that workflow is based on https://github.com/gradle/actions/blob/main/.github/workflows/update-checksums-file.yml (which is a similar use case for processing external resources and including them in the build).

### DIFF
--- a/.github/workflows/process-codeql-sink-models.js
+++ b/.github/workflows/process-codeql-sink-models.js
@@ -64,7 +64,7 @@ function processYamlFile(filePath, relevantKinds, onlyManualModels) {
                 continue
             }
 
-            // Ignore 'on: workflow_call' inputs; seems to be mostly for for 'generated' models
+            // Ignore 'on: workflow_call' inputs; seems to be mostly for 'generated' models
             if (action.includes('/.github/workflows/')) {
                 continue
             }
--- a/.github/workflows/process-codeql-sink-models.js
+++ b/.github/workflows/process-codeql-sink-models.js
@@ -1,0 +1,135 @@
+/*
+ * Processes the CodeQL models from https://github.com/github/codeql/blob/codeql-cli/v2.21.2/actions/ql/lib/ext
+ * and extracts the information needed by zizmor
+ */
+
+// @ts-check
+
+const path = require('node:path')
+const fs = require('node:fs')
+// yaml library is installed by GitHub workflow
+const yaml = require('../../codeql-models-working-dir/node_modules/yaml')
+
+/** @type Map<string, string[]> */
+const codeInjectionSinks = new Map()
+
+/**
+ * @param {fs.PathLike} filePath
+ * @param {Set<string>} relevantKinds which sink kinds are relevant
+ * @param {boolean} onlyManualModels whether to include only models with 'provenance == manual'
+ */
+function processYamlFile(filePath, relevantKinds, onlyManualModels) {
+    const content = yaml.parse(fs.readFileSync(filePath, { encoding: 'utf8' }))
+    const extensions = content['extensions']
+
+    if (extensions === undefined) {
+        throw new Error('Missing extensions: ' + content)
+    }
+
+    for (const extension of extensions) {
+        const addsTo = extension['addsTo']
+        if (addsTo === undefined) {
+            throw new Error('Missing addsTo: ' + content)
+        }
+
+        const extensible = addsTo['extensible']
+        if (extensible !== 'actionsSinkModel') {
+            continue
+        }
+
+        const pack = addsTo['pack']
+        // Fail if CodeQL starts using other packs, have to examine then what this means,
+        // e.g. whether it has lower accuracy or severity
+        if (pack !== 'codeql/actions-all') {
+            throw new Error('Unexpected pack: ' + pack)
+        }
+
+        const data = extension['data']
+        if (data === undefined) {
+            throw new Error('Missing data: ' + content)
+        }
+
+        for (const dataEntry of data) {
+            if (dataEntry.length !== 5) {
+                throw new Error('Contains malformed data entry: ' + dataEntry)
+            }
+
+            // See https://github.com/github/codeql/blob/codeql-cli/v2.21.2/actions/ql/lib/codeql/actions/dataflow/internal/ExternalFlowExtensions.qll#L22-L24
+            /** @type string[] */
+            const [action, version, input, kind, provenance] = dataEntry
+            if (!relevantKinds.has(kind)) {
+                continue
+            }
+            if (onlyManualModels && provenance !== 'manual') {
+                continue
+            }
+
+            // Ignore 'on: workflow_call' inputs seem to be mostly for for 'generated' models
+            if (action.includes('/.github/workflows/')) {
+                continue
+            }
+
+            // Currently all models use only '*' as affected version, so for simplicity only
+            // support that for now
+            if (version !== '*') {
+                throw new Error(
+                    'Non-wildcard versions are not supported yet: ' + version
+                )
+            }
+
+            const inputPrefix = 'input.'
+            if (!input.startsWith(inputPrefix)) {
+                throw new Error('Contains input with unexpected format: ' + input)
+            }
+            const inputName = input.substring(inputPrefix.length)
+
+            let inputs = codeInjectionSinks.get(action)
+            if (inputs === undefined) {
+                inputs = []
+                codeInjectionSinks.set(action, inputs)
+            }
+            inputs.push(inputName)
+        }
+    }
+}
+
+/**
+ * @param {string} codeQlDir
+ * @param {fs.PathLike} outputFile
+ */
+function processModels(codeQlDir, outputFile) {
+    const modelsDir = path.join(codeQlDir, 'actions/ql/lib/ext')
+    const files = fs.readdirSync(modelsDir, {
+        recursive: true,
+        withFileTypes: true,
+    })
+
+    const relevantKinds = new Set(['code-injection'])
+    // For now only include models manually curated by the CodeQL developers
+    const onlyManualModels = true
+
+    let processedCount = 0
+    for (const file of files) {
+        if (file.isFile()) {
+            const name = file.name
+            if (name.endsWith('.yml') || name.endsWith('.yaml')) {
+                processedCount++
+                const filePath = path.join(file.parentPath, name)
+                try {
+                    processYamlFile(filePath, relevantKinds, onlyManualModels)
+                } catch (e) {
+                    throw new Error('Failed processing file: ' + filePath, { cause: e })
+                }
+            }
+        }
+    }
+
+    console.info(`Processed ${processedCount} files`)
+
+    // Important: Data format must match the parsing logic in the Rust code
+    const zizmorModelData = [...codeInjectionSinks].sort().map(entry => `${entry[0]}|${entry[1].join(',')}\n`).join('')
+    fs.writeFileSync(outputFile, zizmorModelData)
+    console.info(`Wrote model data to file: ${outputFile}`)
+}
+
+processModels('codeql', '../code-injection-models.txt')
--- a/.github/workflows/process-codeql-sink-models.js
+++ b/.github/workflows/process-codeql-sink-models.js
@@ -1,5 +1,5 @@
 /*
- * Processes the CodeQL models from https://github.com/github/codeql/blob/codeql-cli/v2.21.2/actions/ql/lib/ext
+ * Processes the CodeQL models from https://github.com/github/codeql/tree/main/actions/ql/lib/ext
  * and extracts the information needed by zizmor
  */
 
@@ -64,7 +64,7 @@ function processYamlFile(filePath, relevantKinds, onlyManualModels) {
                 continue
             }
 
-            // Ignore 'on: workflow_call' inputs seem to be mostly for for 'generated' models
+            // Ignore 'on: workflow_call' inputs; seems to be mostly for for 'generated' models
             if (action.includes('/.github/workflows/')) {
                 continue
             }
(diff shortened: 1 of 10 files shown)


.github/workflows/update-code-injection-models.yml
```diff
@@ -0,0 +1,72 @@
...
+    - cron: '31 7 * * 5'
+  workflow_dispatch:
+
+permissions: {}
+
+jobs:
+  update-models:
+    permissions:
+      contents: write
+      pull-requests: write
```
Marcono1234, line 15: If you prefer, this could be split into two jobs: One with only `read` privileges processing the CodeQL data, and one with `write` privileges which creates the pull request.

.github/workflows/process-codeql-sink-models.js
```diff
@@ -0,0 +1,135 @@
...
+                    throw new Error('Failed processing file: ' + filePath, { cause: e })
+                }
+            }
+        }
+    }
+
+    console.info(`Processed ${processedCount} files`)
+
+    // Important: Data format must match the parsing logic in the Rust code
+    const zizmorModelData = [...codeInjectionSinks].sort().map(entry => `${entry[0]}|${entry[1].join(',')}\n`).join('')
```
Marcono1234, line 130: Used `|` and `,` here are separators since they are probably not valid as part of repository or `inputs` name.

code-injection-models.txt
Marcono1234, line 1: Probably don't want this in the repository root. Any ideas for a better location?

src/audit/template_injection.rs
```diff
@@ -87,40 +89,57 @@ const SAFE_CONTEXTS: &[&str] = &[
...
+                Some(inputs) => inputs,
+                None => &[],
+            },
+        }
+    }
+
+    fn scripts_with_location<'s>(
         step: &impl StepCommon<'s>,
-    ) -> Option<(String, SymbolicLocation<'s>)> {
+    ) -> Vec<(String, SymbolicLocation<'s>)> {
```
Marcono1234, line 122: Changed to `Vec` because this can in theory now return multiple vulnerable `inputs`.

src/models/uses.rs
```diff
@@ -143,6 +147,18 @@ pub(crate) trait RepositoryUsesExt {
...
+    fn path_without_ref(&self) -> String {
+        format!(
+            "{}/{}{}",
+            self.owner,
+            self.repo,
+            self.subpath
+                .as_ref()
+                .map_or("".to_owned(), |subpath| format!("/{subpath}"))
+        )
+        .to_lowercase()
```
Marcono1234, line 159: Performed lowercasing here to be consistent with the other methods which seem to treat this case insensitively as well.

Marcono1234 (commented):
This is really just an initial draft; any feedback is appreciated!

.github/workflows/update-code-injection-models.yml
```diff
@@ -0,0 +1,73 @@
...
+      - name: Update model file
+        run: node ../.github/workflows/process-codeql-sink-models.js
+        working-directory: codeql-models-working-dir
+
+      # If there are no changes, this action will not create a pull request
+      - name: Create or update pull request
+        uses: peter-evans/create-pull-request@271a8d0340265f705b14b6d32b9829c1cb33d45e # v7
+        with:
+          add-paths: |
+            code-injection-models.txt
```
Marcono1234, line 56: Have to explicitly specify this because the `codeql-models-working-dir` is also within the repository working tree.

Maybe that is not ideal because https://github.com/peter-evans/create-pull-request/tree/v7.0.8?tab=readme-ov-file#add-specific-paths mentions that other files (i.e. `codeql-models-working-dir`?) will be stashed and restored.

Might be better to have `codeql-models-working-dir` outside the working tree, maybe can created it as `../codeql-models-working-dir`?

.github/workflows/update-code-injection-models.yml
```diff
@@ -0,0 +1,73 @@
...
+          mkdir codeql-models-working-dir
+
+      - name: Install dependencies
+        run: |
+          npm install yaml@2.7.1 --no-save
+        working-directory: codeql-models-working-dir
+
+      - name: Check out CodeQL
+        id: checkout_codeql
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4
```
Marcono1234, line 40: Currently checks out the latest commit of the default branch. Might be nicer to check out the latest release tag instead for better stability. But maybe not worth it, and might not matter that much or at all assuming that (at least the `manual`) model files probably don't contain bugs which are only present on the latest default branch commits but not releases.

woodruffw:
Nice, thanks @Marcono1234! Taking a look now.

code-injection-models.txt
woodruffw, line 1: Instead of the repo root, maybe this could go in `src/data`? That's where we stash some other machine-readable files that get included into the binary at build time.

.github/workflows/process-codeql-sink-models.js
woodruffw: Instead of creating our own syntax here, perhaps we could use JSON? That'll be slightly slower at initialization time in `zizmor` itself, but IMO it's worth it versus doing bespoke ser/de.

.github/workflows/update-code-injection-models.yml
woodruffw: I'm OK either way -- pulling from the default branch seems fine, since I can always reject PRs that look wrong.

(Getting the latest tag when there's no matching release is very annoying with GitHub's REST API -- you need to list the tags, then sort them by creation date, which requires addition roundtrips for each tag object. Not worth it, IMO :slightly_smiling_face:)

.github/workflows/update-code-injection-models.yml
```diff
@@ -0,0 +1,73 @@
...
+            code-injection-models.txt
+          branch: bot/code-injection-models-update
+          commit-message: |
+            Update code injection models
+
+            https://github.com/github/codeql/commit/${{ steps.checkout_codeql.outputs.commit }}
+          title: Update code injection models
+          # Note: Unfortunately this action cannot trigger the regular workflows for the PR automatically, see
+          # https://github.com/peter-evans/create-pull-request/blob/main/docs/concepts-guidelines.md#triggering-further-workflow-runs
+          # Therefore suggest below to close and then reopen the PR
```
woodruffw, line 65: You can remove this comment since it's mentioned in the PR body as well :slightly_smiling_face: 

```suggestion
```

(I love that this hack still works.)

.github/workflows/update-code-injection-models.yml
woodruffw: Keeping these in the same job seems fine to me for now :slightly_smiling_face:

.github/workflows/process-codeql-sink-models.js
woodruffw, line 1: Two high-level thoughts here:

1. As-is, this code generation step only runs in CI. That's a bit different from other generation steps currently in `zizmor`, which are generally orchestrated through the `Makefile` so that they can run locally as well (e.g. `make snippets`, `make refresh-schemas`, etc.). Ideally we'd do he same here, e.g. with `make template-injection-sinks` (or a better name).
2. I have a pretty strong preference for doing this kind of generation in Python, since most of `zizmor`'s other generation support code is Python (and I'm borderline unfamiliar with JS). Another separate benefit of doing this in Python is that you can add a `codegen` (or a better name) dependency group in `pyproject.toml` and our `uv` setup will handle deps gracefully for us, e.g. via `uv run --only-group codegen support/process-codeql-sink-models.py`.

TL;DR: Ideally this would live somewhere like `support/process-codeql-sink-models.py` and would be invokable both locally and in CI via `make template-injection-sinks`, since that would be similar to what we do in [`refresh-schemas.yml`](https://github.com/woodruffw/zizmor/blob/main/.github/workflows/refresh-schemas.yml).

Let me know what you think about the above -- I can help out with the Python bits as needed.

docs/audits.md
```diff
@@ -979,6 +979,9 @@ used with attacker-controllable expression contexts, such as
 `github.event.issue.title` (which the attacker can fully control by supplying
 a new issue title).
 
+The list of action inputs which are vulnerable to injection attacks is
+based on [GitHub's CodeQL models](https://github.com/github/codeql/blob/main/actions/ql/lib/ext).
```
woodruffw, line 983:
```suggestion
derived from [GitHub's CodeQL models](https://github.com/github/codeql/blob/main/actions/ql/lib/ext).
```

src/audit/template_injection.rs
```diff
@@ -3,17 +3,19 @@
 //! This looks for job steps where the step contains indicators of template
 //! expansion, i.e. anything matching `${{ }}`.
 //!
-//! The following steps are currently supported:
-//! * `run:`, indicating template expansion into a shell script or similar
-//! * `actions/github-script`, indicating template expansion into a JavaScript function
+//! The list of injection sinks is based on [CodeQL's models](https://github.com/github/codeql/blob/main/actions/ql/lib/ext).
```
woodruffw, line 6:
```suggestion
//! The list of injection sinks is derived from [CodeQL's models](https://github.com/github/codeql/blob/main/actions/ql/lib/ext).
```

src/audit/template_injection.rs
```diff
@@ -3,17 +3,19 @@
...
+//! The models are processed and filtered by a workflow on the zizmor repository.
```
woodruffw, line 7: Note: we should also include some kind of attribution statement here (my read of the MIT License is that one *technically* isn't needed because this is a wholesale transformation of data rather than a "copy or substantial portion," but IANAL).

Something like this should be fine:

```
Sinks derived from the CodeQL models are licensed by GitHub, Inc.
under the terms of the MIT License.
```

src/audit/template_injection.rs
```diff
@@ -87,40 +89,57 @@ const SAFE_CONTEXTS: &[&str] = &[
...
-    ) -> Option<(String, SymbolicLocation<'s>)> {
+    ) -> Vec<(String, SymbolicLocation<'s>)> {
         match step.body() {
             models::StepBodyCommon::Uses {
                 uses: Uses::Repository(uses),
                 with,
-            } => {
-                if uses.matches("actions/github-script") {
-                    with.get("script").map(|script| {
+            } => TemplateInjection::action_injection_inputs(&uses.path_without_ref())
```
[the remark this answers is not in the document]
woodruffw: (For using `RepositoryUsesPattern`, the easiest thing to do would probably be to have `MODEL_INJECTION_SINKS` be `HashMap<RepositoryUsesPattern, ...>` instead of `HashMap<&str, ...>`)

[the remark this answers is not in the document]
Marcono1234: > HashMap<RepositoryUsesPattern, ...>

Does that really work as desired? Won't the derived `PartialEq` and `Hash` then simply compare equal `RepositoryUsesPattern` but not actually use the `matches` method (that is for example `::ExactPath` and `::InRepo` will never be equal)?

So maybe `RepositoryUsesPattern` shouldn't really derive `PartialEq` and `Hash`?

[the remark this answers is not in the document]
Marcono1234: In that case I guess it could be a `Vec(RepositoryUsesPattern, ...)` (or similar) and the iterating over the entries and calling their `matches` methods.

Not sure if that would be a performance concern.

.github/workflows/process-codeql-sink-models.js
Marcono1234: > which are generally orchestrated through the `Makefile` so that they can run locally as well

- Should this new script then even be run as GitHub workflow or only manually?
  - Or should the workflow then be similar to `.github/workflows/refresh-schemas.yml`?
- Should this script then include cloning the CodeQL repo (in a temporary directory)? I guess yes to make it standalone.

I can maybe give that Python approach a try, but a review afterwards would be great.


> Ideally this would live somewhere like `support/process-codeql-sink-models.py`

Or what about `code-gen/...`?

woodruffw: > * Or should the workflow then be similar to `.github/workflows/refresh-schemas.yml`?

Yep, exactly -- it should be its own small workflow similar to `refresh-schemas.yml`.

(Or it could even go in that workflow, which we could rename to `refresh-codegen.yml` or similar.)

> * Should this script then include cloning the CodeQL repo (in a temporary directory)? I guess yes to make it standalone.

Yep, that's what I was thinking -- it's a little bit more verbose on the script's side, but doing the clone internally makes everything a bit more hermetic.

> I can maybe give that Python approach a try, but a review afterwards would be great.

Thanks, I'd be happy to!

> Or what about `code-gen/...`?

`codegen/` works for me (no hyphen please :wink:)

src/audit/template_injection.rs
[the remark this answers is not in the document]
woodruffw: > In that case I guess it could be a `Vec(RepositoryUsesPattern, ...)` (or similar) and then iterating over the entries and calling their `matches` method.
> 
> Not sure if that would be a performance concern.

Yeah, this is what I was thinking -- it would be a performance concern at larger sizes, but with the current number of patterns (well under 100) I think a linear scan should be fine.

PR #745, opened 2025-05-04
Extract context patterns from webhook schemata

### LINKED ISSUE #674: Feature: Handle `template-injection` allowlisting schematically
### Pre-submission checks

The current `template-injection` audit has a manually curated allowlist of "safe contexts," which aren't exploitable even if expanded in a shell or similar setting:

https://github.com/woodruffw/zizmor/blob/07d2c2401d0485dad2181f20252f5452ab8ee105/src/audit/template_injection.rs#L36

This manual approach is non-ideal, since it (1) requires ongoing curation, and (2) is susceptible to false positives that require a release cycle (the current cycle has at least three of these alone).

Instead of manually curating `SAFE_CONTEXTS`, we should consume JSON schemata for GitHub's webhook payloads, which are the underlying driver for these contexts.

This has a few moving parts:

1. We need JSON schemata for all of GitHub's webhook payloads, or at least the ones that correspond to workflow triggers;
2. We need to ingest those in a scalable way (perhaps by pre-processing them into a linear set with safe/unsafe expansion categories).

### Additional context

It's unclear to me whether GitHub has public schemata available for its webhook payloads -- payloads are documented, but I couldn't immediately find any schemas online.

See also #673 for context.

woodruffw:
~~WIP. Still thinking through how I want to approach this.~~

Fixes #674.


woodruffw:
Noting for myself: doing a linear scan over thousands of fixed contexts for each dynamic context (i.e. parsed from expressions) is going to get slow, quick. This is probably going to require a prefix tree/trie structure so that I can do efficient prefix lookups.

woodruffw:
I've developed a bit of a yak stack for this:

- To feel confident with my trie/whatever approach here, I need to benchmark it, e.g. with `criterion`;
- But to bring in benchmarks, I need to start breaking out more of `zizmor` as library code;
- And to do that I need to workspace-ify things.

woodruffw:
Preliminary benchmarking shows a mixed effect on performance: running with `--no-online-audits` is about 9% slower on `python/cpython`, but about 6% faster on `grafana/grafana`.

(This is primarily a correctness change so even a small performance hit would be worth it, but it's good to see that it actually makes things faster in some cases.)

Issue #764, opened 2025-05-08
Feature: reduce GitHub REST API usage in favor of clones?


woodruffw:
GitHub imposes rate limits on their REST API; these limits can be onerous for higher-volume users, since some of our audits (particularly `impostor-commit` and `ref-confusion`) require a _lot_ of individual paginated requests to fetch the needed repository state.

To reduce this, we could switch to (or offer a model) for a model where we `git clone` and collect history locally instead, since GitHub has much higher rate limits on clones. 

Pros:

- Fewer rate limit issues.
- Potentially much, much faster overall (since local Git object/history scanning will be a lot faster than our current network roundtrips)

Cons:

- A bit more complicated (but not much more)
- Probably slightly slower on each audit cycle (i.e. per `uses:` clause), since we'd need to `git clone` and pull down more data initially. This will be cached and amortized by the faster filtering (above), but it'll probably make these audits a little less responsive.

Related thoughts:

- Maybe a blobless clone will be faster? I'm not sure if this induces a tradeoff on GitHub's side or not, versus the hot path for a normal clone.
- This is also somewhat related to #278, which seeks to solve the same problem by deploying our own static API. 

(h/t @andrew)

jamesc-grafana:
Hey 👋 

I've been thinking about this issue recently, as with the last issue, I'll preface this with Rust is not my native language at the moment, but I have been playing around with this.

I've taken the approach of changing the existing functionality rather than offering this as a flag. What I've tried to do is use the `git2-rs` library to talk directly to the `Remote` for listing refs and then filtering them into tags/branches for the appropriate functions.

The [working branch](https://github.com/jamesc-grafana/zizmor/tree/git-api) that I've got going doesn't seem to break any of the existing tests when run in online mode. I'm happy to continue working on this, but as before don't want to block progress

Thanks,

James

woodruffw:
Hey @jamesc-grafana, thanks a ton for looking into this!

> The [working branch](https://github.com/jamesc-grafana/zizmor/tree/git-api?rgh-link-date=2025-05-14T14%3A13%3A59.000Z) that I've got going doesn't seem to break any of the existing tests when run in online mode. I'm happy to continue working on this, but as before don't want to block progress

[...]

PR #801, opened 2025-05-16
feat: use git database to fetch references and content instead of github api

jamesc-grafana:
Works towards: #764 using `git2-rs` to talk directly to the Git Remote and bare cloning for the comparison operations


woodruffw:
Thanks @jamesc-grafana! I'll be able to give this a review later today!

crates/zizmor/src/github_api.rs
```diff
@@ -107,99 +119,185 @@ impl Client {
         Self {
             api_base: hostname.to_api_url(),
             http,
+            host: hostname.clone(),
+            token: match token.len() {
+                0 => None,
+                _ => Some(token.to_string()),
+            },
+            fetched_repos: RwLock::new(HashMap::new()),
```
woodruffw, line 127: Thinking out loud: instead using a `bool` as a sentinel here, can we make this a `HashMap<String, PathBuf>` or similar to track the underlying cached directory? I _think_ that would reduce the number of checks needed below.

crates/zizmor/src/github_api.rs
```diff
@@ -107,99 +119,185 @@ impl Client {
...
+                }
+            })
+            .collect())
+    }
 
-            dest.extend(page);
-            pageno += 1;
+    #[instrument(skip(self))]
+    pub(crate) async fn get_repo(&self, owner: &str, repo: &str) -> Result<Repository> {
+        let tmp = std::env::temp_dir().join(format!("zizmor/{owner}-{repo}"));
```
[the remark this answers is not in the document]
jamesc-grafana: Good point, fixed in 0fb174e

crates/zizmor/src/github_api.rs
jamesc-grafana: I've updated the `get_repo` and `run_fetch` to use the PathBuf in the HashMap if it exists to open the repo, we already know about it. Otherwise to check existence and then fetch or clone based on the status of the repo

Changes in 0fb174e

woodruffw: Thanks!

crates/zizmor/src/github_api.rs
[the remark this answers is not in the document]
jamesc-grafana: Ok, had to rebase for `Cargo.toml` so fixed in [ff539fd](https://github.com/zizmorcore/zizmor/pull/801/commits/ff539fddca05c9fb907eb4879fdaaf3a9c1b7bdf) instead

woodruffw:
Hey @jamesc-grafana, sorry for the delay here! This popped a bit further down the stack after the fix APIs 😉 

I'm taking another look at this now, and I think the overall approach looks great! One outstanding concern I have is with performance, however -- on local tests this is a _lot_ slower than the current REST API calls, probably because `git2` ends up pulling a lot more data over the wire. Do you have any thoughts on how we can optimize this?

(Normally I'd reach for shallow clones + sparse checkouts here, but I'm not super familiar with libgit2 or `git2` for Rust 🙂)

jamesc-grafana:
> One outstanding concern I have is with performance, however -- on local tests this is a _lot_ slower than the current REST API calls

Yes, I also find this on an initial run too, however, once the initial clone is complete I was finding it quicker than the API calls. It's currently doing a "bare", so this could be the problem, I think that your suggestion of shallow/sparse clones might help only then fetching more data when required.

> (Normally I'd reach for shallow clones + sparse checkouts here, but I'm not super familiar with libgit2 or `git2` for Rust 🙂)

I will look into `git2-rs` for shallow and sparse checkouts

woodruffw:
Hey @jamesc-grafana! There's some good news on the performance/API quota side of things: #1257 switches a lot of the ref lookups to the equivalent of `git ls-remote`, which avoids the problem of having to fetch any repository state upfront. From my benchmarking, this makes typical online audit workloads between 40% and 90% faster (depending on how many times refs get reused, how hot the cache is, etc.)

jamesc-grafana:
Hey @woodruffw 👋 that's great news, thanks for letting me know. I'll try this out internally to checkout the performance gains. This is very exciting indeed 😄

woodruffw:
I think the bulk of the perf gains here have been accomplished through #1257 and #1273, so I'm closing this for now!

(In the future, it might still make sense to have multiple "backends" for Git operations, but for now I think the bulk of the GitHub API usage has been eliminated through the `ls-refs` trick.)

PR #802, opened 2025-05-16
refactor: add test-support crate, prep for benching

woodruffw:


woodruffw:
This approach isn't going to work.

PR #810, opened 2025-05-18
chore(doc): remove event filters

woodruffw:
Removing these entirely means that we trigger on all `push` and `pull_request` events, which is typically what we want.


woodruffw:
Changed my mind; this causes too many duplicate runs.

PR #838, opened 2025-05-21
feat(cli): add completion generation support for nushell

jcgruenhage:
I'm packaging zizmor for Chimera Linux right now (see https://github.com/chimera-linux/cports/pull/4255), and Chimera Linux is shipping nushell completions where available. As a nushell user, I'd love to see nushell completions for zizmor in addition to the ones already provided. While I was at it, I also added carapace, which is a multi-shell completion engine.

The way I'm doing it here is not great, because it requires replicating `clap_complete::Shell` and extending it to support the additional shells. An alternative that requires less code on the zizmor side is to use something like `clap_allgen`, but that is not intended to be used with a `Shell` enum, it's intended use is to generate all completions with one function call. Going for that would be a breaking change of the CLI interface, which I'd assume would need to wait for a v2, if it's wanted at all?

A completely different alternative I'd like to bring up is generating completions during build time using a `build.rs` file. This would allow cross compiling zizmor and still shipping completion packages without resorting to emulation or compiling twice. That requires bigger changes though, so didn't want to spend time on this without discussing it first.

### DIFF
--- a/crates/zizmor/src/main.rs
+++ b/crates/zizmor/src/main.rs
@@ -156,7 +156,7 @@ enum Shell {
     /// Nushell
     Nushell,
     /// `PowerShell`
-    PowerShell,
+    Powershell,
     /// Z `SHell` (zsh)
     Zsh,
 }
--- a/crates/zizmor/src/main.rs
+++ b/crates/zizmor/src/main.rs
@@ -10,6 +10,7 @@ use anyhow::{Context, Result, anyhow};
 use audit::{Audit, AuditLoadError};
 use camino::{Utf8Path, Utf8PathBuf};
 use clap::{CommandFactory, Parser, ValueEnum};
+use clap_complete::Generator;
 use clap_verbosity_flag::InfoLevel;
 use config::Config;
 use finding::{Confidence, Persona, Severity};
@@ -126,7 +127,7 @@ struct App {
 
     /// Generate tab completion scripts for the specified shell.
     #[arg(long, value_enum, value_name = "SHELL", exclusive = true)]
-    completions: Option<clap_complete::Shell>,
+    completions: Option<Shell>,
 
     /// Enable naches mode.
     #[arg(long, hide = true, env = "ZIZMOR_NACHES")]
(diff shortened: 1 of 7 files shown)


woodruffw:
Hey @jcgruenhage, thanks for opening a PR!

> As a nushell user, I'd love to see nushell completions for zizmor in addition to the ones already provided.

Cool, thank you!

> While I was at it, I also added carapace, which is a multi-shell completion engine.

All things being equal, I'd prefer to defer this until there's a end-user request for it -- that keeps things simpler for me in terms of dependency review. Does that work for you?

For the rest: I'll do a review now 🙂

crates/zizmor/src/main.rs
```diff
@@ -142,6 +144,101 @@ struct App {
...
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        for variant in Self::value_variants() {
+            if variant.to_possible_value().unwrap().matches(s, false) {
+                return Ok(*variant);
+            }
+        }
+        Err(format!("invalid variant: {s}"))
+    }
+}
```
woodruffw, line 187: Apologies if I'm missing it, but where do we this `FromStr` impl?

crates/zizmor/src/main.rs
```diff
@@ -142,6 +144,101 @@ struct App {
     inputs: Vec<String>,
 }
 
+/// Shell with auto-generated completion script available.
+#[derive(Clone, Copy, Debug, Eq, Hash, PartialEq)]
+#[non_exhaustive]
+pub enum Shell {
```
woodruffw, line 150: This is not a public API so IMO we can make it exhaustive + drop the `pub` qualifier 🙂 

```suggestion
enum Shell {
```

jcgruenhage: done

crates/zizmor/src/main.rs
jcgruenhage: We don't, it was part of the block I cargo-culted over from https://github.com/clap-rs/clap/blob/master/clap_complete/src/aot/shells/shell.rs. It was between two parts that were needed, and I forgot to remove it.

crates/zizmor/src/main.rs
```diff
@@ -142,6 +144,101 @@ struct App {
...
+            if variant.to_possible_value().unwrap().matches(s, false) {
+                return Ok(*variant);
+            }
+        }
+        Err(format!("invalid variant: {s}"))
+    }
+}
+
+// Hand-rolled so it can work even when `derive` feature is disabled
+impl ValueEnum for Shell {
```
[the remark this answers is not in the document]
jcgruenhage: Nope, good catch. Have derived this bit now.

crates/zizmor/src/models.rs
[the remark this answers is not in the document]
woodruffw: Hmm, I bet that's because of the recent Rust release and GitHub's runner images lagging behind; I've seen that happen before. No worries here!

crates/zizmor/src/main.rs
```diff
@@ -142,6 +143,48 @@ struct App {
...
+    /// Bourne Again `SHell` (bash)
+    Bash,
+    /// Elvish shell
+    Elvish,
+    /// Friendly Interactive `SHell` (fish)
+    Fish,
+    /// Nushell
+    Nushell,
+    /// `PowerShell`
+    PowerShell,
```
woodruffw, line 159: I think if you drop the case here this will become `powershell` instead of `power-shell` in the `ValueEnum` 🙂 

```suggestion
    Powershell,
```

(That will also allow us to remove the `enum_variant_names` annotation.)

woodruffw:
BTW, don't worry about force-pushing -- I use a squash-merge workflow so having the commit history here is fine.

woodruffw (approved):

woodruffw:
Thanks a ton @jcgruenhage!

PR #840, opened 2025-05-21
fix: handle booleans for insecure-commands correctly

### LINKED ISSUE #839: [BUG]: insecure_commands audit doesn't account for boolean as value
### Pre-submission checks

### zizmor version

1.6.0

### Expected behavior

I expected zizmor to trigger a finding.

### Actual behavior

No findings.

### Reproduction steps

Run the test workflow from [here](https://github.com/zizmorcore/zizmor/blob/main/crates/zizmor/tests/integration/test-data/insecure-commands.yml), but instead of "ACTIONS_ALLOW_UNSECURE_COMMANDS: yes", put "ACTIONS_ALLOW_UNSECURE_COMMANDS: true".

### Logs

```plain text

```

### Additional context

I've looked this up online and it seems like GitHub actually prompted to put true, eg in https://github.com/actions/toolkit/issues/641 :

>  Please upgrade to using Environment Files or opt into unsecure command execution by setting the `ACTIONS_ALLOW_UNSECURE_COMMANDS` environment variable to `true`.

My thesis project is basically "scan a bunch of workflows with a few different scanners and report on results", so it might not be my last ticket. But I understand that it might cause frustration because "ah this person just put a few mindless tickets", so if there's anything else I need to triage the issue/etc, let me know.

woodruffw:
TL;DR: We previously (incorrectly) treated any nonempty string as a "true" value for `insecure-commands`, when in reality the GitHub Actions runner uses C#'s `Boolean.TryParse` for this.

We fix our handling by reproducing `Boolean.TryParse`'s behavior.

Fixes #839.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -770,7 +770,7 @@ dependencies = [
 
 [[package]]
 name = "github-actions-models"
-version = "0.28.3"
+version = "0.29.0"
 dependencies = [
  "indexmap",
  "serde",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,7 +17,7 @@ license = "MIT"
 [workspace.dependencies]
 anyhow = "1.0.98"
 github-actions-expressions = { path = "crates/github-actions-expressions", version = "0.0.3" }
-github-actions-models = { path = "crates/github-actions-models", version = "0.28.3" }
+github-actions-models = { path = "crates/github-actions-models", version = "0.29.0" }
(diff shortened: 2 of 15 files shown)


woodruffw:
For validation, I also checked GitHub's code search:

https://github.com/search?q=ACTIONS_ALLOW_UNSECURE_COMMANDS+language%3AYAML&type=code&l=YAML

PR #843, opened 2025-05-22
docs: describe installation of Chimera Linux package

jcgruenhage:
With https://github.com/chimera-linux/cports/pull/4255 merged, zizmor is now being built for Chimera Linux and Chimera Linux users will be able to install it from there, so I thought I'd update the docs.

How do the icons work? I've looked around and couldn't really find where they are sourced from.

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -112,6 +112,18 @@ description: Installation instructions for zizmor.
     pacman -S zizmor
     ```
 
+=== "Chimera Linux"
+
+    [![Chimera Linux package](https://repology.org/badge/version-for-repo/chimera/zizmor.svg)](https://repology.org/project/zizmor/versions)
+
+    !!! note
+
+        This is a community-maintained package.
+
+    ```bash
+    apk add zizmor
+    ```
+
 === "Other ecosystems"
 
     !!! info


funnelfiasco:
> How do the icons work? I've looked around and couldn't really find where they are sourced from.

It took some digging for me to figure it out, but they come from [Material for MkDocs](https://squidfunk.github.io/mkdocs-material/reference/icons-emojis/), which bundles several icon sources. Unfortunately, it doesn't look like any of them have an icon for Chimera. 

If you have a suitable SVG file of the logo, you may be able to [add it](https://squidfunk.github.io/mkdocs-material/setup/changing-the-logo-and-icons/#additional-icons), but I'm not sure if @woodruffw would want that or not.

woodruffw:
> If you have a suitable SVG file of the logo, you may be able to [add it](https://squidfunk.github.io/mkdocs-material/setup/changing-the-logo-and-icons/#additional-icons), but I'm not sure if @woodruffw would want that or not.

I'm happy with a custom logo/icon SVG, if the Chimera project has one 🙂 

(Please feel free to send a follow-up PR for that!)

woodruffw (approved):
LGTM!

(Thinking out loud: I know a few different `apk`-based distros are planning on packaging `zizmor`, so at some point it might make sense to fold all of them into a single tab since the instructions are the same. But that's a future topic/problem.)

triallax:
the chimera zizmor package is in the `user` repository, which is not enabled by default, you have to install `chimera-repo-user` first if not done already (see https://chimera-linux.org/docs/apk#repositories)

perhaps the section added here should be amended to clarify that

woodruffw:
> the chimera zizmor package is in the `user` repository, which is not enabled by default, you have to install `chimera-repo-user` first if not done already (see [chimera-linux.org/docs/apk#repositories](https://chimera-linux.org/docs/apk#repositories))
> 
> perhaps the section added here should be amended to clarify that

Could you or @jcgruenhage send me a PR for that? I'm not super familiar with apk or Chimera, so it'd be good to have a community member document this.

jcgruenhage:
@triallax thanks, that's an oversight on my part. Enabling the user repo is one of the first things I do on a fresh Chimera install. Will send in a PR to amend that.

PR #851, opened 2025-05-25
chore(docs): swap colon to dot

dublinsubway:
Caught my eye while reading docs. Not sure if the expectation was to have an example of usage or to finish a sentence; I've opted for a second option, but if you want me to add an example her (similar for JSON), let me know and I'll update.


woodruffw (approved):
Thanks, good catch!

PR #854, opened 2025-05-25
Auto-Fix Feature

mostafa:
Hey,

This PR introduces a comprehensive auto-fix feature that allows zizmor to suggest and automatically apply security fixes to GitHub Actions workflows while preserving YAML formatting and comments, on a best-effort basis.

### Auto-Fix CLI Options
* `--fix`: Apply fixes automatically if available
* `--yes`: Skip confirmation prompt when applying fixes
* `--dry-run`: Preview fixes without applying them
* `--diff`: Show a diff of changes before applying

### New Features
* `Fix`: each finding will have one or more Fixes, that either suggest and/or automatically apply fixes to the findings of each audit rule.
* Fixes for all audit rules.
* `yaml_patch`: custom YAML patch engine that maintains formatting, comments, and structure
* `oci_registry`: OCI registry client to figure out the version tags of a Docker image on any registry.
* `massive_vulnerable_workflow.yml`: a god-mode test for all the findings and fixes.
* Lots of tests.

### How-to use
```bash
# Preview fixes without applying
zizmor --fix --dry-run workflow.yml

# Apply fixes with confirmation
zizmor --fix workflow.yml

# Apply fixes automatically with diff preview
zizmor --fix --diff --yes workflow.yml
```

Note to the reviewer(s): this is not perfect and I'd be happy to change and adapt it.

Issue #28, opened 2024-09-22
Support a fix mode?


woodruffw:
Many findings are trivially fixable and insert-only, e.g. inserting `persist-credentials: false`. It would be nice to expose a `--fix` flag or similar that could apply these.

Implementation-wise, this should probably happen at tree-sitter level, since tree-sitter has mature support for modifying trees.

woodruffw:
https://github.com/woodruffw/zizmor/pull/136 and https://github.com/woodruffw/zizmor/discussions/129 are continuing to reveal that there are a _lot_ of false positives with template injection, since the `github.*` context's namespace is massive.

A fix mode could help a little bit here by offloading the work of identifying many of these: GitHub provides [default environment variables](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/store-information-in-variables#default-environment-variables) that cover many common context accesses, and we could suggest/autofix users into those.

netomi:
I would love to see a fix mode similar to e.g. ruff.

We developed a simple tool to do action pinning, see: https://github.com/eclipse-csi/octopin, but ultimately, we would like to so other fixes that your tool detects as well. So I could work on a PR to add some action pinning as initial fix. Though I have not used rust before, so that might require some iterations.

woodruffw:
> So I could work on a PR to add some action pinning as initial fix. Though I have not used rust before, so that might require some iterations.

Please feel free! The main challenge here (I think) will be doing this at the AST level, and not with regular expressions or other string patching techniques.

My plan was to look into doing that with `tree-sitter-yaml`, but I won't get to it in the near future so I welcome any efforts here :slightly_smiling_face:

netomi:
So I looked into tree-sitter for another usecase (to modify jsonnet AST), but apparently, tree-sitter does not support editing the AST it generates: https://github.com/tree-sitter/tree-sitter/discussions/2667

There are workaround that I found, but that is not what you wanna do imho.

go-yaml seems to have some support to preserve whitespace and comments when you do a roundtrip, not sure if something like this exists for rust as well. Having a full blown AST parser that supports source preserving roundtrips would be preferred ofc.

woodruffw:
https://github.com/langston-barrett/tree-sitter-edit looks like it might be exactly what we need here (hello @langston-barrett 😉)

bepp-boop:
Yea I able to edit them by capture which give them index of it and from there you can stringify them and edit it and added back in from the index position. I.E
[...]

Issue #779, opened 2025-05-10
Refactor: add a `Locatable` or similar trait


woodruffw:
A `Locatable` trait that provides `location(&self) -> SymbolicLocation` would be nice to have, since it'd allow us to express locations more generically (and with more specificity/less hacking into place).

For example, it'd allow us to do `impl Locatable for github_actions_models::Foo`, i.e. introduce symbolic locations directly.

woodruffw:
#858 reveals where `Locatable: location(&self) -> SymbolicLocation` or `Routable: route(&self) -> Route` would come in handy -- if we had widespread implementations of one of those hypothetical traits, we wouldn't need a miniature JSON Pointer impl.

woodruffw:
#871

Issue #288, opened 2024-12-13
Feature: encourage use of `shell: bash`


lishaduck:
### Pre-submission checks

- [X] I am **not** reporting a bug (crash, false positive/negative, etc). These must be filed via the bug report template.
- [X] I have looked through the open issues for a duplicate request.

### What's the problem this feature will solve?

When `shell: bash` isn't set, pipefail and friends aren't enabled. This makes it easy for failing ci to go undetected. Not a security issue per se, but it could be probably be exploited.

### Describe the solution you'd like

Warn when there's a `runs:` clause without an explicit shell argument (ideally suggest using `defaults.run.shell`, which is much less verbose and leads people into the pit of sucess).

### Additional context

I discovered zizmor yesterday, and it's great! It automates a lot of the audits I do and pointed out a lot of others I hadn't considered.

woodruffw:
Thanks for the request @lishaduck!

This is an interesting one: I agree that this isn't a security issue *per se*, but is potentially worth flagging at the "auditor" level (since that level is intended for things that need manual review).

For ongoing reference, here's where the current shell behavior in GHA is documented: <https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#custom-shell> -- `set -e` is enabled by default, but `set -o pipefail` is only enabled if `shell: bash` is explicitly set (TIL, I thought setting `bash` explicitly was exactly the same behaviorally!).

Curious what other users think about this as an audit, as well 🙂

woodruffw:
Thanks for the additional example. Yeah, I'd be OK with an auditor or pedantic-level audit for this!

lishaduck:
An even more pedantic suggestion would be to enforce `-u` as well, so `shell: bash -noprofile --norc -euo pipefail {0}`.
(`-u` fails on undefined variables, which is also a potential footgun).

That also pretty verbose, so 🤷‍♂️

PR #858, opened 2025-05-25
Foundations of the Auto-Fix Feature

mostafa:
This PR lays the foundation for the auto-fix functionality, enabling the tool to suggest and apply security fixes to GitHub Actions workflows, while preserving YAML formatting, structure, and comments on a best-effort basis.

### Key Features
* **Fix Framework**: introduces a foundational Fix abstraction that allows each audit finding to include one or more suggested or auto-applicable fixes. While all audit rules are designed to support fix generation, this PR includes fix capabilities only for the template-injection rule as a starting point. Fixes are applied as direct YAML patches using JSON pointers, avoiding the lossy YAML → JSON → Patch → YAML workflow that would strip comments and alter formatting.
* **YAML Patch Engine**: custom-built `yaml_patch` module to apply structured patches while retaining YAML comments and formatting.
* **CLI Options for Fixing**: adds a CLI flag to support auto-fixing:
    * `--fix`: apply available fixes automatically.
* **Testing**: adds extensive test coverage for patching and fix logic.

### How to Use
```bash
# Apply fixes with confirmation
zizmor --fix workflow.yml
```

### Notes
This is a large and foundational PR split from the massive #854 to make review easier. It's still evolving, and feedback is welcome, happy to adapt it further as needed. (This is the smallest possible denomination I could create.)

xref https://github.com/zizmorcore/zizmor/issues/28


crates/zizmor/src/main.rs
```diff
@@ -141,6 +143,22 @@ struct App {
...
+    #[arg(long, requires = "fix")]
+    yes: bool,
+
+    /// Show what fixes would be applied without actually writing them.
+    #[arg(long, requires = "fix")]
+    dry_run: bool,
+
+    /// Show a diff of the changes before applying them.
+    #[arg(long, requires = "fix")]
+    diff: bool,
```
woodruffw, line 161: Mind deferring these to a future PR, just to keep this one small(er)? I think we can start with `--fix` and then look at working these in.

crates/zizmor/src/main.rs
```diff
@@ -653,6 +718,241 @@ fn run() -> Result<ExitCode> {
         OutputFormat::Github => output::github::output(stdout(), results.findings())?,
     };
 
+    if app.fix {
+        use std::collections::HashMap;
```
woodruffw, line 722: Nitpick: `use` at the top of the file, please!

crates/zizmor/src/registry.rs
```diff
@@ -110,9 +110,54 @@ impl InputKey {
...
+                        .join(prefix.as_ref())
+                        .try_into()
+                        .map_err(|e| {
+                            InputError::Other(anyhow!("failed to convert prefix to UTF-8: {}", e))
+                        })?,
+                )
+            }
+        } else {
+            None
+        };
```
woodruffw, line 156: Could you provide some context for these changes? Off the top of my head, I _think_ this is going to cause problems: `given_path` is meant to be the exact path the user gives us without any canonicalization, which we do because SARIF consumers like GitHub struggle with absolute paths.

For some context, see:

* https://github.com/zizmorcore/zizmor/pull/469
* https://github.com/zizmorcore/zizmor/pull/572

(Speculating: if this change is being motivated by the `file_fixes` mapping, then I _think_ the `AuditInput` suggestion should eliminate the need for this change 🙂)

crates/zizmor/Cargo.toml
```diff
@@ -63,6 +63,8 @@ tree-sitter.workspace = true
 tree-sitter-bash.workspace = true
 tree-sitter-powershell.workspace = true
 yamlpath.workspace = true
+diff = "0.1.13"
+jsonptr = "0.7"
```
woodruffw, line 67: Nit: where does this dep get used? I tried to find it in `yaml_patch` but couldn't 🙂

Cargo.toml
```diff
@@ -48,7 +48,7 @@ reqwest = { version = "0.12.15", default-features = false }
 reqwest-middleware = "0.4.2"
 serde = { version = "1.0.219", features = ["derive"] }
 serde-sarif = "0.8.0"
-serde_json = "1.0.140"
+serde_json = { version = "1.0.140", features = ["preserve_order"] }
```
woodruffw, line 51: Not an issue, but out of curiosity: where do we need order-preserving maps here?

(Specifically I'm curious because this is for `serde_json`, not `serde_yaml`.)

crates/zizmor/src/yaml_patch/mod.rs
```diff
@@ -0,0 +1,1926 @@
...
+/// Supports the full JSON Pointer specification (RFC 6901):
+/// - Array indices are automatically detected and parsed as integers
+/// - Special characters are unescaped (~1 becomes /, ~0 becomes ~)
+/// - Empty path components are filtered out
+///
+/// # Examples
+/// - "/permissions/contents" → Query for nested key
+/// - "/jobs/0/steps/1" → Query for array indices
+/// - "/path/with~1slash" → Query for key containing "/"
+pub fn parse_json_pointer_to_query(path: &str) -> Result<yamlpath::Query, YamlPatchError> {
```
woodruffw, line 446: Architectural question: instead of needing an additional bit of syntax (JSON pointers) for patches here, perhaps we could have patch paths/routes reuse the existing `SymbolicLocation` or `Route` APIs?

If I'm thinking about this right, I _think_ that would allow the "finding" and "fix" layers to share more state/common code here. 

(This is not a hard requirement for this PR or feature, but I'm curious if you have thoughts about it! See https://github.com/zizmorcore/zizmor/issues/779 for some more context as well 🙂)

crates/zizmor/src/yaml_patch/mod.rs
```diff
@@ -0,0 +1,1926 @@
...
+            }
+        }
+        _ => {
+            let yaml_str = serde_yaml::to_string(value)?;
+            Ok(yaml_str.trim_end().to_string()) // Remove trailing newline
+        }
+    }
+}
+
+/// Check if a string needs to be quoted in YAML
```
woodruffw, line 491: Nitpick: I'm okay with removing this entirely and having the fix mode always perform quoting.

For context: Rust/LLVM struggle to reduce the number of linear scans needed by each `||` here; see [this Godbolt link](https://godbolt.org/z/Yse3f4ff8). An alternative would be to use a regex to accomplish the same thing, but IMO that'll be brittle/hard to maintain so I'm okay with a slightly conservative quoting approach instead 🙂

mostafa:
@woodruffw Thanks for the review! I'll have time to address them tomorrow.

crates/zizmor/Cargo.toml
mostafa: Leftover from previous testings. It is now manually implemented in the code.

crates/zizmor/src/main.rs
mostafa: I separated the functionality into a `fix.go` file for easier review. Also removed the extra flags.

crates/zizmor/src/registry.rs
mostafa: You are right. We no longer need this change.
https://github.com/zizmorcore/zizmor/pull/858/commits/72f704e6c21b87d58dcd562d417a1d3ef7f0d6b7

Cargo.toml
mostafa: Also a leftover from when I was trying the older approach of conversions between JSON and YAML and I needed to have this feature to preserve order of keys in the output document, because otherwise it was sorting it and messing it up entirely. Removed.

crates/zizmor/src/yaml_patch/mod.rs
mostafa: If you already have an idea how to implement this, feel free to edit this PR.

woodruffw:
Thanks, I'll do that tonight!

crates/zizmor/Cargo.toml
mostafa: https://github.com/zizmorcore/zizmor/pull/858/commits/6bfb7b208210dd53dd34c6d4b5d390f44ded9db6

crates/zizmor/src/yaml_patch/mod.rs
```diff
@@ -0,0 +1,1926 @@
...
+        assert!(result.contains("# Write access for GitHub Actions"));
+        assert!(result.contains("# Read access to issues"));
+
+        // Verify changes were applied
+        assert!(result.contains("contents: write"));
+        assert!(result.contains("packages: read"));
+        assert!(!result.contains("contents: read"));
+
+        println!("Full demo test passed!");
+        println!("📝 Original YAML had comments preserved while applying transformations");
```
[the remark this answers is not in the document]
mostafa: I replaced them with asserts wherever possible: https://github.com/zizmorcore/zizmor/pull/858/commits/480ff0504c5aa2ff7a59c4a192038e7e5eaf1e58.

crates/zizmor/src/yaml_patch/mod.rs
```diff
@@ -0,0 +1,1926 @@
...
+    run: echo "test""#;
+
+        // See what yamlpath extracts for step 0
+        let doc = yamlpath::Document::new(original).unwrap();
+        let step0_query = crate::yaml_patch::parse_json_pointer_to_query("/steps/0").unwrap();
+        let step0_feature = doc.query(&step0_query).unwrap();
+
+        println!("Step 0 extraction:");
+        println!("'{}'", doc.extract(&step0_feature));
+        println!("Byte span: {:?}", step0_feature.location.byte_span);
```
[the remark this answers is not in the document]
mostafa: Same as above.

crates/zizmor/src/main.rs
```diff
@@ -28,12 +28,14 @@ use tracing_subscriber::{EnvFilter, layer::SubscriberExt as _, util::SubscriberI
...
 mod config;
 mod finding;
+mod fix;
 mod github_api;
 mod models;
 mod output;
 mod registry;
 mod state;
 mod utils;
+pub mod yaml_patch;
```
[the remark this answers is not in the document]
mostafa: Fixed in https://github.com/zizmorcore/zizmor/pull/858/commits/9636b9980ac8f777511d32514106423b62efbdd5.

crates/zizmor/src/audit/template_injection.rs
```diff
@@ -303,14 +554,60 @@ impl Audit for TemplateInjection {
...
     fn audit_step<'doc>(&self, step: &Step<'doc>) -> anyhow::Result<Vec<Finding<'doc>>> {
-        self.process_step(step)
+        let mut findings = self.process_step(step)?;
+
+        // Add fixes to each finding for workflow steps
+        for finding in &mut findings {
+            if let Some((script, _)) = Self::scripts_with_location(step).first() {
+                // Extract the expression from the annotation
+                if let Some(location) = finding.locations.iter().find(|l| l.symbolic.is_primary()) {
+                    if let Some(expr_start) = location.symbolic.annotation.find("may expand into") {
```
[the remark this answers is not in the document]
mostafa: This is a very good point. I'll change the signature of the `injectable_template_expressions` function to include the context as well, so I can cleanly extract it for fixes.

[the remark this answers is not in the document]
mostafa: Fixed in https://github.com/zizmorcore/zizmor/pull/858/commits/df4ef43aff85d89a8e8952d0e66eafba3a20341c.

crates/zizmor/src/yaml_patch/mod.rs
```diff
@@ -0,0 +1,1926 @@
...
+            result
+        }
+    } else {
+        // For scalar values, simple key: value format
+        format!("{}: {}", key, new_value_str)
+    };
+
+    let mut result = content.to_string();
+
+    if let Some(last_idx) = last_content_line_idx {
```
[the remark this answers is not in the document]
mostafa: I thought of that as well when implementing, but was more concerned about preserving the structure of the file. I like your idea. Let me see if I can implement it.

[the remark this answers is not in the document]
mostafa: I tried two different approaches of implementing this not only for root-level, but also for nested objects (which quickly turned out to be a hard problem) by:
1. modifying the existing yaml_patch implementation to enable addition disposition.
2. recreating the whole thing with tree-sitter-yaml to get a parse tree and work on that.

Both worked to some extent, and even though the tree-sitter version provided more flexibility, yet both have the same set of issues:
1. Line shifting is a major issue and very hard to detect with the current implementation of `Fix::apply`. At the moment, the Fixes are accumulated per Finding, and applied at the end. Almost all operations change the line count (to some extent), while the Finding still refer to the old line numbers. The other side of the coin, the content reference (yamlpath?), also might pose other set of issues.
2. Operation dependency is also another problem. Later operations see the results of earlier operations, sometimes causing unexpected behavior. This is also the cause of line shifting. We have conflicts between operations.
3. Formatting issues arise when inserting at desired positions, and not the bottom.
4. Bottom disposition is the safest, at least for now.

I am currently trying to only apply this to root-level to see if I can pull this off. Otherwise, we should probably create another issue to change this later, because it is taking too long.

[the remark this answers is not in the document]
mostafa: I am convinced that this is a hard problem to solve. Let's revisit this at a later time, unless you have an idea how to properly solve this.

mostafa:
@woodruffw I am done with the changes.

crates/zizmor/src/audit/template_injection.rs
```diff
@@ -265,32 +276,292 @@ impl TemplateInjection {
...
+            }),
+        }
+    }
+
+    /// Generate a safe environment variable name from an expression
+    fn generate_env_var_name(expr: &str) -> String {
+        // Convert expressions like "github.event.issue.title" to "GITHUB_EVENT_ISSUE_TITLE"
+        // For fromJSON(secrets.foo).bar patterns, extract meaningful parts
+
+        let cleaned_expr = if expr.trim().starts_with("fromJSON(secrets.") {
```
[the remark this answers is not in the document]
mostafa: Yes. That's why I mentioned the suggestions and then fixes. For this, we can resort to suggesting how to fix it instead.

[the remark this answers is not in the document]
mostafa: Fixed in https://github.com/zizmorcore/zizmor/pull/858/commits/88e5d846a6153e30de5b00aa7dbe5cc6a995989f.

crates/zizmor/src/audit/template_injection.rs
```diff
@@ -265,32 +276,292 @@ impl TemplateInjection {
...
+            description: "Add 'shell: bash' to ensure consistent variable expansion syntax across different runners. \
+                When switching to '${VARNAME}', keep in mind that different shells have different environment variable syntaxes. \
+                In particular, PowerShell (the default shell on Windows runners) uses '${env:VARNAME}'. \
+                To avoid having to specialize your handling for different runners, you can set 'shell: bash'.".to_string(),
+            apply: apply_yaml_patch!(vec![YamlPatchOperation::MergeInto {
+                path: format!("/jobs/{}/steps/{}", job_id, step_index),
+                key: "shell".to_string(),
+                value: serde_yaml::Value::String("bash".to_string()),
+            }]),
+        }
```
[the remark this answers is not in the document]
mostafa: This makes sense! I'll defer it to #288.

crates/zizmor/src/finding/mod.rs
```diff
@@ -442,6 +442,23 @@ pub(crate) struct Determinations {
...
 }
 
+/// Represents a suggested fix for a finding.
+pub struct Fix {
+    /// A short title describing the fix.
+    pub title: String,
+    /// A detailed description of the fix.
+    pub description: String,
+    /// A function that, when called, applies the fix and returns the new content (or None if not applicable).
+    pub apply: Box<dyn Fn(&str) -> anyhow::Result<Option<String>> + Send + Sync>,
```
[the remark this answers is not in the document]
mostafa: Fixed in https://github.com/zizmorcore/zizmor/pull/858/commits/86e5736d66efc2f5db6176a1cb78654fcb557a16.

mostafa:
@woodruffw Thanks for the reviews! 🙏 

1. Experimental release is a good way to test user experience and I am all for it.
2. Let's give it a few iterations and it'd be good to go.
3. We can mention this feature (its PR, discussions, related tickets and such) in the next release notes and mention it'll be available in the upcoming release, so as to gather feedback. This is just a suggestion. You ultimately decide what to do with it.
4. I fixed your suggestions and also updated the ticket description to match the current functionality.

mostafa:
@woodruffw 
One more thing! Currently the suggestions are left to the findings. I can create Fixes with no-op apply functions with only the title and description, and then enumerate them in the end with the given suggestion. However, I guess that would be superfluous and redundant.

crates/zizmor/src/fix.rs
```diff
@@ -0,0 +1,186 @@
...
+                }
+            }
+        }
+
+        // Only proceed if there are changes to apply
+        if current_content != original_content {
+            println!("{}", format!("\nFixes").green().bold());
+            let num_fixes = file_applied_fixes.len();
+            for (ident, fix, finding) in file_applied_fixes {
+                let line_info = if let Some(line) = get_primary_line_number(finding) {
```
woodruffw, line 102: FYI you can unwrap here, rather than handling a superfluous `None` case -- the finding builder APIs enforce the presence of a primary location, so it's safe to assume one here as well 🙂 

(Alternatively you could do the unwrap within `get_primary_line_number`, that might be a little cleaner.)

woodruffw:
Thanks again @mostafa! I'll take over the PR for a bit here to clean up the lints, etc.

PR #863, opened 2025-05-27
fix: update error message to handle enterprise servers

tiymat:
While using zizmor at work today I noticed that when specifying `--gh-hostname` and failing to collect inputs from an on-prem GitHub Enterprise Server, zizmor would emit a message making it seem like it was trying to connect to github.com.


woodruffw (approved):
Nice, thanks a ton @tiymat!

PR #865, opened 2025-05-28
refactor: minimize clones in yamlpath/routing

woodruffw:
This replaces a bunch of `String` and `Cow` usage with `&str`, since in practice our yamlpath queries/routes are always bound to the overarching "document" lifetime.


woodruffw:
The next logical refactor would be to remove `Route` and replace it with `QueryBuilder`, but I'll need to think more about how disruptive that'll be (since it's technically part of the public stability commitment via `--format=json`).

PR #866, opened 2025-05-28
chore(docs): bump trophies

cunla:


docs/snippets/trophies.txt
```diff
@@ -27,6 +27,7 @@ cakephp/cakephp#18081
 canonical/operator#1656
 canonical/operator#1721
 canonical/testflinger#630
+cunla/fakeredis#391
```
woodruffw, line 30: Correct me if I'm wrong, but `cunla/fakeredis` appears to be a redirect for https://github.com/cunla/fakeredis-py -- could we use that instead?

PR #871, opened 2025-05-29
refactor: break out location handling

woodruffw:
~~WIP.~~

This is a pretty big internal refactor:

1. I've added the `Locatable` and `Routable` traits, which provide a cleaner abstraction over "things that have a `SymbolicLocation`" and "things that have a `Route`" respectively. `Locatable` is composed over `Routable`.
2. The WIP YAML patching functionality now uses `Route`s instead of a JSON-pointer-style syntax. 
3. I've rewritten/simplified the `Fix` API a bit, to make it less dynamic: there's no longer a dynamic `apply` function; instead, each `Fix` statically declares its YAML patching operations. This lets use reuse more finding-side APIs, which have `'doc` and other lifetimes.
4. I've retooled the approach to fix generation in `template-injection` -- instead of rolling multiple fixes into the same finding for a `run:` block, we give each finding its own fix where possible. We also use the expression/step APIs to be more discerning about when we produce fixes -- for now we only support fixes for `run:` blocks where the finding is for a context-only expression that has no calls or otherwise computed components. We _could_ in principle support computed indices and call-driven contexts as well, but this is for an MVP.

Doing (4) revealed #888 as a bug, which this PR does *not* attempt to fix. I'll look at that more in a follow-up.

This turns concrete/symbolic location handling into its own module, so that I can start thinking about a "routable" trait that'll notch into #858.

### DIFF
--- a/crates/zizmor/src/finding.rs
+++ b/crates/zizmor/src/finding.rs
@@ -205,7 +205,6 @@ impl<'doc> FindingBuilder<'doc> {
         self
     }
 
-    #[allow(dead_code)]
     pub(crate) fn fix(mut self, fix: Fix<'doc>) -> Self {
         self.fixes.push(fix);
         self


mostafa:
@woodruffw Well done! 👏

PR #879, opened 2025-05-30
chore(ci): Small improvements around linting

langston-barrett:
Hey there :wave: 

These are just a few small tweaks to CI/Cargo/Clippy that I thought you might find useful! See the individual commits for the motivation for each change.


woodruffw:
Awesome, thanks @langston-barrett!

PR #888, opened 2025-06-03
refactor: add RewriteFragment for format-preserving YAML string edits

woodruffw:
~~WIP.~~

I noticed this while experimenting in #871 -- it looks like there are some multiline string patches that are currently slightly off.

Specifically, this fails when the multiline string is the value of a mapping in a block-style list, e.g.:

```yaml
foo:
  - bar: |
      abc
```

Replacing `abc` with `def` *should* result in:

```yaml
foo:
  - bar: |
      def
```

...but instead we get:

```yaml
foo:
  - bar: |
    def
```

i.e. invalid YAML syntax.


woodruffw:
Example failing test here:

https://github.com/zizmorcore/zizmor/actions/runs/15407980679/job/43354253490?pr=888

CC @mostafa I'm curious if you have thoughts on how to fix this more generally -- one thought I had was that the current "leading whitespace" logic could be made more resilient by adding another feature extraction API that extracts _just_ the value from a mapping, rather than the whole `key: value`. That could then be used to get the leading whitespace from `value`, which wouldn't be confounded by `key` or a flow-list before `key`.

mostafa:
@woodruffw This is a very intriguing case. As you also noted in the test and description, multi-line text has an extra indent, unlike the rest of the YAML key/values. I like the idea, so I'll have a look to see if I can make it work, unless you already know of a fix (or has been working on it).

woodruffw:
I'm going to mess around with it a bit in #871 (which is now sprawling with changes, unfortunately) so no worries! Was just curious if you had any immediate thoughts on how to handle it 🙂

woodruffw:
Just leaving a note for myself: I think what this needs to make `Replace` more sound is for `yaml_patch` to use "exact" features from `yamlpath` everywhere, and remove the "pretty" features entirely.

Then, to handle absent values and flow-style mappings (e.g. `foo:`, `{ foo: }`, and `{ foo }`), the "exact" feature extraction API needs to really return `(Option<Feature>, Feature)` where the first optional feature is the maybe-absent value itself, and the second mandatory feature is the parent node's feature. That should, in principle, give us all of the context we need to handle each of the different replacement varieties here.

woodruffw:
To summarize a bit here:

* I've added the `RewriteFragment` operation. This operation behaves a bit like `Replace`, but with the trick that it pushes the actual replacement phase into the operation itself, meaning that we get format preservation "for free" regardless of the style or indentation of the string feature being modified.
* I've tweaked the `template-injection` audit to use `RewriteFragment` for its fix operations and added @mostafa's tests (originally in #891) to confirm that we handle multiple fixes correctly without needing to coalesce ops into a single `Fix`.
* I've made some bugfixes to `yamlpath`'s handling of some mapping edge cases; there's still one edge case where a flow mapping contains a member without a colon (e.g. `foo: { bar }`), but that's a pretty niche case.

woodruffw:
NB: This PR is again getting too big, so I'll merge after #891 and then continue my refactoring 🙂

Issue #876, opened 2025-05-30
[META] Fix mode features


woodruffw:
See #28 for original context.

This is a meta-issue for tracking some of the features/goals for a fix mode in `zizmor`.

- [ ] Build the groundwork for a fix mode: #858 
- [ ] Confirm the basic functionality of a fix mode with an initial set of audits
    - [ ] #858 includes `template-injection` as one of the first fixable audits
- [ ] Perform a stable release of `zizmor` containing fix mode support as an **experimental** feature
- [ ] Support more fixable audits
- [ ] Stabilize the fix mode and release another stable release of `zizmor` with it

Some subtasks:

- [ ] Support grouping of fixes into "safe" and "risky" (better terms welcome) categories: "safe" fixes should always apply correctly such that the underlying finding is addressed, while "risky" fixes may be unsound or not fully address the finding.
- [ ] Refactor fix/patch handling so that fixes are situated via the existing location/route APIs, rather than a JSON pointer style syntax
    - [ ] See #871 for foundations on that
- [ ] Top-level patches should be situtable in idiomatic locations rather than just at the bottom of the file; see https://github.com/zizmorcore/zizmor/pull/858/files#r2113973894 for context

woodruffw:
Sorry for the delay!

> Would you like me to add Fixes for other audit rules as well? Well, cherry-pick and clean up, to be exact. If so, which ones do you want to see first?

Yeah, that would be ideal, thanks! Keeping the PRs small (one or two audits max) would make things a lot faster for me to review 🙂 

(I'll leave it up to you to determine whether you want to block or not, but my current plan is to finish my refactoring in #871, which will probably include some API changes around routing that'll affect the `Fix` APIs. So you _might_ want to block to avoid having to deconflict later, but you should also do things as-is and deconflict as needed 🙂)

mostafa:
No worries!

I'll start creating PRs then.

(I don't think there will be massive changes to the `Fix` API. My guess is that there will be one or two new parameters added, unless you want to revamp the entire API and replace it with a new API, which is probably unlikely. I'm okay with fixing conflicts.)

woodruffw:
Sounds good, thanks!

woodruffw:
[...]

PR #891, opened 2025-06-03
Add Fix for artipacked audit rule

mostafa:
This is a follow-up PR on enabling Fixes for all the remaining audit rules and this one does it for [Artipacked](https://docs.zizmor.sh/audits/#artipacked). It adds the following block to the workflow if it finds a checkout block without it:

```yaml
with:
  persist-credentials: false
```

Since it is using a `MergeInto` operations, this fix will add a `persist-credentials: false` key/value to an existing `with` key on the checkout step.

xref #876 

@woodruffw
1. I already had the changes, but I updated the original code to use the new API (#871).
2. I used snapshot testing for this.


woodruffw:
Thanks @mostafa, I'll try and review this tonight!

Also thanks a ton for using snapshots, that makes my review easier!

FWIW you can remove the need for sidecar snapshot files by passing `@""` as the last argument to `assert_snapshot` -- that'll bake the result directly into the test, which is IMO nicer for non-integration tests like these. 

Docs on that here: https://insta.rs/docs/quickstart/#inline-snapshots

Edit: Oops, didn't mean to close.

mostafa:
@woodruffw Thanks! 🙏 

I made all snapshots inline.

mostafa:
@woodruffw One thing I noticed is that the updated version of template injection fix only applies one fix (replaces a single context), even though it creates all env-vars in the `with` block. This feel strange and kinda a limitation (contrary to the previous/my implementation) and one has to run zizmor a couple of times to get things right. (I read your comment as well in the code, and wondering how we want to deal with multiple fixes.)

woodruffw:
> One thing I noticed is that the updated version of template injection fix only applies one fix (replaces a single context), even though it creates all env-vars in the `with` block. This feel strange and kinda a limitation (contrary to the previous/my implementation) and one has to run zizmor a couple of times to get things right. (I read your comment as well in the code, and wondering how we want to deal with multiple fixes.)

Oh hmm -- this might have been an accident on my part :sweat_smile:. The intention with the refactor was to have each finding have exactly one fix, but that _all_ of the findings should compose together to properly replace all of the contexts in a piecewise fashion.

(I'm actually confused as to how it's failing that way -- that suggests that each fix's env changes are being applied, but the replacements aren't somehow?)

mostafa:
@woodruffw I'll try to fix it.

mostafa:
@woodruffw I fixed it in https://github.com/zizmorcore/zizmor/pull/891/commits/47d3be7156267ebf6199972cdd953db5d056a431.

mostafa:
@woodruffw After a few hiccups I was able to fix it.

woodruffw:
> @woodruffw After a few hiccups I was able to fix it.

Your approach looks like it works behaviorally, but it also looks like it's going back to the original architecture that I was trying to avoid 😅 -- if possible, I'd prefer to have each finding contain its own (and just its own) fix, since aggregating them together IMO breaks the conceptual link between a finding and its fix(es).

Given that `MergeInto` operations across multiple fixes appear to compose correctly, I suspect that there's a bug somewhere in either the `Replace` operation or in the patch ordering phase that causes this behavior that needs fixing, so that we don't have to mask it by doing aggregation on the `Fix` API side instead. 

I can look into that using the tests you've written here (thanks a ton for those!) -- would you mind paring this PR down to just the `artipacked` audit changes? I'll poke at the `template-injection` side, since I'm also looking at fixing the string rewriting edge cases in #888 and I suspect they'll overlap quite a bit 🙂

mostafa:
@woodruffw I suspected that you'd say this. 😆 I'll revert the changes to `template-injection`. I guess you already know which tests fail, so I guess it'd be fine to delete them.

mostafa:
@woodruffw Done! I just checked out the changes from main, so you can still access the tests.

woodruffw:
> so I guess it'd be fine to delete them.

Yes, thanks! I'll pick your template-injection tests into #888 to confirm my fix approach.

(I think I know what the underlying bug there was -- it's that each `Replace` was being done on the original copy of the feature, rather than coalescing them together.)

PR #894, opened 2025-06-05
chore: more YAML patch cleanup

woodruffw:


woodruffw:
FYI @mostafa: I've marked a few tests with `#[ignore]` here rather than leaving them uncommented, so that it's easier for us to track which patching operations still need changes/more edge case handling.

I'm curious if you have thoughts on any of these -- the flow vs. block stuff in particular seems pretty tricky to me.

mostafa:
@woodruffw I am wondering if people will use flow syntax in GitHub Actions workflows. 🤷 

If so, then we need to fix these issues by:
* adding detection of flow mapping syntax
* updating the value replacement logic to preserve flow mapping syntax
* adding support for YAML's shorthand mapping syntax
* making sure we preserve the formatting when adding new keys to existing mappings

woodruffw:
> I am wondering if people will use flow syntax in GitHub Actions workflows. 🤷

Unfortunately, I think they do -- block syntax is much more common, but flow syntax shows up in matrix definitions and `on:` blocks pretty often. For example, here's a flow-style mapping in PyCA Cryptography:

https://github.com/pyca/cryptography/blob/5cbd0b1967d6cfc8ae4dc2fc7cc7cfb993e08aa2/.github/workflows/ci.yml#L25-L58

Thankfully it's much less common with `env:` and other mappings, but for full generality it's something that we probably need to support (or at least skip in a structured manner, so that we don't produce unsound patches).

(Another variant is when people codegen their GHA workflows/actions, so we end up with either JSON-style YAML or a mix of block-and-flow layouts depending on their YAML serializer's choices.)

Your list of steps needed makes sense to me! In terms of detecting the different syntaxes used, I _think_ `yamlpath::Feature` could grow a `kind()` API or similar that produces a `FeatureKind` enum, something like:

```rust
enum FeatureKind {
  /// A single-quoted string value.
  SqString, 
  /// A double-quoted string value.
  DqString,
  /// A bare single-line string value.
  BareString,
  /// A multi-line string value.
  MultilineString,
  /// A block-style mapping.
  BlockMapping,
  /// A block-style list.
  BlockList,
  /// A flow-style mapping.
  FlowMapping,
  /// A flow-style list.
  FlowList,
  /// Any other (non-string) literal.
  OtherLiteral,
}
```

I'm curious what you think about the above -- I think that would give us the context needed, but I might be underthinking it.

Edit: The string types could also be collapsed a bit, e.g. `String(StringKind)` instead of linear variants for `SqString`, `DqString`, etc.

mostafa:
@woodruffw I totally forgot about matrix and I use [it](https://github.com/grafana/django-saml2-auth/blob/7744738a1afc8eeadd36ac619c5faf7528faa603/.github/workflows/deploy.yml#L17-L23) as well. 🤦 

I think we can start by detecting the block vs. flow styles per value, and then handle modifications based on each detected style. All block style values can technically be written in flow style, except multiline blocks (literal | and folded >). By detecting the style per value we can pinpoint our modifications and preserve the original formatting intention.

The biggest challenge is that YAML allows mixing different styles within nested structures, which means style detection must happen per value, not per key or document-level. For example:

```yaml
users:
  - name: Alice
    roles: [admin, developer]   # Flow inside a Block list item
  - { name: Carol, roles: [HR] }  # Flow mapping inside a Block list item
  - name: Bob
    roles:
      - designer                # Block inside a Block list item
      - reviewer
```

I am trying something. Let me see if I can fix this.

woodruffw:
> I think we can start by detecting the block vs. flow styles per value, and then handle modifications based on each detected style. All block style values can technically be written in flow style, except multiline blocks (literal | and folded >). By detecting the style per value we can pinpoint our modifications and preserve the original formatting intention.

That's what I was thinking too! 

And yeah, the mixed style thing is a real nightmare 😅 -- the way I was thinking about it was that we can _probably_ get away with keeping patches as local as possible to avoid having to track the different parent styles. The new "exact" feature APIs in `yamlpath` should help with that, I think.

mostafa:
I created [a draft PR](https://github.com/zizmorcore/zizmor/pull/904). The PR handles the three failing cases properly. Now, I added more tests locally, and it fails on certain cases.

mostafa:
I fixed the rest of the issues that I could find based on the new tests.

woodruffw:
Thanks! I'll do a review today!

PR #904, opened 2025-06-06
Handle flow and block style values properly

mostafa:
As discussed in #894, this PR is fixing the code to properly handle both the flow and the block style values. I added a few comprehensive GHA workflow tests to make sure everything works as expected.

xref #876


woodruffw:
Awesome, thanks @mostafa! I'll try and do a full review tonight or tomorrow.

crates/zizmor/src/yaml_patch/mod.rs
```diff
@@ -17,6 +17,75 @@ pub enum Error {
...
+    /// Single quoted scalar style: 'value'
+    SingleQuoted,
+    /// Plain scalar style: value
+    PlainScalar,
+}
+
+impl YamlStyle {
+    // Note: Helper methods removed to avoid unused code warnings
+    // The enum variants provide comprehensive YAML style detection
+}
```
woodruffw: If this `impl` block isn't needed, maybe we could drop it from this PR?

(OTOH, maybe `detect_yaml_style` and `detect_scalar_style` make sense as members?)

crates/zizmor/src/yaml_patch/mod.rs
```diff
@@ -598,28 +762,81 @@ fn apply_mapping_replacement(
...
-        (ws_start, feature.location.byte_span.1)
-    } else {
-        // Replace just the value
-        (feature.location.byte_span.0, feature.location.byte_span.1)
-    };
+    // Check if this is a flow mapping that should be handled specially
+    let trimmed_content = current_content_with_ws.trim();
+    let is_flow_mapping = trimmed_content.starts_with('{')
+        && trimmed_content.ends_with('}')
+        && !trimmed_content.contains('\n');
```
woodruffw, line 904: I might be missing something, but could this use the `YamlStyle` variants + detection APIs you've added above? I think that would be a bit easier to follow :slightly_smiling_face:

crates/zizmor/src/yaml_patch/mod.rs
```diff
@@ -630,23 +847,40 @@ fn apply_value_replacement(
...
+    // Get the byte span for precise replacement
+    let start_byte = feature.location.byte_span.0;
+    let end_byte = feature.location.byte_span.1;
+
+    // Check if we're in a flow mapping context by examining the extracted content
+    // For true flow mappings, the entire content should be a single-line flow mapping
+    let trimmed_content = current_content_with_ws.trim();
+    let is_flow_mapping = trimmed_content.starts_with('{')
+        && trimmed_content.ends_with('}')
+        && !trimmed_content.contains('\n');
```
woodruffw, line 993: Same question about reusing the `YamlStyle` APIs as above :slightly_smiling_face:

woodruffw:
This looks pretty great, thanks @mostafa! I did an initial review pass and I'll do a more in-depth one later as well -- I think the overall approach here looks great, but that we could possibly reduce the complexity a bit further by using the "exact" feature API rather than the "pretty" feature one (since that'll avoid the `key: value` weirdness where we need to skip over `key:` sometimes just to operate on `value`) :slightly_smiling_face:

mostafa:
@woodruffw Glad you like it! Let me see how I can refactor it to make it easier to maintain and possibly more readable.

crates/zizmor/src/yaml_patch/mod.rs
mostafa: Fixed in https://github.com/zizmorcore/zizmor/pull/904/commits/001d8b033703944ae885e77a7ecf99511d106452.

crates/zizmor/src/yaml_patch/mod.rs
mostafa: Same as above.

crates/zizmor/src/yaml_patch/mod.rs
mostafa: Same as above.

mostafa:
@woodruffw I moved the functions into the `YamlStyle` implementation. I'll leave the pretty vs. exact part to you.

woodruffw:
I did a more in-depth review here, and added some more failing tests for edge cases in flow mappings -- we'll need to handle multiline flow mappings, trailing commas, plus weirdness where the user chooses to format their flow mapping in funky ways, like:

```yaml
foo: {
  bar: abc, baz: def,
  qux: ghi, zam: jkl
}
```

I'm not 100% sure how to handle `Op::Add` on that last one, though :sweat_smile: -- one idea is that it's _probably_ optimal to insert the new mapping pair on a new line with matching indentation, e.g.:

```yaml
foo: {
  bar: abc, baz: def,
  qux: ghi, zam: jkl,
  zip: mno
}
```

I'll look into making some of these changes now.

mostafa:
@woodruffw Nice work!

woodruffw:
> @woodruffw Nice work!

Thanks! It's not quite done 😅 -- I've added some more style cases, but the test cases themselves are still unfixed. I'll have more time to experiment with it today.

(In thinking about it, I _think_ I'm OK with not handling the multiline flow cases _as long_ as we can emit a reasonable error message for them -- they're weird/unusual enough that we can document them as a limitation that users have to manually rewrite their workflows to get fixes for.)

mostafa:
> Thanks! It's not quite done 😅 

Still.

> I think I'm OK with not handling the multiline flow cases as long as we can emit a reasonable error message for them

Agreed. Let's not overcomplicate this because someone decided to use a "weird/unusual" syntax.

crates/zizmor/src/yaml_patch/mod.rs
```diff
@@ -474,6 +527,148 @@ fn indent_multiline_yaml(content: &str, base_indent: &str) -> String {
...
+        format!("\n{}{}: {}", indent, key, indented_value)
+    } else {
+        format!("\n{}{}: {}", indent, key, new_value_str)
+    };
+
+    // Find the insertion point
+    let insertion_point = if is_list_item {
+        // For list items, we need to find the end of the actual step content,
+        // not including trailing comments that yamlpath may have included
+        find_step_content_end(content, &feature, &doc)
```
woodruffw: Flagging for myself: I want to push this fix up into `yamlpath`, since it's arguably a deficiency in the feature abstraction that shouldn't need to be handled at the patching layer.

(More generally, it's arguably a bug in the tree-sitter YAML parser, which attaches comments in pretty unintuitive places. But the buck stops with `yamlpath` 🙂)

woodruffw: I got halfway with this: pushing it all the way into `yamlpath` would be nontrivial, but `yamlpath::Document` now caries a `LineIndex` that makes this and other operations much simpler.

woodruffw:
OK, I think this is pretty much good to go. Some key changes:

1. I've refactored the `Op::Add` helpers to operate on "exact" features + only modify the feature itself, and then return the modified feature back to the main patch handler for reinsertion.
2. I've given each `yamlpath::Document` a `LineIndex` that makes offset-coordinate mapping more resilient (and faster), and allows us to replace/rewrite the various whitespace/comment helpers while minimizing allocations.
3. Every `yamlpath::Feature` now has a `kind() -> FeatureKind`, which we use to inform the style detection. That could also be made more feature complete, but the current split seems fine enough to me.
4. I've fixed some more `Op::Add` edge cases with flow mappings, and rewrote the core `handle_flow_mapping_addition` so that it round-trips through a flow serializer instead of attempting to situate a flow-level patch. That will potentially disturb some user formatting but IMO not to a degree that's worth fixing (especially since comments are preserved fine).
5. I've added a _lot_ of tests for the rewritten whitespace/feature/line index helpers that'll hopefully suss out more bugs.

In terms of next steps here, I think I want to further decompose the ops, but I need to think more about that.

PR #911, opened 2025-06-07
feat: handle default envvars when fixing template injections

woodruffw:
We need to do this for two reasons:

1. It makes the patches smaller;
2. It's technically more correctly, since the GHA runner's values for these variables always take precedence anyways (at least, according to GitHub's docs).

For (2):

> You can't overwrite the value of the default environment variables named GITHUB_* and RUNNER_*. Currently you can overwrite the value of the CI variable. However, it's not guaranteed that this will always be possible. For more information about setting environment variables, see [Defining environment variables for a single workflow](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/store-information-in-variables#defining-environment-variables-for-a-single-workflow) and [Workflow commands for GitHub Actions](https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-environment-variable).

Per: <https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/store-information-in-variables#default-environment-variables>

### DIFF
--- a/crates/github-actions-expressions/src/context.rs
+++ b/crates/github-actions-expressions/src/context.rs
@@ -43,14 +43,6 @@ impl<'src> Context<'src> {
         parent.parent_of(self)
     }
 
-    /// Returns the tail of the context if the head matches the given string.
-    pub fn pop_if(&self, head: &str) -> Option<&str> {
-        match self.parts.first()? {
-            Expr::Identifier(ident) if ident == head => Some(self.raw.split_once('.')?.1),
-            _ => None,
-        }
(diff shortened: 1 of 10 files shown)


woodruffw:
This changeset also made me realize that the current `env_is_static` logic is slightly wrong, since we only check the literal `env` blocks and not the runner's default environment. So I'll also fix that as part of this PR (and flip the default static assumption, since we can't really assume anything thanks to `GITHUB_ENV`).

PR #915, opened 2025-06-07
Apply fixes based on persona

mostafa:
This PR is an attempt to implement persona-based fixing, based on the https://github.com/zizmorcore/zizmor/issues/876#issuecomment-2952925472 by @woodruffw. Each Fix will only be applied if the given persona (or default) matches that of the Fix.


crates/zizmor/src/fix.rs
```diff
@@ -59,6 +63,11 @@ pub fn apply_fixes(results: &FindingRegistry, registry: &InputRegistry) -> Resul
         // First, try to apply each fix independently to the original content
         // to collect which fixes can be applied successfully
         for (ident, finding, fix) in fixes {
+            // Skip fixes that are not applicable to the current persona
+            if fix.persona <= persona {
```
mostafa: @woodruffw Do you think this is a good way to do this, considering that we are comparing integers (enum fields)?

woodruffw: The comparison looks right to me, but I _think_ we don't want the _fix_ to have its own persona :sweat_smile: -- instead it should be assumed to have the parent finding's persona, since `--persona=pedantic` should behave consistently across both fixes and findings :slightly_smiling_face:

mostafa: Makes sense! I'll fix this.

mostafa: I tried: https://github.com/zizmorcore/zizmor/pull/915/commits/17b30109f7bc208a136b576692ff592a8a32b34e.

woodruffw: Thanks, I'll take a look in a moment!

woodruffw:
Oh, this was dumb of me -- I forgot that `FindingRegistry::findings()` already internally filters/sorts findings based on their persona, so we don't need any additional processing here :facepalm: 

Sorry for the wild goose chase here @mostafa!

PR #921, opened 2025-06-08
Add Fix for bot-conditions audit rule

mostafa:
This PR adds Fixes to bot-conditions audit rule, as suggested by the documentation. The bot-conditions audit rule provides three fixes that work together to secure the workflow:

1. `RewriteFragmet` for actor context: fine-grained rewrites of `github.actor` to `github.event.pull_request.user.login` in conditions to ensure we check the PR author rather than the last actor.
2. ~`Replace` for trigger: Changes `pull_request_target` to `pull_request` to prevent running with elevated permissions.~
3. ~`Remove` for conditions: Optionally removes bot conditions entirely if they're not needed.~

These operations are applied in sequence to transform the workflow from using spoofable bot checks to a more secure configuration.

xref #876


crates/zizmor/src/audit/bot_conditions.rs
```diff
@@ -200,11 +215,114 @@ impl BotConditions {
...
+                    operation: Op::Replace(serde_yaml::Value::String(
+                        expr.replace("github.actor", "github.event.pull_request.user.login")
+                            .replace("GitHub.actor", "github.event.pull_request.user.login")
+                            .replace("GitHub.ACTOR", "github.event.pull_request.user.login")
+                            .replace("github.triggering_actor", "github.event.pull_request.user.login")
+                            .replace("GitHub.triggering_actor", "github.event.pull_request.user.login")
+                            .replace("GitHub.TRIGGERING_ACTOR", "github.event.pull_request.user.login")
+                            .into()
+                    )),
+                }],
```
woodruffw: Unfortunately, I think this is going to be insufficiently general -- we need to handle all possible valid productions of things like `github.actor`, including `github['actor']`, `GITHUB['ACTOR']`, etc.

I think a good approach here might be to try and do the replacement here like we do in `template-injection` -- we need the exact original fragment, and then we can use an `Op::RewriteFragment` to replace it regardless of the syntax/capitalization the user wrote originally.

crates/zizmor/src/audit/bot_conditions.rs
```diff
@@ -200,11 +215,114 @@ impl BotConditions {
...
+                _key: step.location().key,
+                patches: vec![Patch {
+                    route: step.route().with_keys(&["if".into()]),
+                    operation: Op::Remove,
+                }],
+            })
+        } else {
+            None
+        }
+    }
```
woodruffw: I might be misunderstanding the intent here, but I don't _think_ we want to propose removing the entire check here 😅 -- doing so would potentially make some workflows more exploitable, since e.g. an automerge workflow without any actor check would allow _any_ triggerer to induce a merge.

crates/zizmor/src/audit/bot_conditions.rs
mostafa: Makes sense! I figured you'd say that, as this is a naive way to fix it. 😬

crates/zizmor/src/audit/bot_conditions.rs
mostafa: I'll fix this and the above.

crates/zizmor/src/audit/bot_conditions.rs
mostafa: Fixed in https://github.com/zizmorcore/zizmor/pull/921/commits/37606ca7a5c40932746aa79135dca825436dc513.

crates/zizmor/src/audit/bot_conditions.rs
mostafa: Fixed in https://github.com/zizmorcore/zizmor/pull/921/commits/37606ca7a5c40932746aa79135dca825436dc513.

woodruffw:
Thanks @mostafa! I'll do another review tomorrow most likely.

(I'm going to look into improving the expression APIs more generally -- they _should_ be returning spanned AST nodes so that we don't ever have to do any offset guesswork.)

crates/zizmor/src/audit/bot_conditions.rs
```diff
@@ -200,11 +228,245 @@ impl BotConditions {
...
+                                false
+                            }
+                        } else {
+                            false
+                        }
+                    }) {
+                        "github.event.pull_request.user.login"
+                    } else {
+                        // This is an actor_id context, replace with user.id
+                        "github.event.pull_request.user.id"
```
woodruffw: For full generality I _think_ we need to handle different event types here. Specifically, I think we need to look at the `on` trigger block and see which event(s) are used to suggest the right context here (e.g. `github.event.comment.user.login`). Otherwise we'll end up suggesting a mismatched context here 🙂

crates/zizmor/src/audit/bot_conditions.rs
```diff
@@ -200,11 +228,245 @@ impl BotConditions {
...
+                    let mut map = serde_yaml::Mapping::new();
+                    map.insert(
+                        serde_yaml::Value::String("pull_request".into()),
+                        serde_yaml::Value::Mapping(Default::default()),
+                    );
+                    serde_yaml::Value::Mapping(map)
+                }),
+            }],
+        }
+    }
```
woodruffw: I like the idea for this fix, but I think we should leave it out for now 😅 -- it overlaps with the fix we'll want to suggest for `dangerous-triggers`, plus it's hard to suggest confidently (since a lot of users unfortunately do need `pull_request_target` for 3p triggers).

woodruffw:
Hey @mostafa, sorry for the delay here! I've been working on some expression spanning APIs that  will hopefully make future interactions with the expression APIs + getting the "raw" expression from a parsed AST a bit easier.

crates/zizmor/src/audit/bot_conditions.rs
```diff
@@ -200,11 +228,245 @@ impl BotConditions {
...
+
+    /// Create a fix for job-level conditions
+    fn create_replace_actor_fix_for_job<'doc>(
```
woodruffw: Q: can we unify this fix builder with `create_replace_actor_fix`? They seem to share a lot of the same logic, but I might be missing some key differences between the two.

mostafa:
@woodruffw Thanks for the reviews! 🙏 I rebased from main and addressed your comments.

crates/zizmor/src/audit/bot_conditions.rs
```diff
@@ -114,6 +134,195 @@ impl Audit for BotConditions {
...
+    fn has_relevant_events(workflow: &Workflow) -> bool {
+        use github_actions_models::workflow::Trigger;
+
+        match &workflow.on {
+            Trigger::BareEvent(event) => Self::is_relevant_event(event),
+            Trigger::BareEvents(event_list) => event_list
+                .iter()
+                .any(Self::is_relevant_event),
+            Trigger::Events(event_map) => {
+                !matches!(event_map.issue_comment, OptionalBody::Missing)
```
woodruffw: Out of curiosity, why are we filtering out the `OptionalBody::Missing` case here? I don't think that affects the overall event payload in this context.

crates/zizmor/src/audit/bot_conditions.rs
```diff
@@ -256,4 +663,397 @@ mod tests {
...
+                on: issue_comment
+
+                jobs:
+                  test:
+                    runs-on: ubuntu-latest
+                    if: github.actor == 'dependabot[bot]'
+                    steps:
+                      - name: Test Step
+                        if: github.event.comment.user.login == 'dependabot[bot]'
+                        run: echo "hello"
```
woodruffw: NB: I updated these tests to use snapshots instead of substring checks, and I noticed that we only seem to be fixing the step-level `if` and not the job-level `if`, even though there _should_ be two separate fixes applied here. Any idea what's going on here @mostafa?

crates/zizmor/src/audit/bot_conditions.rs
mostafa: AFAIK, the `OptionalBody::Missing` properly checks if any event is present as a trigger (either with or without configuration), which determines whether bot conditions are relevant for workflows that have these events as triggers. I might have misunderstood this comment:

https://github.com/zizmorcore/zizmor/blob/ec2f674ca324f191307c5d5421a448ca65811a5c/crates/github-actions-models/src/workflow/event.rs#L156-L159

crates/zizmor/src/audit/bot_conditions.rs
mostafa: What I understood from [your comment](https://github.com/zizmorcore/zizmor/pull/921#discussion_r2160828999) is to remove this in favor of `dangerous-triggers`. If that wasn't your intention, I can revert the change.

> I like the idea for this fix, but I think we should leave it out for now 😅 -- it overlaps with the fix we'll want to suggest for `dangerous-triggers`, plus it's hard to suggest confidently (since a lot of users unfortunately do need `pull_request_target` for 3p triggers).

crates/zizmor/src/audit/bot_conditions.rs
woodruffw: Oh, my bad -- I missed the *negation* of `matches!` here, so I thought we were _only_ matching the `Missing` variant and not everything _except_ the `Missing` variant. This makes sense to me now!

mostafa: No worries!

crates/zizmor/src/audit/bot_conditions.rs
woodruffw: Hmm, I think maybe we're talking about two different things -- that comment was about not patching `on:` because we'll want to do that with `dangerous-triggers`, but we _should_ be patching every relevant `if:` clause.

mostafa: Makes sense now! I'll fix it tonight.

mostafa: Fixed in https://github.com/zizmorcore/zizmor/pull/921/commits/b0dd87fb1d9a8af3dbd09b4696fdb8211da03420.

woodruffw: Thanks, I'll do another pass today!

woodruffw:
Thanks a ton @mostafa! I deduped the patch generation a bit, but otherwise this is good to go 🙂

PR #923, opened 2025-06-08
Add Fix for cache-poisoning audit rule

mostafa:
The Fixes generated by this audit suggest how to disable caching in GitHub Actions workflows when publishing artifacts. They use the `MergeInto` operation to update the relevant action's `with` block, setting the cache-related field (e.g., `enable-cache`) to a value (such as `'false'` or `true`) that disables caching according to the action's configuration. For non-configurable actions, the Fix suggests manual removal or replacement.


mostafa:
@woodruffw I fixed the issues and reverted most of the changes to `coordinate.rs` and I think it is ready. Feel free to change it however you see fit, especially for dispositions.

woodruffw:
Thanks @mostafa! I'll be able to review this after the weekend.

crates/zizmor/src/audit/cache_poisoning.rs
```diff
@@ -57,7 +60,7 @@ static KNOWN_CACHE_AWARE_ACTIONS: LazyLock<Vec<ActionCoordinate>> = LazyLock::ne
         ActionCoordinate::Configurable {
             uses_pattern: "astral-sh/setup-uv".parse().unwrap(),
             control: ControlExpr::single(
-                Toggle::OptOut,
+                Toggle::OptIn,
```
woodruffw: Hmm, why did this change? FWICT `setup-uv`'s behavior is to enable the cache by default on github-hosted runners, so this should probably remain `OptOut`:

https://github.com/astral-sh/setup-uv/blob/d4219d1620c011976ed77813dca7d21d6f676d73/src/utils/inputs.ts#L25-L31

crates/zizmor/src/audit/cache_poisoning.rs
```diff
@@ -284,6 +287,91 @@ impl CachePoisoning {
...
+
+        match matching_coord {
+            ActionCoordinate::NotConfigurable(_pattern) => {
+                // For non-configurable actions, suggest removing or replacing
+                Some(Fix {
+                    title: "Remove or replace cache-enabled action".to_string(),
+                    key: step.location().key,
+                    disposition: FixDisposition::default(),
+                    patches: vec![], // No automatic fix
+                })
```
woodruffw: I think we should make this `None` for now -- we probably shouldn't flag a finding with a fix if we can't suggest any patches for it 🙂

crates/zizmor/src/audit/cache_poisoning.rs
```diff
@@ -284,6 +287,91 @@ impl CachePoisoning {
...
+                            "Set '{field_name}' to 'false' to disable caching in this publishing workflow."
+                        ),
+                    ),
+                    (Toggle::OptOut, ControlFieldType::String) => (
+                        serde_yaml::Value::String("false".to_string()),
+                        format!("Set {field_name}: false to disable caching"),
+                        format!(
+                            "Set '{field_name}' to 'false' to disable caching in this publishing workflow."
+                        ),
+                    ),
```
woodruffw: Unfortunately I think this two can't be made sound generally: a "string" control field means that the action expects something other than a boolean, and we don't necessarily know what it is.

`setup-node` is a good example of this: it expects `cache: ''` to disable the cache, whereas `cache: 'false'` will cause the action to fail with an opaque error about the "false" package manager not being supported.

My suggestion is to make these `None` for now 🙂

crates/zizmor/src/audit/cache_poisoning.rs
mostafa: Fixed in https://github.com/zizmorcore/zizmor/pull/923/commits/2bf8160510c597706fde371d87883d95e0b44907.

crates/zizmor/src/audit/cache_poisoning.rs
mostafa: Fixed in https://github.com/zizmorcore/zizmor/pull/923/commits/8edcf07cde9ade2096e13562f8278c1daed81bf6.

crates/zizmor/src/audit/cache_poisoning.rs
mostafa: Fixed in https://github.com/zizmorcore/zizmor/pull/923/commits/44d80c43a25754fc0d5c7134bb848e99ed36c683.

mostafa:
@woodruffw Thanks for the review! I fixed the issues.

woodruffw:
Thanks @mostafa, I'll do another review tonight!

PR #948, opened 2025-06-16
docs: Unfix accidentally-fixed pinning example

SyntaxColoring:
This example in the docs was deliberately using a ref pin (`@v3`) for the sake of demonstrating `unpinned-uses`, but it got fixed accidentally by #817.

It unfortunately seems like this is likely to happen again if `pinact` is run again. I don't know what the right long-term solution is.

### DIFF
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -395,7 +395,7 @@ sensitive `zizmor`'s analyses are:
     as its pin instead of a hashed pin:
 
     ```yaml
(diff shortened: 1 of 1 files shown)


woodruffw:
Oops, great catch. I'll look into a way to exclude that line when running `pinact` 🙂

woodruffw (approved):
Thanks @SyntaxColoring!

Issue #569, opened 2025-02-25
[BUG]: couldn't parse expression


hugovk:
### Pre-submission checks

- [x] I am **not** filing a feature request. These should be filed via the feature request form instead.
- [x] I have looked through the [open issues](https://github.com/woodruffw/zizmor/issues?q=is%3Aissue+is%3Aopen+) for a duplicate report.

### Expected behavior

I expect zizmor not to give the following parse warnings on our admittedly-complex workflow files.

### Actual behavior

```console
❯ zizmor .
 INFO zizmor: skipping impostor-commit: can't run without a GitHub API token
 INFO zizmor: skipping ref-confusion: can't run without a GitHub API token
 INFO zizmor: skipping known-vulnerable-actions: can't run without a GitHub API token
 INFO audit: zizmor: 🌈 completed .github/workflows/reusable-windows.yml
 INFO audit: zizmor: 🌈 completed .github/workflows/reusable-ubuntu.yml
 INFO audit: zizmor: 🌈 completed .github/workflows/require-pr-label.yml
 INFO audit: zizmor: 🌈 completed .github/workflows/new-bugs-announce-notifier.yml
 INFO audit: zizmor: 🌈 completed .github/workflows/lint.yml
 INFO audit: zizmor: 🌈 completed .github/workflows/tail-call.yml
 INFO audit: zizmor: 🌈 completed .github/workflows/verify-ensurepip-wheels.yml
 INFO audit: zizmor: 🌈 completed .github/workflows/jit.yml
 INFO audit: zizmor: 🌈 completed .github/workflows/reusable-macos.yml
 INFO audit: zizmor: 🌈 completed .github/workflows/add-issue-header.yml
 INFO audit: zizmor: 🌈 completed .github/workflows/stale.yml
 WARN audit:audit{input=Workflow(file://./.github/workflows/reusable-context.yml)}: zizmor::audit::overprovisioned_secrets: couldn't parse expression: #        fromJSON(needs.build-context.outputs.run-tests)
      #        && 'truthy-branch'
      #        || 'falsy-branch'
      #
 WARN audit:audit{input=Workflow(file://./.github/workflows/reusable-context.yml)}: zizmor::audit::unredacted_secrets: couldn't parse expression: #        fromJSON(needs.build-context.outputs.run-tests)
      #        && 'truthy-branch'
      #        || 'falsy-branch'
      #
 INFO audit: zizmor: 🌈 completed .github/workflows/reusable-context.yml
[...]

PR #958, opened 2025-06-20
template-injection: add subfeatures

woodruffw:
Following #949.

### DIFF
--- a/crates/github-actions-expressions/src/context.rs
+++ b/crates/github-actions-expressions/src/context.rs
@@ -13,22 +13,24 @@ use super::{Expr, SpannedExpr};
 /// identifier.
 #[derive(Debug, PartialEq)]
 pub struct Context<'src> {
-    raw: &'src str,
     /// The individual parts of the context.
     pub parts: Vec<SpannedExpr<'src>>,
 }
 
 impl<'src> Context<'src> {
-    pub(crate) fn new(raw: &'src str, parts: impl Into<Vec<SpannedExpr<'src>>>) -> Self {
+    pub(crate) fn new(parts: impl Into<Vec<SpannedExpr<'src>>>) -> Self {
         Self {
-            raw: raw.trim(),
             parts: parts.into(),
         }
     }
 
-    /// Returns the raw string representation of the context.
-    pub fn as_str(&self) -> &'src str {
-        self.raw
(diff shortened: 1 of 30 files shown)


woodruffw:
I still need to de-regress #569 with this change:

```
template-injection failed on reusable-tsan.yml

Caused by:
    failed to find subfeature 'fromJSON(inputs.free-threading)
        && '_free_threading'
        || ''
      ' in feature '|
            echo "TSAN_OPTIONS=log_path=${GITHUB_WORKSPACE}/tsan_log suppressions=${GITHUB_WORKSPACE}/Tools/tsan/suppressions${{
                fromJSON(inputs.free-threading)
                && '_free_threading'
                || ''
              }}.txt handle_segv=0" >> "$GITHUB_ENV"
            echo "CC=clang" >> "$GITHUB_ENV"
            echo "CXX=clang++" >> "$GITHUB_ENV"'
```

This is caused by the subfeature having multiple lines, which don't match the exact substring on the YAML side (since the extracted YAML feature has extra indentation to make it syntactically correct). The subfeature on the other hand comes from the parsed YAML, so we don't see the indentation difference until it's too late.

I'll need to think about the best way to handle this -- it's interestingly non-trivial because the YAML feature equivalent of the subfeature is actually longer/has more bytes in it.

Edit: Thinking out loud: one idea is to match the span on the first line, allow all intermediate lines, and then match the span on the final line. Not pretty, but doesn't require any special normalization or other steps that would mess with the concrete feature.

woodruffw:
De-regressed, but I'm not very happy with how brittle the approach is. I need to think about it more.

woodruffw:
I thought of another solution to this, one that should be less brittle: instead of trying to match the "head" and "tail" of the subfeature's fragment, we could support pattern-style fragments that accept differences in whitespace.

This would still be mostly useful only in the case of multi-line subfeature fragments, but would allow us to transform a fragment like this:

```
foo.bar
  && abc
  || def
```

into:

```
foo\.bar\s+&&\s+abc\s+||\s+def
```

...which would then match correctly, regardless of how the subfeature is indented on the YAML syntax side of things.

This should be pretty easy to do by walking an `Expr<'src>` and building up a `Regex` with the appropriate APIs from the `regex` crate (incl. `regex::escape`).

PR #962, opened 2025-06-22
chore(tests): update default branch for homebrew actions

chenrui333:
chore(workflows): update default branch for homebrew actions

---

update per default branch name changes

https://github.com/Homebrew/actions/pull/704


woodruffw (approved):
Thanks @chenrui333! FYI there's no need to keep these updated in the future -- these are backstop tests for parsing behavior, so they don't need to be semantically correct w/r/t tags 🙂

PR #966, opened 2025-06-24
chore: refactor fix input key handling

woodruffw:
This refactors `zizmor::fix` to use each `Fix::key` instead of trying to infer it from the parent `Finding`'s first location. In doing so we also save ourselves a handful of clones.


woodruffw:
CC @mostafa for viz 🙂

mostafa:
@woodruffw Thanks for letting me know! 🙏

PR #972, opened 2025-06-25
feat: new audit: ref-version-mismatch

### LINKED ISSUE #643: Feature: detect mismatches between pinned full-length commit SHA and version tag
### Pre-submission checks

If an action is pinned to a full-length commit SHA and does not have an inline comment with a matching version tag, it would be great for zizmor to raise a warning.

This is important for Dependabot. If the full-length commit SHA and tag are mismatched, Dependabot will update the SHA and not adjust the tag. (It seems reasonable to me for Dependabot not to make assumptions and overwrite in-line comments it does not understand.)

E.g., here is a Dependabot commit based on a mismatch:
```
-      - uses: step-security/harden-runner@4d991eb9b905ef189e4c376166672c3f2f230481 # v2.8.0
+      - uses: step-security/harden-runner@c6295a65d1254861815972266d5933fd6e532bdf # v2.8.0
```
and here is what Dependabot does if the initial tag is matched correctly:
```
-    - uses: step-security/harden-runner@4d991eb9b905ef189e4c376166672c3f2f230481 # v2.11.0
+    - uses: step-security/harden-runner@c6295a65d1254861815972266d5933fd6e532bdf # v2.11.1
```

The commit SHA would need to be checked against the upstream repo to verify that it is the same commit which introduces the tag.

### Additional context

_No response_

segiddins:
Adds a new audit that detects when commit SHAs don't match their corresponding version comment tags in GitHub Actions workflows.

This addresses a case where Dependabot will not pick up on potential updates when the ref and version comment does not match.

Fixes #643


crates/zizmor/src/audit/ref_version_mismatch.rs
```diff
@@ -0,0 +1,130 @@
...
+        // Matches "# tag=v2.8.0" or "# tag=v1.2.3"
+        Regex::new(r"#\s*tag\s*=\s*(v\d+(?:\.\d+)*(?:\.\d+)?)").unwrap(),
+        // Matches "# v2.8.0"
+        Regex::new(r"#\s*(v\d+(?:\.\d+)*(?:\.\d+)?)").unwrap(),
+        // Matches version without 'v' prefix: "# tag=2.8.0"
+        Regex::new(r"#\s*tag\s*=\s*(\d+(?:\.\d+)*(?:\.\d+)?)").unwrap(),
+        // More flexible: "# version: 2.8.0"
+        Regex::new(r"#\s*(?:version|ver)\s*[:=]\s*(v?\d+(?:\.\d+)*(?:\.\d+)?)").unwrap(),
+    ]
+});
```
woodruffw, line 40: Nitpick: you can use a `RegexSet` here instead of a `Vec<Regex>` to turn N haystack scans into 1: https://docs.rs/regex/latest/regex/struct.RegexSet.html

crates/zizmor/src/audit/ref_version_mismatch.rs
```diff
@@ -0,0 +1,130 @@
...
+        let Some(client) = state.github_client() else {
+            return Err(AuditLoadError::Skip(anyhow!(
+                "can't run without a GitHub API token"
+            )));
+        };
+
+        Ok(Self { client })
+    }
+
+    fn audit_step<'doc>(&self, step: &Step<'doc>) -> anyhow::Result<Vec<Finding<'doc>>> {
```
woodruffw: NB: We'll probably also need `audit_composite_step` (for `uses:` in composite actions) and `audit_reusable_job` (for reusable workflows) here.

The `process_step` pattern (e.g. in `forbidden-uses`) might be a good reference for deduping these 🙂

crates/zizmor/src/main.rs
```diff
@@ -627,6 +627,7 @@ fn run() -> Result<ExitCode> {
     register_audit!(audit::dangerous_triggers::DangerousTriggers);
     register_audit!(audit::impostor_commit::ImpostorCommit);
     register_audit!(audit::ref_confusion::RefConfusion);
+    register_audit!(audit::ref_version_mismatch::RefVersionMismatch);
```
woodruffw: Nitpick: new audits are added at the bottom!

(Elsewhere it's alphabetic, here the registry records them in order of addition.)

crates/zizmor/src/audit/ref_version_mismatch.rs
```diff
@@ -0,0 +1,130 @@
...
+            return Ok(findings);
+        };
+
+        // Only check steps that have commit refs (not symbolic refs like v1.0.0)
+        let Some(commit_sha) = uses.commit_ref() else {
+            return Ok(findings);
+        };
+
+        // For now, we'll search the entire document source for version patterns
+        // that might correspond to this uses line. This is a simplified approach.
```
woodruffw: Q: is this comment accurate still? It looks like the implementation below doesn't scan the whole doc source, only the comment next to the `uses:` clause (which is the right way to do it!)

crates/zizmor/src/audit/ref_version_mismatch.rs
```diff
@@ -0,0 +1,130 @@
...
+                    .longest_tag_for_commit(&uses.owner, &uses.repo, commit_sha)?
+            {
+                annotation.push_str(&format!(
+                    ", try updating the comment to {:?}",
+                    suggestion.name
+                ));
+            }
+            findings.push(
+                Self::finding()
+                    .severity(Severity::Low)
```
woodruffw: I could see an argument for this being medium -- it breaks tools like Dependabot and makes people think they're up to date/using a specific version, but actually aren't. OTOH I don't feel super strong about it; thoughts?

crates/zizmor/src/finding/location.rs
```diff
@@ -519,6 +519,12 @@ impl Comment<'_> {
     }
 }
 
+impl AsRef<str> for Comment<'_> {
+    fn as_ref(&self) -> &str {
+        self.0
+    }
+}
```
woodruffw: Flagging: not something that needs doing in this PR, but I'd be nice to avoid puncturing the newtype here and instead add the `VERSION_COMMENT_PATTERNS` matching within `Comment`'s own interface. But that can be a follow-on refactor.

crates/zizmor/src/audit/ref_version_mismatch.rs
segiddins: can't because we reference the captures

crates/zizmor/src/github_api.rs
```diff
@@ -212,29 +212,46 @@ impl Client {
...
+        }
+
+        let Some(mut tag_ref) = self
+            ._get_ref(owner, repo, &format!("ref/tags/{git_ref}"))
+            .await?
+        else {
+            return Ok(None);
+        };
+
+        while !matches!(tag_ref.object.r#type, Some(ObjectType::Commit)) {
```
segiddins: isnt git fun...
should this maybe have a limit to avoid tag cycles?

woodruffw: Hmm, I _think_ we could maybe avoid this loop entirely by switching endpoints here -- instead of using the "get a reference" endpoint we could use the "get a commit" endpoint, which should always produce a commit object.

This specifically: https://docs.github.com/en/rest/commits/commits?apiVersion=2022-11-28#get-a-commit

woodruffw: (The docs don't say it super clearly, but `gh api /repos/Swatinem/rust-cache/commits/refs/tags/v2.7.8` works and similar for `refs/heads/...`)

segiddins: So the issue with that endpoint is it doesn't return 404 when the ref doesnt exist, and instead it returns a 422, which is indistinguishable from e.g. running out of rate limit usage

woodruffw: Huh, that seems 100% wrong from GitHub. Let me see if I can get some attention from them on that.

(I'd definitely prefer to have this use a single request without a loop if possible, for similar rate limit reasons. So maybe they can fix this somewhat quickly.)

woodruffw:
(Marking as blocked until GitHub fixes the REST API bug here.)

crates/zizmor/src/github_api.rs
eslerm: Has this issue been raised to GitHub? If not, I'm happy to file and raise awareness.

(Also, than you so much for working on this new feature!!)

woodruffw: Yep, I’ve flagged it with some people I know, who confirmed that it’s an issue. But certainly having more people raise awareness of it would probably increase its relative priority internally 🙂

logan-porelle: Hiya folks, @eslerm raised this to me at GitHub and I've been talking to our API team internally. My expectation is that we won't change the failed return from 422 to 404 as that would be seen as a breaking change.

I will note that 422 **shouldn't** be the return you get when you run out of rate limits, that should be `429` or `403`. `422` is used as a validation error as documented [here](https://docs.github.com/en/rest/using-the-rest-api/troubleshooting-the-rest-api?apiVersion=2022-11-28#validation-failed). If you are getting 422 on rate limit usage please let me know so we can investigate.

Finally on a 422 response you can inspect the message returned to validate that it states the ref doesn't exist.

woodruffw: Hi @logan-porelle, thanks for responding! I really appreciate it.

> My expectation is that we won't change the failed return from 422 to 404 as that would be seen as a breaking change.

Out of curiosity, is there a reason the regression itself isn't considered a breaking change? I recall this endpoint working as expected years ago (probably close to a decade ago...). It would make perfect sense to me if the answer was "because it's been broken for long enough now that we can't fix it," but I wanted to make sure I understand why so I don't keep nagging people about this 🙂 


> I will note that 422 **shouldn't** be the return you get when you run out of rate limits, that should be `429` or `403`. `422` is used as a validation error as documented [here](https://docs.github.com/en/rest/using-the-rest-api/troubleshooting-the-rest-api?apiVersion=2022-11-28#validation-failed). If you are getting 422 on rate limit usage please let me know so we can investigate.

Thanks for clarifying this! I believe Sam just meant any kind of error above, but I'll let you know if I experience this returning 422s instead of 429s/403s on rate limiting.

One thing I noticed: the docs you've linked indicate that the response body *will* include an `errors` field on 422 responses, but it doesn't in this case:

```
% gh api /repos/Swatinem/rust-cache/commits/refs/tags/doesnotexist
{
  "message": "No commit found for SHA: refs/tags/doesnotexist",
  "documentation_url": "https://docs.github.com/rest/commits/commits#get-a-commit",
  "status": "422"
}
gh: No commit found for SHA: refs/tags/doesnotexist (HTTP 422)
```

In particular, I think `errors.code == "missing"` would have made this easy enough to treat as semantically equivalent to a 404, but that's not present.

Are the docs wrong here about `errors` being mandatory on 422 responses, or is this a separate errata in this endpoint?



> Finally on a 422 response you can inspect the message returned to validate that it states the ref doesn't exist.

Thanks, this makes sense. The reason I was hesitant to do this is because the message varies with the request (it includes the ref name), so it felt pretty hacky to do a string prefix test on it. I was also concerned that GHES/GHEC deployments might have different i18n settings that change the structure of the human-readable error here. Can you confirm that it's safe to rely on the prefix of this error messaging never changing from `No commit found for SHA`, at least for this version of the API?

logan-porelle: > Out of curiosity, is there a reason the regression itself isn't considered a breaking change? I recall this endpoint working as expected years ago (probably close to a decade ago...). It would make perfect sense to me if the answer was "because it's been broken for long enough now that we can't fix it," but I wanted to make sure I understand why so I don't keep nagging people about this 🙂

Swapping to 422 was a conscious decision that was made ~7 years ago. At that time GitHub had two different queries to get the same ref info, one had 422 as a response and one had 404. The decision was made to go with 422 moving forward. Since it has been 7 years as 422 the risk of changing it is larger for current integrators relying on 422.

> gh: No commit found for SHA: refs/tags/doesnotexist (HTTP 422)
In particular, I think errors.code == "missing" would have made this easy enough to treat as semantically equivalent to a 404, but that's not present.

Let me chat with the team and get back to you on that. Thats definitely a valid point for usability

> Thanks, this makes sense. The reason I was hesitant to do this is because the message varies with the request (it includes the ref name), so it felt pretty hacky to do a string prefix test on it. I was also concerned that GHES/GHEC deployments might have different i18n settings that change the structure of the human-readable error here. Can you confirm that it's safe to rely on the prefix of this error messaging never changing from No commit found for SHA, at least for this version of the API?

I expect this will be the same across GHES/GHEC but I'll run some tests to confirm and will respond here.

Thanks for the quick replies as well

woodruffw: > Swapping to 422 was a conscious decision that was made ~7 years ago. At that time GitHub had two different queries to get the same ref info, one had 422 as a response and one had 404. The decision was made to go with 422 moving forward. Since it has been 7 years as 422 the risk of changing it is larger for current integrators relying on 422.

Thanks for the explanation, that makes sense! 

FWIW, the current docs for [get a commit](https://docs.github.com/en/rest/commits/commits?apiVersion=2022-11-28#get-a-commit) only mention 404 for the "not found" case, not 422. I think it might be helpful if the docs were to mention that 422 is also (or solely) used for the "not found" case 🙂 

> I expect this will be the same across GHES/GHEC but I'll run some tests to confirm and will respond here.

Thank you!

woodruffw:
@kusari-inspector feedback It would be really nice to be able to suppress these per-commit comments: they add a lot of visual noise to PRs and they don't contain a lot of useful information versus the top-level inspector comment (which they just link to).

CC @funnelfiasco as well ^ 🙂

crates/zizmor/src/github_api.rs
logan-porelle: Hiya @woodruffw. 

Thanks for getting the pull request through! I still want to give you some answers on your questions so I'll continue digging. 

Two I have right now are:
> Does this work the same on GHES as GHEC: 

**Yes**. Here is a sample output from my test.
```
{
  "message": "No commit found for SHA: 2342",
  "documentation_url": "https://docs.github.com/enterprise-server@3.16/rest/commits/commits#get-a-commit",
  "status": "422"
```

> Why is 404 in the not found rest api docs:

I am still getting confirmation from the team on how this comes into play but 404 is a response you **can** get. It will occur if the repository doesn't exist. for example `gh api /repos/ThisRepoDoesntExist/ReallyItDoesnt/commits/refs/tags/v2.7.8` will return a 404.

I can see the value in returning a different error if a repository doesn't exist vs a specific commit though at minimum I think we should update the docs to clarify each. I still want to understand the error vs message questions and I continue to chase that.

One final thought since this Pull Request is merged, is there a better place to communicate with you or would you prefer to keep it here?

woodruffw: > Yes. Here is a sample output from my test.

Thanks a ton for confirming that!

>  but 404 is a response you can get. It will occur if the repository doesn't exist. for example gh api /repos/ThisRepoDoesntExist/ReallyItDoesnt/commits/refs/tags/v2.7.8 will return a 404.

Oh, that makes perfect sense, thanks for explaining! Yeah, I think a clarification to the doc would help here, but I can totally see why it's useful/sensible to have 404s for the "root" object and 422s for things under it. I'll update my note in the code 🙂

> is there a better place to communicate with you or would you prefer to keep it here?

Here is completely fine by me! If you find it annoying to follow a comment thread on a closed PR though, we can track with a separate discussion at the org level.

PR #973, opened 2025-06-25
refactor: more fix API cleanup

woodruffw:
Some more fix API cleanup. In particular this makes `yamlpath::Document` the top level input/output type for patch application, since we round-trip through it anyways and doing so gives us more construction guarantees (since `yamlpath::Document` will fail on invalid YAML).

### DIFF
--- a/crates/yamlpath/src/lib.rs
+++ b/crates/yamlpath/src/lib.rs
@@ -272,6 +272,7 @@ enum QueryMode {
 }
 
 /// Represents a queryable YAML document.
+#[derive(Clone)]
 pub struct Document {
     source: String,
     tree: Tree,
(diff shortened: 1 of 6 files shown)


crates/yamlpath/src/lib.rs
```diff
@@ -272,6 +272,7 @@ enum QueryMode {
 }
 
 /// Represents a queryable YAML document.
+#[derive(Clone)]
```
woodruffw, line 275: NB: I don't love that this is now `Clone`; the only reason I did this is to make the "loop while creating new documents" pattern a bit simpler. But I'm going to try and remove it.

woodruffw: It turns out this is annoying to remove -- I could have the fix layer take ownership of the `InputRegistry`, but doing so breaks a bunch of other borrows that aren't worth refactoring for this.

PR #999, opened 2025-07-01
refactor: simplify MergeInto patch op

woodruffw:
This simplifies the `MergeInto` patch op a bit by reducing its scope and eliminating the need for any special-case top-level handling. It leaves the `foo:` i.e. empty mapping body case unimplemented, although now it'll fail with a more contextually appropriate error.

I also went ahead and removed two tests that appear to be duplicates of other tests, despite their names:

* `test_merge_into_with_shell_key`
* `test_merge_into_key_reuse_with_different_value_types`


woodruffw:
I thought through it a bit, and I think the easiest way to handle the `foo:` i.e. empty mapping case it to decompose it into a `(Remove, Add)` pair of operations. However to do that we'll need some other functional fixes to the `Remove` op, so I'm not doing that with this PR.

PR #1010, opened 2025-07-03
fix: treat empty config path as unset

iainlane:
We had a workflow start failing when we updated Zizmor to a version which contained `ZIZMOR_CONFIG` support. The reason is that we have some logic (in a CI job) which supplies a config file dynamically using `--config`, and this logic uses the same environment variable name.

The problem was that the logic was setting `ZIZMOR_CONFIG` to an empty string when we didn't have a config file to supply. This was previously fine, but since 1e123cfbe5ad7534cf7c099fd21e74293880876b Zizmor itself now looks at this variable. An empty string is treated as valid, so we try to load a config file from the empty path, which fails.

```console
$ ZIZMOR_CONFIG= zizmor .
fatal: no audit was performed
error: failed to load config: No such file or directory (os error 2)
 = note: check your configuration file for errors
 = note: see: https://docs.zizmor.sh/configuration/
```

This can be a bit obscure to track down and fix, so what we do here is treat the empty string the same as if the variable (or flag) was unset, so we don't try to load a config file in this scenario.

### DIFF
--- a/crates/zizmor/src/config.rs
+++ b/crates/zizmor/src/config.rs
@@ -87,7 +87,12 @@ impl Config {
             return Ok(Self::default());
         }
 
-        let config = match &app.config {
+        let app_config = app
+            .config
+            .as_ref()
+            .filter(|config| !config.as_str().is_empty());
+
+        let config = match app_config {
             Some(path) => serde_yaml::from_str(&fs::read_to_string(path)?)?,
             None => {
                 // If the user didn't pass a config path explicitly with
--- a/crates/zizmor/tests/integration/common.rs
+++ b/crates/zizmor/tests/integration/common.rs
@@ -104,6 +104,10 @@ impl Zizmor {
         self
     }
 
+    pub fn has_config_set_from_env(&self) -> bool {
+        self.cmd.get_envs().any(|(key, _)| key == "ZIZMOR_CONFIG")
+    }
+
     pub fn run(mut self) -> Result<String> {
         if self.offline {
             self.cmd.arg("--offline");
@@ -115,7 +119,7 @@ impl Zizmor {
 
         if let Some(config) = self.config {
             self.cmd.arg("--config").arg(config);
-        } else {
+        } else if !self.has_config_set_from_env() {
             self.cmd.arg("--no-config");
         }
 
--- a/crates/zizmor/tests/integration/e2e.rs
+++ b/crates/zizmor/tests/integration/e2e.rs
@@ -212,6 +212,26 @@ fn invalid_config_file() -> Result<()> {
 }
 
 #[test]
+fn empty_config_env_var() -> Result<()> {
+    zizmor()
+        .setenv("ZIZMOR_CONFIG", "")
+        .input(input_under_test("e2e-menagerie"))
+        .run()?;
+
+    Ok(())
+}
+
+#[test]
+fn empty_config_arg() -> Result<()> {
+    zizmor()
+        .config("")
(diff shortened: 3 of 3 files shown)


woodruffw:
Hey @iainlane, thanks for opening a PR.

I'll be able to give this a full review after the weekend, but at a high level: I'm a bit hesitant to special-case the handling of `ZIZMOR_CONFIG`, since it otherwise behaves consistently with the other clap-influenced environment variables. I definitely think the current behavior isn't ideal, but I prefer consistency over a point-wise fix here.

Out of curiosity: are you strongly bound to using `ZIZMOR_CONFIG`? If you need this kind of "empty means don't pass" behavior then it might make sense to have your own `OUR_ZIZMOR_CONFIG` variable that you use to conditionally set `--config`, as was the norm in older versions. OTOH I can see that being tedious/frustrating given that the built-in environment variable *almost* works, so I'm curious what you think there 🙂

TL;DR: I'm not opposed to changing the behavior here, but I think I'd like it to be consistent across all current environment variables rather than a special carve out for just `ZIZMOR_CONFIG`. I'm not sure if clap makes that easy to do however, or whether it would need to be an additional generic layer on top of the current derived argument parser.

woodruffw:
(Noting: this may be doable with clap's `TypedValueParser`. There's an existing `NonEmptyStringValueParser`, but that might not do what we want here...)

iainlane:
Heya, thanks for the quick reply.

> I'll be able to give this a full review after the weekend, but at a high level: I'm a bit hesitant to special-case the handling of `ZIZMOR_CONFIG`, since it otherwise behaves consistently with the other clap-influenced environment variables. I definitely think the current behavior isn't ideal, but I prefer consistency over a point-wise fix here.

This makes sense.

> Out of curiosity: are you strongly bound to using `ZIZMOR_CONFIG`? If you need this kind of "empty means don't pass" behavior then it might make sense to have your own `OUR_ZIZMOR_CONFIG` variable that you use to conditionally set `--config`, as was the norm in older versions. OTOH I can see that being tedious/frustrating given that the built-in environment variable _almost_ works, so I'm curious what you think there 🙂

Yes! That's what I've done, actually. I just found the behaviour surprising and suspected it might not be intentional, so I felt that I'd send in a PR and see 😁

> NonEmptyStringValueParser

I tried this initially, but it _fails_ parsing if it's empty. However, to achieve the goal of consistency, one thing we could do (hopefully) is implement a parser which is almost this, but which returns `None` on empty instead of failing.

woodruffw:
No problem!

And yeah, this behavior is in the unfortunate category of "not intentional but also known because of clap" 😅 -- I'm not a huge fan of it, but it's clap's default for reasons that strike me as not entirely unreasonable in other contexts.

> However, to achieve the goal of consistency, one thing we could do (hopefully) is implement a parser which is almost this, but which returns None on empty instead of failing.

This makes sense to me!

woodruffw:
I looked into the value parser approach a bit, and it's nontrivial: we'd ideally have a custom `TypedValueParser` that produces `Option<T>`, but clap's architectural assumption is that value parsers operate _after_ the `Option<...>` state has already been stripped away.

Given that, I'm tempted to do this in two parts: I'm going to move the current paths to `NonEmptyStringValueParser`, and then look at a more general fix. `NonEmptyStringValueParser` doesn't do exactly what we want here (per above), but it at least produces a more intelligible error message than the one that currently happens when we attempt to use empty strings as if they're meaningful values.

PR #1018, opened 2025-07-09
Use older Ubuntu runners for Debian compatibility

### LINKED ISSUE #1017: [BUG]:  not working on debian stable
### Pre-submission checks

### zizmor version

any

### Expected behavior

❯ zizmor .github/workflows/release_doc.yml doing some works (I don't know yet the product, wanted to try it)

### Actual behavior

❯ zizmor .github/workflows/release_doc.yml
zizmor: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.39' not found (required by zizmor)

### Reproduction steps

mise install zizmor@latest
mise ues -g zizmor@latest
zizmor .github/workflows/release_doc.yml
zizmor: /lib/x86_64-linux-gnu/libc.so.6: version `GLIBC_2.39' not found (required by zizmor)

### Logs

```plain text

```

### Additional context

I think you use a too young version of linux runner to build your versions

funnelfiasco:
Debian stable's GCC is too far behind ubuntu-latest, which means the
binaries won't work on Debian.

Fixes #1017

Signed-off-by: Ben Cotton <ben@kusari.dev>

### DIFF
--- a/.github/workflows/release-binaries.yml
+++ b/.github/workflows/release-binaries.yml
@@ -14,9 +14,9 @@ jobs:
     strategy:
       matrix:
         include:
-          - runner: ubuntu-latest
+          - runner: ubuntu-22.04
             target: x86_64-unknown-linux-gnu
-          - runner: ubuntu-24.04-arm
+          - runner: ubuntu-22.04-arm
             target: aarch64-unknown-linux-gnu
           - runner: macos-latest
             target: x86_64-apple-darwin


funnelfiasco:
I'm not sure I'm in love with this as the solution to #1017, but it would seem to solve the problem for now. (AFAIK, nothing in zizmor requires newer glibc)

Pros:

* Increases compatibility for Linux platforms. In addition to Debian stable, I wouldn't be surprised if the glibc issues affect Ubuntu 22.04, RHEL & derivates, etc.

Cons:

* Requires manual update of workflow matrix on occasion, which
    * is likely to get missed
    * could still lead to an abrupt end to support for stable, enterprisey distros

Overall, I think this is best approach available, apart from maintaining runners or refactoring the build process to use containers.

woodruffw:
Thanks @funnelfiasco! This makes sense to me, I'll do a full review tonight.

(Going forwards I'll probably be more persnickety about anything that "fixes" these release binaries, since I don't especially want to keep them around. But for now there are `mise` users who expect them, so this seems good.)

woodruffw (approved):
Thanks @funnelfiasco!

PR #1038, opened 2025-07-24
feat: CLI benchmarking harness

woodruffw:
WIP.

This adds `benchmark.py`, which is essentially just a bit of instrumentation around `hyperfine` and benchmark planning/execution (including input setup).

It's a little bit overkill, but the thinking behind it is that it'll give me a decent foundation to add more benchmarks in the future, plus gives me structured output (JSON files in `bench/results`). For the latter, I'm going to evaluate [bencher.dev](https://bencher.dev/).

### DIFF
--- a/.github/workflows/benchmark-base.yml
+++ b/.github/workflows/benchmark-base.yml
@@ -1,0 +1,62 @@
+# benchmark-base.yml: submit benchmarks to Bencher.
+#
+# This workflow provides baseline results, via the main branch.
+
+name: Benchmark baseline
+
+on:
+  push:
+    branches: [main]
+
+permissions: {}
+
+jobs:
+  benchmark_base_branch:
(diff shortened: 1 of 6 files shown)


woodruffw:
Saving some things to think about later:

* The current benchmarks are probably outliers, in the sense that they're huge repos. Consequently, the benchmark time is _probably_ dominated by `zizmor`'s collection phase, not by auditing itself. It might make sense to start with normal-ish sized repos.
* Writing each benchmark out manually is going to be tedious and plan-heavy. It might be nice to be able to parametrize each benchmark so that the archive/extract parts of the plan can be reused.

woodruffw:
I can't tell if this is wrong or not, probably because there's no `main` baseline. So I'm going to merge and hope for the best.

PR #1041, opened 2025-07-25
refactor: remove RwLock from github-env audit

woodruffw:
Now that I have benchmarks, maybe they'll reveal
whether this was a good idea or not.


woodruffw:
...inconclusive. Good enough for me!

PR #1045, opened 2025-07-29
Add Fix for insecure-commands audit rule

mostafa:
xref #876

### DIFF
--- a/docs/release-notes.md
+++ b/docs/release-notes.md
@@ -20,6 +20,8 @@ of `zizmor`.
 * The [use-trusted-publishing] audit now detects many more patterns, including
   `cargo publish` and other `#!yaml run:` blocks that make use of publishing
   commands directly (#1042)
+* The [insecure-commands] audit now supports auto-fixes for many findings
+  (#1045)
 
 ### Bug Fixes 🐛
 


woodruffw:
Thanks @mostafa! I'm reviewing now.

crates/zizmor/src/audit/insecure_commands.rs
```diff
@@ -161,3 +182,273 @@ impl Audit for InsecureCommands {
...
+
+                // Check that ACTIONS_ALLOW_UNSECURE_COMMANDS is removed
+                assert!(
+                    !fixed_document
+                        .source()
+                        .contains("ACTIONS_ALLOW_UNSECURE_COMMANDS")
+                );
+
+                // Check that other environment variables are preserved
+                assert!(fixed_document.source().contains("OTHER_VAR: keep-me"));
```
woodruffw: Could you make this an inline snapshot instead?

crates/zizmor/src/audit/insecure_commands.rs
```diff
@@ -22,6 +25,21 @@ audit_meta!(
 );
 
 impl InsecureCommands {
+    /// Creates a fix that removes the ACTIONS_ALLOW_UNSECURE_COMMANDS environment variable.
+    fn create_remove_fix<'doc>(&self, location: SymbolicLocation<'doc>) -> Fix<'doc> {
```
woodruffw: Nit: `create_remove_fix` is kind of a brain-bender with two verbs, maybe just:

```suggestion
    fn create_fix<'doc>(&self, location: SymbolicLocation<'doc>) -> Fix<'doc> {
```

woodruffw: (Matching the other fix helpers where we don't say which internal ops the fix does as well.)

woodruffw (commented):
This looks pretty good to me, thanks @mostafa! Just some small nits.

One additional request: could we add a negative testcase for the "maybe present" scenario, i.e. to ensure we don't suggest a fix when we aren't sure whether `ACTIONS_ALLOW_UNSECURE_COMMANDS` is present?

mostafa:
@woodruffw Thanks for the review! I'll have a look tonight.

crates/zizmor/src/audit/insecure_commands.rs
mostafa: Fixed in https://github.com/zizmorcore/zizmor/pull/1045/commits/f3b276a6ba8bf463a704de9eaafaaa70a8251cab.

crates/zizmor/src/audit/insecure_commands.rs
mostafa: Fixed in https://github.com/zizmorcore/zizmor/pull/1045/commits/c12f95061517768c4778fc7bab44a5820e767f7c.

mostafa:
@woodruffw Addressed your comments.

woodruffw (approved):
Thanks @mostafa!

PR #1064, opened 2025-08-07
Detect `shell` before applying template expression fixes

### LINKED ISSUE #1049: [BUG]: Auto-fixes for `template-injection` are incorrect on non-sh shells
### Pre-submission checks

### zizmor version

1.11.0

### Expected behavior

Observed this today when doing some auto-fixes -- we currently insert `${GENERATED_VARIABLE}` everywhere, but that's only correct for sh-style shells (like bash).

Some other syntaxes we should probably support correctly:

* Windows `cmd`: `%GENERATED_VARIABLE%`
* Windows PowerShell/pwsh: `$env:GENERATED_VARIABLE`

...those are probably sufficient for now. If we detect a shell other than those three, we should probably just refrain from suggesting an auto-fix.

CC @mostafa -- let me know if you're interested in this one, otherwise I'll probably do it in the coming days 🙂 

### Actual behavior

Our current actual behavior is that we insert `${GENERATED_VARIABLE}` into any `run:` block, regardless of whether it's valid syntax in that block's shell or not.

### Reproduction steps

This:

```yaml
      - name: Build
        if: ${{ ! matrix.dry-run }}
        shell: cmd
        run: |
          call "C:\Program Files\Microsoft Visual Studio\2022\Enterprise\VC\Auxiliary\Build\${{ matrix.vcvars }}"
          py.exe -3.12 build-windows.py --python cpython-${{ matrix.python }} --sh c:\cygwin\bin\sh.exe --options ${{ matrix.build_options }}
```

Patches incorrectly as:

```yaml
      - name: Build
        if: ${{ ! matrix.dry-run }}
        shell: cmd
        run: |
          call "C:\Program Files\Microsoft Visual Studio\2022\Enterprise\VC\Auxiliary\Build\${MATRIX_VCVARS}"
          py.exe -3.12 build-windows.py --python cpython-${MATRIX_PYTHON} --sh c:\cygwin\bin\sh.exe --options ${MATRIX_BUILD_OPTIONS}
        env:
          MATRIX_VCVARS: ${{ matrix.vcvars }}
          MATRIX_PYTHON: ${{ matrix.python }}
          MATRIX_BUILD_OPTIONS: ${{ matrix.build_options }}
```

Ref: <https://github.com/astral-sh/python-build-standalone/blob/a376f32981a4e690963fe29a6ae31203afb440ae/.github/workflows/windows.yml#L142-L147>

### Logs

```plain text
N/A
```

### Additional context

_No response_

mostafa:
This PR fixes #1049 by detecting all the supported shells except Python and applying the correct environment variable syntax for fixing the finding(s), according to [GitHub workflow syntax](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#jobsjob_iddefaultsrunshell).

### DIFF
--- a/crates/zizmor/src/audit/template_injection.rs
+++ b/crates/zizmor/src/audit/template_injection.rs
@@ -180,7 +180,7 @@ impl TemplateInjection {
         }
 
         // Get the effective shell, which includes defaults and fallbacks
-        let shell = step.effective_shell()?;
+        let shell = step.shell()?;
 
         match shell {
             // sh-style shells (bash, sh, zsh, etc.)
--- a/crates/zizmor/src/audit/template_injection.rs
+++ b/crates/zizmor/src/audit/template_injection.rs
@@ -189,8 +189,7 @@ impl TemplateInjection {
             "cmd" => Some(format!("%{env_var}%")),
             // PowerShell (both Windows PowerShell and PowerShell Core)
             "pwsh" | "powershell" => Some(format!("$env:{env_var}")),
-            // Python shell (uses os.environ)
-            "python" => Some(format!("{{os.environ.get('{env_var}')}}")),
+
             // For unknown shells, don't provide a fix to avoid incorrect syntax
             _ => None,
         }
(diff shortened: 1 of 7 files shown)


crates/zizmor/src/audit/template_injection.rs
```diff
@@ -170,6 +170,32 @@ impl TemplateInjection {
...
+
+        match shell {
+            // sh-style shells (bash, sh, zsh, etc.)
+            "bash" | "sh" | "zsh" => Some(format!("${{{env_var}}}")),
+            // Windows Command Prompt
+            "cmd" => Some(format!("%{env_var}%")),
+            // PowerShell (both Windows PowerShell and PowerShell Core)
+            "pwsh" | "powershell" => Some(format!("$env:{env_var}")),
+            // Python shell (uses os.environ)
+            "python" => Some(format!("{{os.environ.get('{env_var}')}}")),
```
woodruffw: Let's drop this one -- there's probably too much context sensitivity with Python to insert a meaningful patch 🙂

mostafa: Fixed in https://github.com/zizmorcore/zizmor/pull/1064/commits/a13a0f847b0584f1d4b22889f0946acd9b3be38b.

crates/zizmor/src/models.rs
```diff
@@ -54,6 +54,11 @@ pub(crate) trait StepCommon<'doc>: Locatable<'doc> + HasInputs {
 
     /// Returns the document which contains this step.
     fn document(&self) -> &'doc yamlpath::Document;
+
+    /// Returns the effective shell for this step, if it can be determined.
+    /// This includes the step's explicit shell, job defaults, workflow defaults,
+    /// and runner defaults. Returns `None` if the shell cannot be statically determined.
+    fn effective_shell(&self) -> Option<&str>;
```
woodruffw: Nitpick: Could we name this just `shell()`, for consistency with `Step::shell`?

mostafa: Fixed in https://github.com/zizmorcore/zizmor/pull/1064/commits/034e16835362095a87f5b57c2afa74fc0c344453.

woodruffw (approved):
Thanks @mostafa, this looks great to me!

Issue #1023, opened 2025-07-13
Feature: Require or recommend the `timeout-minutes` property for all jobs


johnbillion:
### Pre-submission checks

- [x] I am **not** reporting a bug (crash, false positive/negative, etc). These must be filed via the bug report template.
- [x] I have looked through the open issues for a duplicate request.

### What's the problem this feature will solve?

The `timeout-minutes` property can be used to prevent run-away jobs from consuming all your allowed GitHub Actions runner minutes. [The default value is 360 minutes](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idtimeout-minutes) (6 hours).

Examples include scripts that accidentally hang or go into a loop, package managers which slow to a crawl when the registry is responding slowly, compilers that get stuck, or really anything else that can unexpectedly slow but not be constrained by its own timeout.

I use this property on all jobs in all my workflows.

### Describe the solution you'd like

I would like Zizmor to enforce or recommend the use of a `timeout-minutes` property on all jobs. This property is not strictly related to security, but I think it's such a useful configuration that it could be part of the "pedantic" or "auditor" persona.

At a stretch, you could suggest that a vulnerability in a workflow could allow an attacker to consume all your GitHub Actions minutes, and this would protect against that, but honestly it's more likely that `npm install` hangs and eats all your allowance.

### Additional context

Jobs which call a reusable workflow (with `uses`) do not support `timeout-minutes`. The property needs to be defined in the jobs within the reusable workflow that it calls.

woodruffw:
Thanks @johnbillion! Yeah, I agree this would be good as a "pedantic" finding. 

(I agree with your comments about security relevance -- this is funny in that it cuts both ways, i.e. an attacker could waste resources without `timeout-minutes`, *or* they could induce an availability risk with `timeout-minutes`. Which one is more important probably ends up being context- and trigger-sensitive.)

EmmanuelBerkowicz:
Hey @woodruffw , @johnbillion,
I'd like to work on this issue. 
I'm relatively new to contributing to Python projects but have experience with CI/CD workflows and GitHub Actions. 
Could you point me to any existing checks that would be good examples to follow for implementing this feature?

woodruffw:
Hi @EmmanuelBerkowicz, thanks for offering your time.

This is a Rust project, not a Python project, just so you know 😉 

> Could you point me to any existing checks that would be good examples to follow for implementing this feature?

[...]

PR #1067, opened 2025-08-10
Feature: Require or recommend the timeout-minutes property for all jobs #1023

EmmanuelBerkowicz:
feat: add basic timeout-minutes audit for GitHub Actions jobs

Summary: I have made a start on the open issue, requesting feedback on progress as this is my first rust project.

Implements a new audit rule that detects missing timeout-minutes property on GitHub Actions jobs to prevent runaway jobs from consuming runner minutes.

What's implemented:
- Basic audit structure following existing patterns (unpinned-images)
- Detection of missing timeout-minutes on normal jobs
- Proper finding generation with medium severity
- Integration with pedantic persona
- Registration in audit registry

What's still needed (for future iterations):
- Handle reusable workflows (jobs with 'uses' don't support timeout-minutes directly)
- Check step-level timeout-minutes and transitive coverage by job timeouts
- Add comprehensive test coverage
- Consider different severity levels or personas

Addresses #1023

### DIFF
--- a/crates/zizmor/src/audit/timeout_minutes.rs
+++ b/crates/zizmor/src/audit/timeout_minutes.rs
@@ -1,0 +1,62 @@
+use anyhow::Result;
+
+use crate::{
+    finding::{
+        Confidence, Finding, Persona, Severity,
+        location::{Locatable as _, SymbolicLocation},
+    },
+    models::workflow::JobExt as _,
+    state::AuditState,
+};
+
+use super::{Audit, AuditLoadError, audit_meta};
+
+pub(crate) struct TimeoutMinutes;
+
+impl TimeoutMinutes {
+    fn build_finding<'doc>(
+        &self,
+        location: SymbolicLocation<'doc>,
+        annotation: &str,
+        job: &super::NormalJob<'doc>,
+    ) -> Result<Finding<'doc>> {
+        let mut annotated_location = location;
+        annotated_location = annotated_location.annotated(annotation);
+        Self::finding()
+            .severity(Severity::Medium)
+            .confidence(Confidence::High)
+            .add_location(annotated_location)
+            .persona(Persona::Pedantic)
+            .build(job.parent())
+    }
+}
+
+audit_meta!(
+    TimeoutMinutes,
+    "timeout-minutes",
+    "missing timeout-minutes on jobs"
+);
+
+impl Audit for TimeoutMinutes {
+    fn new(_state: &AuditState<'_>) -> Result<Self, AuditLoadError> {
+        Ok(Self)
+    }
+
+    fn audit_normal_job<'doc>(
+        &self,
+        job: &super::NormalJob<'doc>,
+    ) -> anyhow::Result<Vec<Finding<'doc>>> {
+        let mut findings = vec![];
+
+        // Check if timeout-minutes is missing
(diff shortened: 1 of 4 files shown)


woodruffw:
Thanks for tackling this @EmmanuelBerkowicz!

Please give me a ping when this is ready for an initial review (that can be now, but I didn't want to assume).

EmmanuelBerkowicz:
@woodruffw ,
No worries mate! It was a bit more complicated than I thought which is why it took me so long.

I've given it my best shot, but like I said in the PR, it's not the entire issue scope just yet.

If you could give it a once over and let me know if my approach is in the right direction that would be a huge help! If I am reading the scope correctly, there is still a fair bit to go and it will be a lot easier for me to do if I know that I'm heading in the right direction :)

crates/zizmor/src/audit/timeout_minutes.rs
```diff
@@ -0,0 +1,62 @@
...
+        &self,
+        job: &super::NormalJob<'doc>,
+    ) -> anyhow::Result<Vec<Finding<'doc>>> {
+        let mut findings = vec![];
+
+        // Check if timeout-minutes is missing
+        if job.timeout_minutes.is_none() {
+            findings.push(self.build_finding(
+                job.location().primary(),
+                "job is missing timeout-minutes",
```
woodruffw, line 55:
```suggestion
                "missing timeout-minutes",
```

crates/zizmor/src/audit/timeout_minutes.rs
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    fn audit_normal_job<'doc>(
+        &self,
+        job: &super::NormalJob<'doc>,
+    ) -> anyhow::Result<Vec<Finding<'doc>>> {
+        let mut findings = vec![];
+
+        // Check if timeout-minutes is missing
+        if job.timeout_minutes.is_none() {
```
woodruffw, line 52: This approach looks right to me! Let's do the same for each step in the job as well :slightly_smiling_face:

test-workflow.yml
woodruffw, line 1: This should probably be moved under `crates/zizmor/tests/integration/test-data` and named `timeout-minutes.yml` or similar, once you're ready :slightly_smiling_face:

crates/zizmor/src/registry.rs
```diff
@@ -352,6 +352,7 @@ impl AuditRegistry {
         register_audit!(audit::self_hosted_runner::SelfHostedRunner);
         register_audit!(audit::known_vulnerable_actions::KnownVulnerableActions);
         register_audit!(audit::unpinned_uses::UnpinnedUses);
+        register_audit!(audit::timeout_minutes::TimeoutMinutes);
```
woodruffw, line 355: Move this to the end, please -- right now these are declared in order of introduction, not alphabetically.

woodruffw (commented):
Thanks @EmmanuelBerkowicz!

Your approach here looks good to me; I think once this includes checks for step-level timeouts and some tests then it'll be good to go.

(Let me know if you have any questions about how to run the tests, including updating the snapshots. Most of it is documented here: <https://docs.zizmor.sh/development/#testing>)

woodruffw:
Triage: #2166

PR #1088, opened 2025-08-16
Add Fixes for obfuscation audit rule

mostafa:
This PR introduces an expression evaluator inspired by the GitHub Actions [expressions docs](https://docs.github.com/en/actions/reference/workflows-and-actions/expressions) and its [reference implementation](https://github.com/actions/languageservices/tree/main/expressions) in Typescript. It builds on top of the `github-actions-expressions` crate.

Additionally, it adds auto-fixes for the obfuscation audit rule covering both `uses` statements and expressions.

xref #876

### DIFF
--- a/docs/release-notes.md
+++ b/docs/release-notes.md
@@ -45,6 +45,8 @@ of `zizmor`.
   See [`rules.<id>.disable`](./configuration.md#rulesiddisable)
   for details (#1132)
 
+* The [obfuscation] audit now supports auto-fixes for many findings (#1088)
+
 ### Bug Fixes 🐛
 
 * `zizmor` now correctly honors `--strict-collection` when collecting from


woodruffw:
Thanks @mostafa! I'll review this one today.

crates/github-actions-expressions/src/lib.rs
```diff
@@ -674,6 +674,346 @@ impl From<bool> for Expr<'_> {
     }
 }
 
+/// The result of evaluating a GitHub Actions expression.
+///
+/// This type represents the possible values that can result from evaluating
+/// constant-reducible GitHub Actions expressions.
+#[derive(Debug, Clone, PartialEq)]
+pub enum EvaluationResult {
```
woodruffw: Nitpick: Can we rename this to `Evaluation`? I think that's sufficiently clear :slightly_smiling_face:

crates/github-actions-expressions/src/lib.rs
```diff
@@ -674,6 +674,346 @@ impl From<bool> for Expr<'_> {
...
+                if n.fract() == 0.0 {
+                    format!("{}", *n as i64)
+                } else {
+                    n.to_string()
+                }
+            }
+            EvaluationResult::Boolean(b) => b.to_string(),
+            EvaluationResult::Null => String::new(),
+        }
+    }
```
woodruffw: Could we move this into the `Display` impl directly and remove this API? I think that would be a bit more idiomatic (and all users could use `to_string()` instead).

crates/github-actions-expressions/src/lib.rs
```diff
@@ -674,6 +674,346 @@ impl From<bool> for Expr<'_> {
...
+
+    /// Convert to a boolean following GitHub Actions truthiness rules.
+    ///
+    /// GitHub Actions truthiness:
+    /// - false and null are falsy
+    /// - Numbers: 0 is falsy, everything else is truthy
+    /// - Strings: empty string is falsy, everything else is truthy
+    pub fn to_boolean(&self) -> bool {
```
woodruffw: Nitpick: rename to `as_boolean` since `to_*` usually implies ownership transfer in Rust:

```suggestion
    pub fn as_boolean(&self) -> bool {
```

crates/github-actions-expressions/src/lib.rs
```diff
@@ -674,6 +674,346 @@ impl From<bool> for Expr<'_> {
...
+                    a.to_github_string().parse::<f64>(),
+                    b.to_github_string().parse::<f64>(),
+                ) {
+                    a_num.partial_cmp(&b_num)
+                } else {
+                    Some(a.to_github_string().cmp(&b.to_github_string()))
+                }
+            }
+        }
+    }
```
woodruffw: Nitpick: Could we make this a `PartialOrd` impl instead?

crates/github-actions-expressions/src/lib.rs
mostafa: Fixed in https://github.com/zizmorcore/zizmor/pull/1088/commits/179c19bfa5a9d451c44a895da2469c9c5616f503.

crates/github-actions-expressions/src/lib.rs
mostafa: Fixed in https://github.com/zizmorcore/zizmor/pull/1088/commits/d1816442a7f1104b7bad7a3ef72a8ad714361195.

crates/github-actions-expressions/src/lib.rs
mostafa: Fixed in https://github.com/zizmorcore/zizmor/pull/1088/commits/f04a001888fed50bcd80b9600806f54d072a47fd.

crates/github-actions-expressions/src/lib.rs
mostafa: Fixed in https://github.com/zizmorcore/zizmor/pull/1088/commits/12cb66d33930f232742a78657dc56837c49f1332.

woodruffw:
Just noting: I haven't forgotten about this or the other audit autofix PRs, I'm just working on #1094 and will deconflict these once that's done :slightly_smiling_face:

mostafa:
@woodruffw Thanks for letting me know. I saw the PR the other day and wondered it might be the reason.

crates/github-actions-expressions/src/lib.rs
```diff
@@ -43,6 +43,156 @@ impl PartialEq<str> for Function<'_> {
...
+        let mut result = format_template;
+
+        // Replace {0}, {1}, {2}, etc. with the corresponding arguments
+        for (i, arg) in args.iter().skip(1).enumerate() {
+            let placeholder = format!("{{{}}}", i);
+            result = result.replace(&placeholder, &arg.to_string());
+        }
+
+        Some(Evaluation::String(result))
+    }
```
woodruffw, line 150: @mostafa FYI, I think this and some of the other consteval function implementations don't 100% match the reference implementation (which I've linked in the comments for each function). 

Some of these will be nontrivial to implement, so we can make a judgement call here 🙂 -- I'd personally be okay with having some of the more annoying ones return `None` for now so that we can get this out as an MVP, and then iterate on their implementations going forwards. But I'm curious what you think.

(The consequence of that will be that we sometimes won't have a full evaluation to offer for a constant expression, so we won't have a fix in those cases. I think that's okay for the MVP.)

crates/github-actions-expressions/src/lib.rs
```diff
@@ -43,6 +43,156 @@ impl PartialEq<str> for Function<'_> {
...
+            s => {
+                // Try to parse as number
+                if let Ok(n) = s.parse::<f64>() {
+                    Some(Evaluation::Number(n))
+                } else {
+                    None
+                }
+            }
+        }
+    }
```
woodruffw: `fromJSON` can also be given an array or dictionary literal, in which case it'll return an `Array` or `Dictionary` type (in the reference implementation).

To handle that, I think we probably want to parse the input string with a full JSON parser, and then filter out the non-primitive cases as non-const for now (technically they can be const, but that'll require use to tweak `Evaluation` to add variants for that).

crates/github-actions-expressions/src/lib.rs
mostafa: How about matching the reference implementation as much as we can, while we are at it? 🤔

woodruffw: Works for me!

mostafa: I'll be able to fix it on Sep 4th onward. If you have too much to release, I guess we can have this in the next.

crates/github-actions-expressions/src/lib.rs
mostafa: Hey! 👋 I refactored all functions to make them match the reference implementation in TS. I also added the missing `join` function.

woodruffw: Awesome, thank you @mostafa! I'll try and do a review tonight.

crates/github-actions-expressions/src/lib.rs
woodruffw, line 1: This file is now way too large 😅 -- it's fine for this PR, but I'm going to do a followup to break it apart a bit.

woodruffw:
Thanks @mostafa!

PR #1089, opened 2025-08-16
Add Fix to unsound-condition audit rule

mostafa:
This PR adds auto-fixes to the unsound-condition audit rule plus tests to verify that it works as expected.

xref #876

### DIFF
--- a/crates/zizmor/src/audit/unsound_condition.rs
+++ b/crates/zizmor/src/audit/unsound_condition.rs
@@ -93,7 +93,9 @@ impl UnsoundCondition {
 
         // Create a patch that replaces the scalar indicator
         Some(Fix {
-            title: format!("replace unsound block scalar style '{old_indicator}' with sound style '{new_indicator}'").into(),
+            title: format!(
+                "replace unsound block scalar style '{old_indicator}' with sound style '{new_indicator}'"
+            ),
             key: loc.key,
             disposition: FixDisposition::Safe,
             patches: vec![Patch {


crates/zizmor/src/audit/unsound_condition.rs
```diff
@@ -58,6 +59,59 @@ impl UnsoundCondition {
...
+                .collect::<Vec<_>>()
+                .join(" ");
+
+            // Remove any existing ${{ }} wrapping and re-wrap properly
+            let unwrapped = if cleaned.starts_with("${{") && cleaned.ends_with("}}") {
+                cleaned[3..cleaned.len() - 2].trim()
+            } else {
+                &cleaned
+            };
+            format!("${{{{ {unwrapped} }}}}")
```
woodruffw: I think this transformation has the potential to be pretty disruptive, in terms of turning a user's carefully formatted expression into a single line. 

Instead of doing this, I think we would ideally do one of two things here (either is fine):

1. Keep the user's newlines, but remove the fence (`${{ }}`) around the expression. This will preserve the user's formatting intent while making the condition sound (since the newlines will be parsed as part of the expression).
2. Keep the expression exactly as it is, but replace the unsound string style (`|` or `>`) with a sound one (`|-` or `>-`). If we go for this route, we should probably replace them on a pairwise basis, i.e. `|` becomes `|-`. 

The main advantage to (2) is that it fully preserves the user's formatting and style choices while fixing the bug. The downside is that it might require some changes to `yamlpatch`, since I'm not sure if we currently expose the string styling state in a useful way here.

mostafa: Sorry for the delay! I fixed it the 2nd way in https://github.com/zizmorcore/zizmor/pull/1089/commits/1307f2575e200ae5930a6a0438567c35e694560e.

woodruffw: No problem at all, thanks! I'll do a review now.

PR #1090, opened 2025-08-16
Add Fix to imposter-commit audit rule

mostafa:
This PR adds auto-fixes to the imposter-commit audit rule with online tests.

### DIFF
--- a/crates/zizmor/src/audit/impostor_commit.rs
+++ b/crates/zizmor/src/audit/impostor_commit.rs
@@ -388,10 +388,6 @@ jobs:
     #[cfg(feature = "gh-token-tests")]
     #[test]
     fn test_no_impostor_with_valid_tag() {
-        use insta::assert_snapshot;
-
-        use crate::models::AsDocument as _;
-
         use super::*;
         use crate::{models::workflow::Workflow, registry::input::InputKey};
 
--- a/crates/zizmor/src/audit/ref_version_mismatch.rs
+++ b/crates/zizmor/src/audit/ref_version_mismatch.rs
@@ -218,6 +218,11 @@ mod tests {
     #[test]
     fn test_fix_version_comment_mismatch() {
         use crate::config::Config;
+        use crate::{
+            models::{AsDocument, workflow::Workflow},
+            registry::input::InputKey,
+        };
+
         let workflow_content = r#"
 name: Test Version Comment Mismatch
 on: push
@@ -281,6 +286,11 @@ jobs:
     #[test]
     fn test_fix_version_comment_different_formats() {
         use crate::config::Config;
+        use crate::{
+            models::{AsDocument, workflow::Workflow},
+            registry::input::InputKey,
+        };
+
         let workflow_content = r#"
 name: Test Different Version Formats
 on: push


mostafa:
@woodruffw Any updates here?

woodruffw:
> @woodruffw Any updates here?

Nope, I need to do another review pass. My plan is to get to it tonight.

crates/zizmor/src/audit/impostor_commit.rs
```diff
@@ -111,6 +121,98 @@ impl ImpostorCommit {
...
+        let mut uses_slug = format!("{}/{}", uses.owner, uses.repo);
+        if let Some(subpath) = &uses.subpath {
+            uses_slug.push_str(&format!("/{subpath}"));
+        }
+        let fixed_uses = format!("{uses_slug}@{latest_tag}");
+
+        Some(Fix {
+            title: format!("pin to latest tag {latest_tag}"),
+            key,
+            disposition: FixDisposition::Safe,
```
woodruffw: I *think* we should characterize this as unsafe, for two reasons:

1. We can't guarantee compatibility between action versions
2. Impostor commits are *usually* a very strong signal of compromise or malicious activity, so we probably want users to explicitly acknowledge an auto-fix (since it isn't the end of the remediation they'll need to do)

crates/zizmor/src/audit/impostor_commit.rs
```diff
@@ -111,6 +121,98 @@ impl ImpostorCommit {
...
+            }
+            Err(e) => {
+                tracing::warn!(
+                    "Failed to get latest tag for {}/{}: {}",
+                    uses.owner,
+                    uses.repo,
+                    e
+                );
+                return None;
+            }
```
woodruffw, line 196: I _think_ this should be a hard error rather than a warning, unless I'm missing something.

crates/zizmor/src/audit/impostor_commit.rs
```diff
@@ -192,15 +300,147 @@ impl Audit for ImpostorCommit {
...
+        // This creates an impostor scenario: valid commit, wrong repository
+        let workflow_content = r#"
+name: Test Impostor Commit Fix
+on: push
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/hello-world-javascript-action@692973e3d937129bcbf40652eb9f2f61becf3332  # This is a commit from actions/checkout, not hello-world
+"#;
```
woodruffw, line 343: Huh, I'm actually kind of surprised this works -- the audit wasn't really meant to detect "transposed" commits like this, but commits that are valid in the fork network of the parent repo. I guess those two scenarios look pretty similar from our API queries, though.

(No changes required, I'm just surprised this works!)

crates/zizmor/src/audit/impostor_commit.rs
```diff
@@ -192,15 +300,147 @@ impl Audit for ImpostorCommit {
...
+            // Apply the fix and snapshot test the result
+            let new_doc = findings[0].fixes[0].apply(input.as_document()).unwrap();
+            assert_snapshot!(new_doc.source(), @r"
+            name: Test Impostor Commit Fix
+            on: push
+            jobs:
+              test:
+                runs-on: ubuntu-latest
+                steps:
+                  - uses: actions/hello-world-javascript-action@v1.1  # This is a commit from actions/checkout, not hello-world
```
woodruffw, line 383: Just noting: I'm guessing this snapshot will eventually go stale, right? I don't expect this repo to change very often, but making sure I understand.

crates/zizmor/src/audit/impostor_commit.rs
```diff
@@ -192,15 +300,147 @@ impl Audit for ImpostorCommit {
...
+        let _workflow = Workflow::from_string(workflow_content.to_string(), key).unwrap();
+        let audit_state = crate::state::AuditState::default();
+
+        match ImpostorCommit::new(&audit_state) {
+            Err(crate::audit::AuditLoadError::Skip(_)) => {
+                // Expected behavior when no GitHub client is available
+            }
+            _ => panic!("Expected audit to skip when no GitHub client is available"),
+        }
+    }
```
woodruffw: I think this can be removed 🙂

crates/zizmor/src/audit/impostor_commit.rs
mostafa: Makes sense!

crates/zizmor/src/audit/impostor_commit.rs
mostafa: Fair enough!

crates/zizmor/src/audit/impostor_commit.rs
mostafa: Technically, yes. Do you want me to remove it?

woodruffw: Nope, I'm good with it! Just wanted to make sure I understood it so I'm not confused if/when it does go stale 🙂 

(If it happens more than once or twice, we'll probably want to make a minimal reproducer repo instead. But for now this is totally fine.)

woodruffw (approved):
Thanks @mostafa!

Thinking out loud: technically this won't produce a fixpoint with `ref-version-mismatch`, since our edit here won't also edit the version comment (if present). I think that's OK for now since this is an "unsafe" fix.

Issue #1055, opened 2025-08-03
Rethinking `zizmor`'s configuration and input collection


woodruffw:
Here's how `zizmor` currently determines its configuration:

* If the user explicitly passes in a `--config`, we use that
* Otherwise, we look at `$CWD/.github/zizmor.yml` or `$CWD/zizmor.yml` (in that order) to auto-discover a config file

This approach is simple, but it has friction/inertia implications for `zizmor`'s internals:

* There's a fundamental impedance mismatch between `zizmor` taking a single config file and also multiple inputs, each of which might be a full repo (or remote slug reference): a user who does `zizmor foo/ bar/ baz/` probably expects each repo directory's config to be respected, but in reality we look for a probably nonexistent config in their `$CWD`
* Similarly, all configuration is currently conceptually "global" across all inputs, when in reality things like config ignore rules are really "local" to the repo/directory they were loaded from.

This has made it harder to make architectural changes to input collection as well (cf. #1044, #678, #1002).

## The plan/idea

Right now, we have an `InputRegistry`. This registry is more or less a flat mapping of all collected inputs, i.e. `InputKey -> AuditInput`.

The idea here is to make this a two-level mapping instead: an `InputRegistry` will contain a set of `InputGroup`s, and each `InputGroup` will contain an `InputKey -> AuditInput` mapping. 

Additioanlly, each `InputGroup` will have a `config: Option<Config>`, which will be discovered appropriately based on the "kind" of group:

* For a directory input (`zizmor foo/`) we'd attempt to discover the config relative to `foo/` like we do with `$CWD` correctly.
* For a repo slug input (`zizmor example/example`) we'd use GitHub's REST API to perform a similar discovery procedure.
* For a single file input, we'd either load no config at all (leaning towards this) or maybe walk back up the directory tree if the path to the file "makes sense" to us (e.g. is in `.github/workflows`).

With this, each group would have its own config, and the audit machinery would then load the appropriate config for an input based on its group.

For backwards compatibility, we'd also continue to support `--config` as a "global" config, which would be merged with each group's config when explicitly provided. This would then be removed with 2.0, or maybe renamed to `--global-config` with just a subset of the normal config to maybe make it clear what it actually does.

The overall compatibility ramifications of this should be pretty small -- `zizmor .` will continue to have the exact same config-loading behavior.

PR #1092, opened 2025-08-17
refactor: add groups to the input registry

woodruffw:
~~WIP.~~ Towards #1055.


woodruffw:
The basic structure of this is in place. I need to think more through `Group` and if/how I want to deduplicate that against what's already in `InputKey`.

woodruffw:
I'm happy enough with this. It might make sense at some point to dedupe the group a bit further, since it's really just the prefix/directory/whatever the input(s) were loaded from.

PR #1094, opened 2025-08-17
feat: load separate configs for input groups

### LINKED ISSUE #1055: Rethinking `zizmor`'s configuration and input collection

woodruffw:
~~Very WIP.~~

- [x] Tests for various config loading scenarios
- [x] Doc updates

Closes #1055.


woodruffw:
Thought about this a bit more, and realized that having both a per-group config and a "global" config is unnecessarily confusing and hard to make backwards compatible. Instead of doing that I'm doing to have `--config` mean that every group loads the "global" config instead. That'll then be deprecated over the coming releases.

woodruffw:
We're now honoring ignore configs on a per-group basis, although we're not yet correctly loading the "global" config over the per-group config when the user explicitly passes `--config whatever`.

woodruffw:
@funnelfiasco is there a way to make the Kusari comments a little less noisy? It'd be nice if the "rerun" comments were folded together much like how the top-level analysis comment gets updated instead of being appended :slightly_smiling_face: 

(Or even disable them outright, since the main thing they do is link back to the top comment!)

woodruffw:
Leaving some more notes for myself: now that we have multiple config files (one per group), we can't do the easy thing of loading per-audit configs in each audit's `::new`. The naive way to fix that would be to have each audit load its special config fields on demand, but that would (1) be slower, and (2) produce a worse user experience (audits randomly failing midway through on invalid configs, rather than upfront).

The more correct and slightly harder way to do this would be to split `Config` into two types: `RawConfig` for the model itself, and then `Config` for a "reified" version that pre-computes/validates per-audit configs. Most of the APIs would then only touch `Config`, which will load `RawConfig` under the hood.

PR #1103, opened 2025-08-20
chore(docs): `make pinact`

woodruffw:


woodruffw:
@kusari-inspector rerun

woodruffw:
@funnelfiasco Any thoughts on the results from Kusari's inspector above? I'm kind of confused by it since there's no details on what analysis ran (and failed), and the contents of this PR itself should be pretty innocuous (mostly docs changes plus a `mise.toml` file).

funnelfiasco:
I'll check with the team and get back to you

funnelfiasco:
Want to hear something funny? It failed running zizmor on this repo (by the way, it uses zizmor!) because we scoped the check too broadly and it was sad about `.github/pinact.yml`. We're narrowing the scope to only run when there's a change in `.github/actions` or `.github/workflows` to avoid these kinds of errors.

woodruffw:
...lol. I'm guessing you're doing something like `zizmor .github/**/*.yml` internally? If so I think it might be better to relax that to `zizmor .`, since zizmor will take care of identifying action and workflow definitions internally when given a directory instead of an expanded glob of files.

funnelfiasco:
@kusari-inspector rerun

PR #1125, opened 2025-09-01
Add azd325/gitkraken to the trophies

Azd325:


woodruffw:
Hi @Azd325, thanks for opening a PR.

Unfortunately I'm deciding to pass on this -- I appreciate you adding your repo, but the trophy case is getting pretty large and I've decided that I actually need to prune it back a bit (and raise the bar for inclusion from 100 stars).

At the same time, I really appreciate you taking the time to open a PR here (and for using zizmor of course)!

PR #1131, opened 2025-09-01
Introduce a rule which suggests that permissions are documented

### LINKED ISSUE #1127: Feature: Recommend documenting each permission
### Pre-submission checks

I noticed that many of the permissions in the workflow files that zizmor uses are documented with a trailing comment. Example: https://github.com/johnbillion/zizmor/blob/b5545d796f7c085e2832b60489e0bcdbb0f5a38e/.github/workflows/codegen.yml#L20-L23

This seems like a good idea. zizmor should include a rule which recommends this approach as a best practice.

### Additional context

_No response_

johnbillion:
Fixes #1127

This introduces a pedantic `undocumented-permissions` rule which suggests that all the permissions directives at the root level and the job level are documented with a trailing comment. The `contents: read` permission is excluded because it's self-explanatory and very common. I don't think it makes sense to require that be documented everywhere.

I've added full test coverage and updated the docs.

AI disclaimer: I've never written a line of Rust in my life. This was almost entirely produced by an AI-powered coding agent following my guidance. I'd love to know whether the code that it produced for the new rule is any good.

### DIFF
--- a/crates/zizmor/src/audit/undocumented_permissions.rs
+++ b/crates/zizmor/src/audit/undocumented_permissions.rs
@@ -1,0 +1,94 @@
+use github_actions_models::common::{Permission, Permissions};
+
+use super::{Audit, AuditLoadError, Job, audit_meta};
+use crate::finding::location::Locatable as _;
+use crate::{
+    AuditState,
+    finding::{Confidence, Persona, Severity, location::SymbolicLocation},
+};
+
+audit_meta!(
+    UndocumentedPermissions,
+    "undocumented-permissions",
+    "permissions without explanatory comments"
+);
+
+pub(crate) struct UndocumentedPermissions;
+
+impl Audit for UndocumentedPermissions {
+    fn new(_state: &AuditState) -> Result<Self, AuditLoadError>
+    where
+        Self: Sized,
+    {
+        Ok(Self)
+    }
+
+    fn audit_workflow<'doc>(
+        &self,
+        workflow: &'doc crate::models::workflow::Workflow,
+        _config: &crate::config::Config,
+    ) -> anyhow::Result<Vec<crate::finding::Finding<'doc>>> {
+        let mut findings = vec![];
+
+        // Check workflow-level permissions
+        if let Some(finding) = self.check_permissions_documentation(
+            &workflow.permissions,
+            workflow.location(),
+            workflow,
+        )? {
+            findings.push(finding);
+        }
+
+        // Check job-level permissions
+        for job in workflow.jobs() {
+            let (permissions, job_location) = match job {
+                Job::NormalJob(job) => (&job.permissions, job.location()),
+                Job::ReusableWorkflowCallJob(job) => (&job.permissions, job.location()),
+            };
+
+            if let Some(finding) =
+                self.check_permissions_documentation(permissions, job_location, workflow)?
+            {
+                findings.push(finding);
+            }
+        }
+
+        Ok(findings)
+    }
+}
+
+impl UndocumentedPermissions {
+    fn check_permissions_documentation<'a>(
+        &self,
+        permissions: &Permissions,
+        location: SymbolicLocation<'a>,
+        workflow: &'a crate::models::workflow::Workflow,
+    ) -> anyhow::Result<Option<crate::finding::Finding<'a>>> {
+        // Only check explicit permissions blocks
+        match permissions {
+            Permissions::Explicit(perms) if !perms.is_empty() => {
+                // Check if this permissions block needs documentation
+                // Skip if it only contains "contents: read" which is common and self-explanatory
+                if perms.len() == 1
+                    && perms.get("contents").map_or(false, |p| *p == Permission::Read) {
+                    return Ok(None);
+                }
+
+                // For explicit permissions, recommend documenting each permission
+                let perm_location = location.primary().with_keys(["permissions".into()]);
+                Ok(Some(
+                    Self::finding()
+                        .severity(Severity::Low)
+                        .confidence(Confidence::High)
+                        .persona(Persona::Pedantic)
+                        .add_location(perm_location.annotated(
+                            "consider adding comments to document each permission's purpose",
+                        ))
+                        .build(workflow)?,
+                ))
+            }
+            _ => Ok(None),
+        }
+    }
+}
+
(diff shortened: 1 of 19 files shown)


woodruffw:
Thanks @johnbillion! I'll do a review pass now.

> AI disclaimer: I've never written a line of Rust in my life. This was almost entirely produced by an AI-powered coding agent following my guidance. I'd love to know whether the code that it produced for the new rule is any good.

Thanks for disclaiming this 🙂

crates/zizmor/src/audit/undocumented_permissions.rs
```diff
@@ -0,0 +1,94 @@
...
+                Ok(Some(
+                    Self::finding()
+                        .severity(Severity::Low)
+                        .confidence(Confidence::High)
+                        .persona(Persona::Pedantic)
+                        .add_location(perm_location.annotated(
+                            "consider adding comments to document each permission's purpose",
+                        ))
+                        .build(workflow)?,
+                ))
```
woodruffw: I don't believe this actually checks for the presence of a comment, so it'll just blanket flag every explicit permission. Checking for the presence of a comment here would require us to concretize each permission's location, and then access its comments. Most of the current audits don't do anything like this, but there's an example of this technique in #972 here:

https://github.com/zizmorcore/zizmor/pull/972/files#diff-868df916ad72bcdc27e9c3c9117b4902d9872425aa8671131ac261251745686fR73-R78

crates/zizmor/tests/integration/test-data/undocumented-permissions.yml
woodruffw, line 1: This testcase and the other ones look great, but they aren't actually tested anywhere -- they need a matching function in `snapshot.rs` 🙂

crates/zizmor/src/audit/undocumented_permissions.rs
johnbillion: > I don't believe this actually checks for the presence of a comment

Spectacular. I'll continue working on it and improve the test coverage.

crates/zizmor/src/audit/undocumented_permissions.rs
```diff
@@ -0,0 +1,153 @@
...
+            let comment_content = comment_text
+                .strip_prefix('#')
+                .unwrap_or(comment_text)
+                .trim();
+
+            // Require a non-empty comment
+            if !comment_content.is_empty() {
+                return true;
+            }
+        }
```
woodruffw: I'd recommend using the `SymbolicLocation::concretize` API here -- that'll give you a `Location` that contains a `concrete: Feature` which in turn has a `comments` field.

(That API is basically identical internally to what's written here, but it's pre-composed and well covered by existing tests.)

crates/zizmor/tests/integration/test-data/undocumented-permissions.yml
```diff
@@ -0,0 +1,44 @@
...
+
+jobs:
+  test-job-1:
+    runs-on: ubuntu-latest
+    # This job's permissions block has a comment explaining why
+    permissions:
+      contents: read  # Need to read repo contents
+      issues: write   # Need to write issue comments
+    steps:
+      - uses: actions/checkout@v4
```
woodruffw, line 19: Nitpick: addressing the other findings here and below will make the snapshots a lot less noisy 🙂 

```suggestion
      - uses: actions/checkout@v4
        with:
          persist-credentials: false
```

crates/zizmor/src/audit/undocumented_permissions.rs
```diff
@@ -0,0 +1,136 @@
...
+        // Only create a finding if there are actually undocumented permissions
+        if !undocumented_permissions.is_empty() {
+            let perm_location = base_location.with_keys(["permissions".into()]);
+
+            Ok(Some(
+                Self::finding()
+                    .severity(Severity::Low)
+                    .confidence(Confidence::High)
+                    .persona(Persona::Pedantic)
+                    .add_location(perm_location.annotated(
```
woodruffw: We _probably_ want the `individual_perm_location` here, so that the finding's span is as minimal as possible 🙂

crates/zizmor/src/audit/undocumented_permissions.rs
```diff
@@ -0,0 +1,136 @@
...
+                        "consider adding comments to document the purpose of each permission",
```
woodruffw: Nitpick: make a bit terser:

```suggestion
                        "needs an explanatory comment",
```

crates/zizmor/src/audit/undocumented_permissions.rs
```diff
@@ -0,0 +1,136 @@
...
+        workflow: &crate::models::workflow::Workflow,
+    ) -> bool {
+        let document = workflow.as_document();
+
+        // Use the concretize API to get a Location with concrete Feature
+        let Ok(concrete_location) = location.clone().concretize(&document) else {
+            // If we can't concretize the location, assume it's undocumented to be safe.
+            // This handles rare edge cases like malformed routes or yamlpath internal errors.
+            return false;
+        };
```
woodruffw: We should bubble this error up instead of turning it into a fallback here -- the concretization would end up failing later anyways, so we might as well fail early.

woodruffw: (In other words, this should return `Result<bool>`).

crates/zizmor/src/audit/undocumented_permissions.rs
```diff
@@ -0,0 +1,136 @@
...
+
+        // Check if there are any comments
+        !concrete_location.concrete.comments.is_empty()
```
woodruffw: Nitpick: We _probably_ should check for empty comments as well, i.e. we shouldn't allow `id-token: write #` to satisfy the check here.

(Where "empty" means "only contains whitespace.")

johnbillion:
@woodruffw I was waiting for you to approve the GitHub Actions workflows so they ran :D

woodruffw:
> @woodruffw I was waiting for you to approve the GitHub Actions workflows so they ran :D

Oops, I forgot they don't auto-run 🤦

johnbillion:
@woodruffw I can't figure out what the failing snapshot is telling me. Locally everything passes, so I presume there's something different about the way the tests run on CI (maybe due to a difference with and without TTY) but I'm not sure. Any pointers?

woodruffw:
> @woodruffw I can't figure out what the failing snapshot is telling me. Locally everything passes, so I presume there's something different about the way the tests run on CI (maybe due to a difference with and without TTY) but I'm not sure. Any pointers?

I can take a look tonight! Sorry, I've been very busy.

woodruffw:
Thanks a ton @johnbillion!

PR #1150, opened 2025-09-13
docs: add coveragepy, and correct the number of stars in the instructions

nedbat:
(PS: why are all other docs commits categorized as "chore"?)


woodruffw:
> (PS: why are all other docs commits categorized as "chore"?)

Just habit on my part 😅 -- I think early in the repo I had some automation that skipped parts of the CI on "chore" changes, so I got into the habit of marking things that way. But it's purely vestigial now.

Issue #737, opened 2025-05-02
Feature: policies for `unpinned-images`


woodruffw:
Follow-on idea from #733: like with `unpinned-uses`, `unpinned-images` should be configurable with policies.

For example: I might want `tag-pin` on the default registry, but `hash-pin` on `registry.sketchy.example.com`.

woodruffw:
Some scattered thoughts about this:

1. This is a little more complicated than policies for `unpinned-uses`, since Docker/OCI image references have more moving parts: optional domain, path components, tag and hash can both be present, etc. So we'll probably want something like the current `RepositoryUses` pattern, i.e. looks a lot like a `fnmatch` pattern but is carefully controlled to stop people from shooting themselves in the foot with too broad/narrow patterns.
2. The kinds of restriction policies will probably be `any` (equivalent to `any` for `unpinned-uses`), `tag-pin` (equivalent to `ref-pin`), and `hash-pin` (equivalent to `hash-pin`).

Rough sketch:

```yaml
rules:
  unpinned-images:
    config:
      policies:
        # require hash-pinning by default
        '*': hash-pin
        # allow tag-pinning for the implicit registry, e.g. `docker pull ubuntu`
        docker.io: tag-pin
        # ...but require hash-pinning specifically for a namespace under the the implicit registry
        docker.io/sketchy/*: hash-pin
```

bart-vmware:
Is there a way to turn off this rule entirely via `zizmor.yml`? I'm asking because we reference images from our own registry, where other internal actions push to. Therefore, the versions and SHAs are constantly changing. We always need to reference the latest available.

woodruffw:
> Is there a way to turn off this rule entirely via `zizmor.yml`?

No, not yet. The idea behind the issue here is to give users the ability to disable/relax this audit on specific registries, so that they don't have to ignore it in a blanket fashion.

In other words, in your case, you'll probably want something like this:

```yaml
rules:
  unpinned-images:
    config:
      policies:
        # require hash-pinning by default
        '*': hash-pin
        # allow tag-pinning for an internal corporate registry
        docker.corp.example.com: tag-pin
```

(But note that doesn't exist yet.)

thiagowfx:
Reply to: https://github.com/zizmorcore/zizmor/issues/1115#issuecomment-3226009461:

> zizmor could keep a list of "known bad" container registries, and flag them when they're seen in inputs

[...]

PR #1151, opened 2025-09-14
feat: begin adding docker image uses patterns

woodruffw:
Towards #737.


woodruffw:
Saving for myself: https://github.com/distribution/reference/blob/main/reference.go

PR #1161, opened 2025-09-17
feat: add npm trusted publishing detection

KristianGrafana:
This PR adds comprehensive npm trusted publishing detection to zizmor.

## Changes Made

- **Add npmjs registry detection**: Detects when `actions/setup-node` uses npmjs registry with manual authentication
- **Add npm command patterns**: Detects `npm publish`, `yarn run publish`, `pnpm publish`, etc.
- **Improve hybrid scenario detection**: Flags workflows that have `id-token: write` but still use manual tokens (incomplete migration)
- **Add comprehensive tests**: 137 lines of test cases covering various npm publishing scenarios

This addresses the gap in npm package publishing security by recommending trusted publishing over manual token usage.


woodruffw:
Awesome, thanks @KristianGrafana! I'll try and review this tonight.

crates/zizmor/src/audit/use_trusted_publishing.rs
```diff
@@ -320,15 +356,20 @@ impl Audit for UseTrustedPublishing {
...
             for subfeature in self.trusted_publishing_command_candidates(run, shell)? {
+                // Adjust confidence based on whether id-token permission is present
+                let confidence = if step.parent.has_id_token() {
+                    // Higher confidence when id-token is present but manual tokens are still used
+                    // This indicates a hybrid/incomplete migration that should be flagged
+                    Confidence::High
+                } else {
+                    // Medium confidence when no id-token - could be intentional for non-TP registries
+                    Confidence::Medium
+                };
```
woodruffw: @KristianGrafana Thoughts on this being low-confidence instead? I suspect this will snare a lot of third-party index users (especially in ecosystems like Python where 3p indices are common).

woodruffw: (But I also don't feel super confident about that either, so I'm curious what you think.)

KristianGrafana: @woodruffw Good point, agree. id-token is mandatory for TP so Low makes more sense. Thanks! Added in https://github.com/zizmorcore/zizmor/pull/1161/commits/d8794bbe86474695ad550c1946a93d1bc516445f

woodruffw:
Thanks again @KristianGrafana!

PR #1183, opened 2025-09-26
fix(ref-version-mismatch): correct commit SHA in audit message

### LINKED ISSUE #1181: [BUG]: ref-version-mismatch complains but also says it's the same commit
### Pre-submission checks

### zizmor version

1.14.0

### Expected behavior

I wanted a clear problem expressed, but I found the message contradictory.

### Actual behavior

I got this complaint:
```
warning[ref-version-mismatch]: detects commit SHAs that don't match their version comment tags
  --> .github/workflows/codeql-analysis.yml:54:84
   |
54 |         uses: github/codeql-action/init@192325c86100d080feab897ff886c34abd4c83a3 # v3
   |         ------------------------------------------------------------------------   ^^ points to commit 192325c86100
   |         |
   |         is pointed to by tag v3.30.3
   |
```
The commit mentioned in the warning is the same commit in the .yaml.  Is that what it's supposed to do?

### Reproduction steps

Run zizmor on https://github.com/nedbat/coveragepy/

### Logs

```plain text

```

### Additional context

_No response_

woodruffw:
This fixes #1181 by showing the right short-SHA (the one underlying the tag, not the one we already know).

I unfortunately don't have a great way to regression-test this, since GitHub does floating tags like `# v3` that'll change too often in the snapshots. But the existing snapshot fix indirectly covers this by demonstrating that we now show the right hash.

### DIFF
--- a/crates/zizmor/src/audit/ref_version_mismatch.rs
+++ b/crates/zizmor/src/audit/ref_version_mismatch.rs
@@ -93,6 +93,8 @@ impl RefVersionMismatch {
         };
 
         if commit_for_ref != commit_sha {
+            tracing::warn!("{commit_for_ref} != {commit_sha}");
+
             let subfeature = Subfeature::new(
                 uses_location.concrete.location.offset_span.end,
                 version_from_comment,
@@ -107,7 +109,7 @@ impl RefVersionMismatch {
                     // that avoids collisions in Linux-sized repositories.
                     uses_location.symbolic.clone().primary().annotated(format!(
                         "points to commit {short_commit}",
-                        short_commit = &commit_sha[..12]
+                        short_commit = &commit_for_ref[..12]
                     )),
                     Feature::from_subfeature(&subfeature, step),
                 ));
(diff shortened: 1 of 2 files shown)


crates/zizmor/src/audit/ref_version_mismatch.rs
```diff
@@ -93,6 +93,8 @@ impl RefVersionMismatch {
         };
 
         if commit_for_ref != commit_sha {
+            tracing::warn!("{commit_for_ref} != {commit_sha}");
```
Marcono1234, line 96: Was this added intentionally, or was it only temporary for debugging?

woodruffw: Ah, that was unintentional. I'll fix that in a bit (unless you feel like sending a PR) 😉

Marcono1234: Have created #1202

PR #1187, opened 2025-09-27
fix: use-trusted-publishing: recognize crates.io Trusted Publishing for cargo publish

haya14busa:
## Summary

Avoid false positives from the `use-trusted-publishing` audit when publishing Rust crates via crates.io Trusted Publishing.

This change recognizes workflows that use `rust-lang/crates-io-auth-action` and then run `cargo publish` with the action’s OIDC-derived token (either via `env.CARGO_REGISTRY_TOKEN` or inline `--token`). In these cases, we now treat authentication as Trusted Publishing and do not flag `cargo publish`.

Docs: https://crates.io/docs/trusted-publishing

## Context

Workflows like the following correctly use Trusted Publishing, but were previously flagged with an informational finding suggesting to "prefer trusted publishing":

```yaml
permissions:
  id-token: write
steps:
  - uses: rust-lang/crates-io-auth-action@v1
    id: auth
  - run: cargo publish
    env:
      CARGO_REGISTRY_TOKEN: ${{ steps.auth.outputs.token }}
```

Because `cargo publish` appears in `run:`, the audit previously matched it as a generic publish command. This PR makes the audit aware of the crates.io TP pattern so it won’t produce a false positive in this case.

## Changes

- `use-trusted-publishing` audit:
  - If a prior step uses `rust-lang/crates-io-auth-action` (with an `id`) and the `cargo publish` step references `${{ steps.<id>.outputs.token }}` (either via `env.CARGO_REGISTRY_TOKEN` or inline `--token`), skip the finding (treat as Trusted Publishing).
  - Scope limited to `cargo publish` commands.

- Tests:
  - Added `use-trusted-publishing/cargo-publish-trusted.yml` input that exercises the crates.io Trusted Publishing flow.
  - Added a new integration snapshot asserting no findings for the TP case.

## Why this is safe

crates.io Trusted Publishing uses OIDC to issue a scoped token at runtime for the publishing workflow. Using the action’s output token (derived from OIDC) is materially different from using a long-lived secret and should not be flagged as "manual" credential usage.

Reference: https://crates.io/docs/trusted-publishing

## Example behavior

- Trusted (no finding):
  - Previous step: `uses: rust-lang/crates-io-auth-action@v1` with `id: auth`.
  - `run: cargo publish` with `env.CARGO_REGISTRY_TOKEN: ${{ steps.auth.outputs.token }}`.

- Not trusted (finding preserved):
  - `run: cargo publish` with `env.CARGO_REGISTRY_TOKEN: ${{ secrets.CRATESIO_PUBLISH_TOKEN }}` or similar manually configured credentials.

### DIFF
--- a/crates/zizmor/src/audit/use_trusted_publishing.rs
+++ b/crates/zizmor/src/audit/use_trusted_publishing.rs
@@ -201,6 +201,77 @@ audit_meta!(
 );
 
 impl UseTrustedPublishing {
+    // Detect whether this job is using crates.io Trusted Publishing by checking for
+    // a prior step that uses rust-lang/crates-io-auth-action and that this step
+    // wires its token into cargo via env or inline token usage.
+    fn uses_crates_io_trusted_publishing(
+        &self,
+        step: &crate::models::workflow::Step,
+        run: &str,
+    ) -> bool {
+        // Collect IDs of prior steps in the same job that use rust-lang/crates-io-auth-action
+        let mut auth_step_ids: Vec<&str> = Vec::new();
+
+        for prior in step.job().steps() {
+            if prior.index >= step.index {
+                break;
+            }
+
+            // Only consider uses: steps that reference the crates-io auth action
+            let Some(common_uses) = prior.uses() else { continue };
+            let github_actions_models::common::Uses::Repository(repo_uses) = common_uses else {
+                continue;
+            };
+
+            if repo_uses.owner.eq_ignore_ascii_case("rust-lang")
+                && repo_uses.repo.eq_ignore_ascii_case("crates-io-auth-action")
+            {
+                if let Some(id) = &prior.id {
+                    auth_step_ids.push(id);
+                }
+            }
+        }
+
+        if auth_step_ids.is_empty() {
+            return false;
+        }
+
+        // Helper: does a string reference any of the known auth step outputs?
+        let references_auth_output = |s: &str| {
+            auth_step_ids
+                .iter()
+                .any(|id| s.contains(&format!("steps.{id}.outputs.token")))
+        };
+
+        // 1) Check inline usage in the run: body, e.g.:
+        //    cargo publish --token ${{ steps.auth.outputs.token }}
+        if references_auth_output(run) {
+            return true;
+        }
+
+        // 2) Check env wiring, e.g.:
+        //    env:
+        //      CARGO_REGISTRY_TOKEN: ${{ steps.auth.outputs.token }}
+        match &step.env {
+            github_actions_models::common::expr::LoE::Literal(env) => {
+                if let Some(val) = env.get("CARGO_REGISTRY_TOKEN") {
+                    if references_auth_output(&val.to_string()) {
+                        return true;
+                    }
+                }
+            }
+            _ => {}
+        }
+
+        false
+    }
+
+    fn is_cargo_publish_command(cmd: &str) -> bool {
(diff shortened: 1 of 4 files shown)


woodruffw:
Hey @haya14busa, thanks for the PR. I'll be able to give this a full review hopefully later today 🙂

woodruffw:
FYI some of the false positives here were possibly fixed with #1191, which fixes a regression introduced with #1161. I'll still do a full review here, however 🙂

crates/zizmor/src/audit/use_trusted_publishing.rs
```diff
@@ -201,6 +201,77 @@ audit_meta!(
...
+            }
+
+            // Only consider uses: steps that reference the crates-io auth action
+            let Some(common_uses) = prior.uses() else { continue };
+            let github_actions_models::common::Uses::Repository(repo_uses) = common_uses else {
+                continue;
+            };
+
+            if repo_uses.owner.eq_ignore_ascii_case("rust-lang")
+                && repo_uses.repo.eq_ignore_ascii_case("crates-io-auth-action")
```
woodruffw, line 227: Could you write this as a `RepositoryUsesPattern`? That'll ensure we fully match all of the different ways a user could write this `uses:` clause.

(It implements `FromStr`, so you can do `"rust-lang/crates-io-auth-action".parse().unwrap()` to get one.)

crates/zizmor/src/audit/use_trusted_publishing.rs
```diff
@@ -201,6 +201,77 @@ audit_meta!(
...
+
+        if auth_step_ids.is_empty() {
+            return false;
+        }
+
+        // Helper: does a string reference any of the known auth step outputs?
+        let references_auth_output = |s: &str| {
+            auth_step_ids
+                .iter()
+                .any(|id| s.contains(&format!("steps.{id}.outputs.token")))
```
woodruffw, line 243: Similarly: could you write this as a `ContextPattern`? That'll ensure we match the various syntax variants context accesses.

(It also implements `FromStr`)

crates/zizmor/src/audit/use_trusted_publishing.rs
```diff
@@ -356,6 +427,21 @@ impl Audit for UseTrustedPublishing {
...
+                    subfeature::Fragment::Raw(cmd) => {
+                        Self::is_cargo_publish_command(cmd)
+                            && self.uses_crates_io_trusted_publishing(step, run)
+                    }
+                    _ => false,
+                };
+
+                if skip_for_crates_io_tp {
+                    continue;
+                }
```
woodruffw, line 443: Hmm -- I think we'd ideally skip these lower-level checks entirely, i.e. we should probably remove `cargo publish` from `self.trusted_publishing_command_candidates` now that it's specialized like this.

haya14busa:
Thanks for the fix and the review!

I confirmed the false positive was fixed at HEAD. 
Do we still want to address comments and include this change? 
While it's more precise if we check `crates-io-auth-action`, I'm fine with the current fix which checks id-token permission.

woodruffw:
> Do we still want to address comments and include this change?
> While it's more precise if we check `crates-io-auth-action`, I'm fine with the current fix which checks id-token permission.

I think so, although I think there's probably a bigger refactor that needs to happen first (not your fault or problem, the existing structure doesn't really make it easy to do these kinds of mixed uses/run checks!). So I'm tempted to close this for now and revisit it later, since as you note the false positive is currently gone 🙂 

Thanks again for submitting here!

PR #1200, opened 2025-09-30
perf: use jemalloc as the global allocator on non-Windows

woodruffw:
I'm not sure how much of a boost this will be; we'll see.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2652,6 +2652,26 @@ dependencies = [
 ]
 
 [[package]]
+name = "tikv-jemalloc-sys"
+version = "0.6.0+5.3.0-1-ge13ca993e8ccb9ba9847cc330696e02839f328f7"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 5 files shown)


woodruffw:
This shows a >20% win on CI, which is huge -- big enough that it tripped the alert for a potential regression! I don't see as big of a win locally (on macOS + arm64), but even locally I see 5-7% improvements with jemalloc. Not bad for a 10 line change.

striezel:
@woodruffw:
As far as I can tell from this pull request, jemalloc is intended as allocator  "on non-Windows" systems, but the code actually only checks for non-MSVC, so that change affects people using a GCC or Clang toolchain on Windows, like for example MSYS2 does.

Basically, if that change is meant for non-Windows systems, it should also say so in the code and not assume that MSVC is the only option available on Windows, even though MSVC is probably the most common one for Windows.

woodruffw:
Hey @striezel, thanks for raising that. From my read of the Rust docs, I should probably be using `target_family = "windows"` here instead. Does that seem right to you?

striezel:
Yes, `target_family = "windows"` is definitely better and should fix it.

PR #1205, opened 2025-10-01
Add Fix for ref-version-mismatch audit rule

mostafa:
This PR adds auto-fix for ref-version-mismatch audit rule with online tests against GitHub API.

xref #876

### DIFF
--- a/crates/zizmor/src/audit/ref_version_mismatch.rs
+++ b/crates/zizmor/src/audit/ref_version_mismatch.rs
@@ -262,27 +262,22 @@ jobs:
             .unwrap();
 
         // We expect at least one finding if there's a version mismatch
-        if !findings.is_empty() {
-            if !findings[0].fixes.is_empty() {
-                let new_doc = findings[0].fixes[0].apply(input.as_document()).unwrap();
-                insta::assert_snapshot!(new_doc.source(), @r"
-                name: Test Version Comment Mismatch
-                on: push
-                permissions: {}
-                jobs:
-                  test:
-                    runs-on: ubuntu-latest
-                    steps:
-                      - name: Checkout with mismatched version comment
-                        uses: actions/checkout@a81bbbf8298c0fa03ea29cdc80d8d0ce8b6c2f2c # v3.0.2
-                ");
-            } else {
-                println!(
-                    "Finding detected but no fix available - commit hash may not have a corresponding tag"
-                );
-            }
-        } else {
-            println!("No findings detected - commit hash may not have a version mismatch");
+        assert!(!findings.is_empty(), "Expected to find version mismatch");
+
+        // Only test the fix if one is available (depends on GitHub API response)
+        if !findings[0].fixes.is_empty() {
+            let new_doc = findings[0].fixes[0].apply(input.as_document()).unwrap();
+            insta::assert_snapshot!(new_doc.source(), @r"
+            name: Test Version Comment Mismatch
+            on: push
+            permissions: {}
+            jobs:
+              test:
+                runs-on: ubuntu-latest
+                steps:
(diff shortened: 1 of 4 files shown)


crates/zizmor/src/audit/ref_version_mismatch.rs
```diff
@@ -160,3 +182,178 @@ impl Audit for RefVersionMismatch {
...
+                ");
+            } else {
+                println!(
+                    "Finding detected but no fix available - commit hash may not have a corresponding tag"
+                );
+            }
+        } else {
+            println!("No findings detected - commit hash may not have a version mismatch");
+        }
+    }
```
woodruffw: Could we drop these prints? I don't think they're super helpful outside of initial test validation.

crates/zizmor/src/audit/ref_version_mismatch.rs
```diff
@@ -160,3 +182,178 @@ impl Audit for RefVersionMismatch {
...
+                    uses: actions/checkout@a81bbbf8298c0fa03ea29cdc80d8d0ce8b6c2f2c # v3.0.2
+                  - name: Simple format
+                    uses: actions/checkout@a81bbbf8298c0fa03ea29cdc80d8d0ce8b6c2f2c # v3.0.0
+                  - name: Version format
+                    uses: actions/checkout@a81bbbf8298c0fa03ea29cdc80d8d0ce8b6c2f2c # version: 3.0.0
+            ");
+        } else {
+            println!(
+                "No fixes available - this is expected if the commit hash doesn't have corresponding tags"
+            );
```
woodruffw: Same here 🙂

crates/zizmor/src/audit/ref_version_mismatch.rs
```diff
@@ -160,3 +182,178 @@ impl Audit for RefVersionMismatch {
...
+
+        let audit = RefVersionMismatch::new(&state).unwrap();
+
+        let input = workflow.into();
+        let findings = audit
+            .audit(RefVersionMismatch::ident(), &input, &Config::default())
+            .unwrap();
+
+        // We expect at least one finding if there's a version mismatch
+        if !findings.is_empty() {
```
woodruffw: This should probably be an assert instead, i.e. the test should fail if we don't get a mismatch here?

crates/zizmor/src/audit/ref_version_mismatch.rs
```diff
@@ -160,3 +182,178 @@ impl Audit for RefVersionMismatch {
...
+            if !findings[0].fixes.is_empty() {
```
woodruffw: Same here -- needs to be an assert, I think.

crates/zizmor/src/audit/ref_version_mismatch.rs
```diff
@@ -160,3 +182,178 @@ impl Audit for RefVersionMismatch {
...
+        );
+
+        let audit = RefVersionMismatch::new(&state).unwrap();
+
+        let input = workflow.into();
+        let findings = audit
+            .audit(RefVersionMismatch::ident(), &input, &Config::default())
+            .unwrap();
+
+        if !findings.is_empty() && !findings[0].fixes.is_empty() {
```
woodruffw: Needs to be an assert instead of a condition, I think.

woodruffw (commented):
Thanks @mostafa! This looks good to me overall, some test nitpicks.

mostafa:
@woodruffw Thanks for the review. Addressed all your comments in a single commit.

woodruffw (approved):

woodruffw:
Thanks @mostafa!

PR #1213, opened 2025-10-05
Add `--fix-format` parameter for JSON Patch output

mostafa:
This is a POC of https://github.com/zizmorcore/zizmor/issues/876#issuecomment-3368804575.


mostafa:
I'll close this until we actually decide to do it, rather than being nerd-sniped.

Issue #1225, opened 2025-10-09
Feature: green software practices


joewallwork:
### Pre-submission checks

- [x] I am **not** reporting a bug (crash, false positive/negative, etc). These must be filed via the bug report template.
- [x] I have looked through both the [open](https://github.com/zizmorcore/zizmor/issues?q=is%3Aissue+is%3Aopen+) and [closed](https://github.com/zizmorcore/zizmor/issues?q=is%3Aissue%20state%3Aclosed) issues for a duplicate request.

### What's the problem this feature will solve?

Thanks for developing zizmor - I've found it to be a really useful tool and am making use of it in various projects.

I've been thinking about green software engineering practices and have recently been looking into green CI, i.e., reducing resource wastage in CI workflows. Some colleagues and I created https://github.com/Cambridge-ICCS/green-ci during a hackathon and it documents some of the things I mean.

### Describe the solution you'd like

It'd be great to add some such rules to zizmor to encourage green CI usage. What I propose:
- [ ] 1. Always include a timeout on jobs to avoid wasted resources when jobs stall (should be fixed by https://github.com/zizmorcore/zizmor/pull/1067).
- [x] 2. Always set up a `concurrency` check that cancels running workflows when they get re-triggered (you don't need the original jobs to finish if you've already updated the branch).
- [ ] 3. Use a separation of concerns approach to separate out workflows (tests, static analysis, docs) and only trigger each when associated files change.

If you agree that 2 sounds like a good idea then I'd be happy to take it on. I've been learning Rust this week and it'd be a great way to practice.

With 3 it wouldn't apply in all cases and doesn't seem like a good fit for a hard rule, but it might be nice to have a warning or info message to encourage this?

woodruffw:
Hi @jwallwork23, thanks for opening this issue!

I'll respond to each of your specific ideas below (all of which sound good to me), but as a more categorical response first: the goal with zizmor is primarily *security* audits, not general code quality or other IMO worthwhile checks. If you're looking to write checks that don't have a security angle for the GitHub Actions ecosystem, I would definitely recommend checking out `actionlint` -- they do a much larger set of CQA checks than zizmor intends to 🙂

With that said, I _do_ think there's a strong argument to be made that "resource waste" vectors _can_ be a security risk, so that's not a categorical rejection! It's just something I wanted to broadcast as a deciding factor for whether I'm likely to accept an audit.

Now, for the specific ideas:

[...]

PR #1227, opened 2025-10-10
New audit: concurrency-limits

joewallwork:
Towards #1225.

Okay I think I'm most of the way there, although there are still a few things to do:

- [x] Fix the `FIXME` check  - it says `cancel-in-progress` is false even when it's true.
- [x] ~~Check the `group` entry as well as the `cancel-in-progress` one.~~
- [x] Write a test that checks `cancel-in-progress: true` doesn't raise an issue.
- [x] Write docs for the change.

Also, @woodruffw I thought I'd let you confirm whether you're happy with how this will work before I go through and update all the other test data to include concurrency+cancel-in-progress. (Currently the tests raise lots of errors.)

### DIFF
--- a/crates/zizmor/src/audit/concurrency_limits.rs
+++ b/crates/zizmor/src/audit/concurrency_limits.rs
@@ -77,7 +77,6 @@ impl Audit for ConcurrencyLimits {
                             workflow
                                 .location()
                                 .primary()
-                                .with_keys(["name".into()])
                                 .annotated("missing concurrency setting"),
                         )
                         .build(workflow)?,
--- a/crates/zizmor/tests/integration/snapshot.rs
+++ b/crates/zizmor/tests/integration/snapshot.rs
@@ -1047,12 +1047,17 @@ fn concurrency_limits() -> Result<()> {
             .run()?,
         @r"
     help[concurrency-limits]: insufficient job-level concurrency limits
-     --> @@INPUT@@:1:1
-      |
-    1 | name: Workflow without concurrency
-      | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ missing concurrency setting
-      |
-      = note: audit confidence → High
+      --> @@INPUT@@:1:1
+       |
+     1 | / name: Workflow without concurrency
+     2 | | on: push
+     3 | | permissions: {}
+    ...  |
+    10 | |     - name: 1-ok
+    11 | |       run: echo ok
+       | |___________________^ missing concurrency setting
+       |
+       = note: audit confidence → High
 
     1 finding: 0 informational, 1 low, 0 medium, 0 high
     "
--- a/crates/zizmor/tests/integration/snapshots/integration__e2e__gha_hazmat.snap
+++ b/crates/zizmor/tests/integration/snapshots/integration__e2e__gha_hazmat.snap
@@ -1000,4 +1000,4 @@ error[dangerous-triggers]: use of fundamentally insecure workflow trigger
    |
    = note: audit confidence → Medium
 
-162 findings (76 suppressed): 7 informational, 0 low, 29 medium, 50 high
+183 findings (97 suppressed): 7 informational, 0 low, 29 medium, 50 high
--- a/crates/zizmor/tests/integration/snapshots/integration__e2e__issue_569.snap
+++ b/crates/zizmor/tests/integration/snapshots/integration__e2e__issue_569.snap
@@ -1,6 +1,6 @@
 ---
 source: crates/zizmor/tests/integration/e2e.rs
-expression: "zizmor().offline(false).output(OutputMode::Both).args([\"--no-online-audits\",\n\"--collect=workflows-only\"]).input(\"python/cpython@f963239ff1f986742d4c6bab2ab7b73f5a4047f6\").run()?"
+expression: "zizmor().offline(false).output(OutputMode::Both).args([\"--no-online-audits\",\n\"--collect=workflows\"]).input(\"python/cpython@f963239ff1f986742d4c6bab2ab7b73f5a4047f6\").run()?"
 ---
 🌈 zizmor v@@VERSION@@
  INFO zizmor::registry: skipping impostor-commit: offline audits only requested
@@ -173,4 +173,4 @@ error[unpinned-uses]: unpinned action reference
    |
    = note: audit confidence → High
 
-85 findings (1 ignored, 66 suppressed): 0 informational, 1 low, 0 medium, 17 high
+97 findings (1 ignored, 78 suppressed): 0 informational, 1 low, 0 medium, 17 high
--- a/crates/zizmor/tests/integration/snapshots/integration__e2e__issue_726.snap
+++ b/crates/zizmor/tests/integration/snapshots/integration__e2e__issue_726.snap
@@ -15,4 +15,4 @@ expression: "zizmor().offline(false).output(OutputMode::Both).args([\"--no-onlin
  INFO audit: zizmor: 🌈 completed .github/workflows/hello.yml
  INFO audit: zizmor: 🌈 completed arbitrary/subdir/.github/workflows/hello.yml
  INFO audit: zizmor: 🌈 completed arbitrary/subdir/custom-action/action.yml
-No findings to report. Good job! (2 suppressed)
+No findings to report. Good job! (4 suppressed)
--- a/crates/zizmor/tests/integration/snapshots/integration__snapshot__artipacked-4.snap
+++ b/crates/zizmor/tests/integration/snapshots/integration__snapshot__artipacked-4.snap
@@ -1,28 +1,19 @@
 ---
 source: crates/zizmor/tests/integration/snapshot.rs
-assertion_line: 83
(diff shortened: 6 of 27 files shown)


woodruffw:
Thanks @jwallwork23, I'll try and find some time to review this in the upcoming week!

crates/zizmor/src/audit/concurrency_limits.rs
woodruffw, line 1: Nitpick: I'd suggest naming this `concurrency-limits` or similar (and then `ConcurrencyLimits`, etc.). My thinking there is that it's nice to gave more generic audit names, since we might put other kinds of concurrency checks under this ID in the future.

crates/zizmor/src/audit/concurrency_cancel.rs
```diff
@@ -0,0 +1,95 @@
...
+};
+use anyhow::Result;
+use github_actions_models::{common::expr::BoE, workflow::Concurrency};
+
+pub(crate) struct ConcurrencyCancel;
+
+audit_meta!(
+    ConcurrencyCancel,
+    "concurrency-cancel",
+    "cancel running jobs when they are re-triggered"
```
woodruffw: Nitpick: this would match the other audit descriptions, which tend to be descriptive rather than imperative:

```suggestion
    "insufficient job-level concurrency limits"
```

crates/zizmor/src/audit/concurrency_limits.rs
```diff
@@ -0,0 +1,95 @@
...
+                cancel_in_progress,
+            }) => {
+                match &cancel_in_progress {
+                    BoE::Literal(cancel) => {
+                        // FIXME: It's saying false even when true
+                        println!("cancel-in-progress is set to {cancel}");
+                        if !cancel {
+                            findings.push(
+                                Self::finding()
+                                    .confidence(Confidence::High)
```
woodruffw: Nit: I'd suggest making this (and the other findings) pedantic-only for the time being, since I suspect they'll be pretty noisy for the average user. We can evaluate dropping them to the "normal" i.e. default persona over time 🙂 

(That'll be `.persona(Persona::Pedantic)` in this context.)

crates/zizmor/src/audit/concurrency_cancel.rs
```diff
@@ -0,0 +1,95 @@
...
+            }) => {
+                match &cancel_in_progress {
+                    BoE::Literal(cancel) => {
+                        // FIXME: It's saying false even when true
+                        println!("cancel-in-progress is set to {cancel}");
+                        if !cancel {
+                            findings.push(
+                                Self::finding()
+                                    .confidence(Confidence::High)
+                                    .severity(Severity::Medium)
```
woodruffw: Nit: I *think* we should probably consider this low- or informational-severity for the time being. My rationale for that is that medium-severity is typically reserved for things that have a semi-immediate security impact, while in this case the security impact is mostly on the nuisance/resource waste side.

crates/zizmor/tests/integration/test-data/concurrency-limits/cancel-false.yml
woodruffw, line 1: Annoying nitpick on my part, but please `kebab-case.yml` these test files like the others 🙂

crates/zizmor/src/audit/concurrency_cancel.rs
```diff
@@ -0,0 +1,95 @@
...
+        match &workflow.concurrency {
+            Some(Concurrency::Rich {
+                group,
+                cancel_in_progress,
+            }) => {
+                match &cancel_in_progress {
+                    BoE::Literal(cancel) => {
+                        // FIXME: It's saying false even when true
+                        println!("cancel-in-progress is set to {cancel}");
+                        if !cancel {
```
woodruffw: I fixed this with [`17bf424` (#1227)](https://github.com/zizmorcore/zizmor/pull/1227/commits/17bf4247f800b28b3e3813e7ecf74f24aa888009) -- serde has a slightly confusing distinction between `rename_all` and `rename_all_fields`, and the old enum definition was using the wrong one. Sorry about that!

joewallwork: Oh great, thanks!

crates/zizmor/src/audit/concurrency_limits.rs
joewallwork: Addressed in 3f160a53fa51a428b54f077c1bf25141fdb56be1.

crates/zizmor/tests/integration/test-data/concurrency-limits/cancel-false.yml
joewallwork: Addressed in 3f160a53fa51a428b54f077c1bf25141fdb56be1.

crates/zizmor/src/audit/concurrency_cancel.rs
joewallwork: Done in ed3474c90dd56818eadec5248ce345d5b5ca98f6.

crates/zizmor/src/audit/concurrency_limits.rs
joewallwork: Done in 2c628032322e651625661434d7d37d655cbf6c59.

joewallwork:
Returning to this, a few updates:
* I decided not to check for the `group` entry here because I couldn't think of a meaningful check that could be done statically.
* Added some docs in https://github.com/zizmorcore/zizmor/pull/1227/commits/e377893dd49b90f618b7f7fc7926ca9b828993ef, hopefully they're along the right lines.
* I updated the snapshot tests, hopefully I did that right, too.

crates/zizmor/tests/integration/test-data/concurrency-limits/cancel-false.yml
woodruffw: Hmm, did this not get applied? These files are still `snake_case` not `kebab-case`.

docs/audits.md
```diff
@@ -246,6 +246,41 @@ intended to publish build artifacts:
 * Set an action-specific input to disable cache restoration when appropriate,
   such as `lookup-only` in @Swatinem/rust-cache.
 
+## `concurrency-limits`
+
+| Type     | Examples                | Introduced in | Works offline  | Enabled by default | Configurable |
+|----------|-------------------------|---------------|----------------|--------------------| ---------------|
+| Workflow | [concurrency-limits/]   | v1.16.0       | ✅             | ✅                 | ❌  |
```
woodruffw: Nit: the table actually changed a bit recently, the `Enabled by default` column should now be `Auto-fixes available` and should be an ❌ in this case:

```suggestion
| Type     | Examples                | Introduced in | Works offline  | Auto-fixes available | Configurable |
|----------|-------------------------|---------------|----------------|--------------------| ---------------|
| Workflow | [concurrency-limits/]   | v1.16.0       | ✅             | ❌                 | ❌  |
```

docs/audits.md
```diff
@@ -246,6 +246,41 @@ intended to publish build artifacts:
...
+
+Include a `concurrency` setting in your workflow that sets the
+`cancel-in-progress` option either to `true` or to an expression that will be
+true in most cases. Specifying `false` would allow separate instances of the
+workflows to run concurrently, whereas `true` will imply that running jobs are
+cancelled as soon as the workflow is re-triggered.
+
+!!! example
+
+    ```yaml title="cancel_true.yml"
```
woodruffw:
```suggestion
    ```yaml title="cancel-true.yml"
```

crates/zizmor/tests/integration/test-data/concurrency-limits/cancel-false.yml
joewallwork: Apologies, I missed some. Fixed in 857c3e7c60c91a427d01b8bf94167f7da0707c94.

joewallwork:
> Thanks @jwallwork23, this looks good to me overall! Just some small nitpicks left.

Great, thanks @woodruffw! I think I've addressed all your remaining comments now.

woodruffw:
> > Thanks @jwallwork23, this looks good to me overall! Just some small nitpicks left.
> 
> Great, thanks @woodruffw! I think I've addressed all your remaining comments now.

Yep, thank you! I'll deconflict this and plan it for tomorrow.

crates/zizmor/src/audit/concurrency_limits.rs
```diff
@@ -0,0 +1,90 @@
...
+                findings.push(
+                    Self::finding()
+                        .confidence(Confidence::High)
+                        .severity(Severity::Low)
+                        .persona(Persona::Pedantic)
+                        .add_location(
+                            workflow
+                                .location()
+                                .primary()
+                                .with_keys(["name".into()])
```
woodruffw: I'll fix it here, but noting for future reference: workflow names are optional, so this would fail on some real-world workflow inputs (which the online tests caught) 🙂

woodruffw:
Thanks @jwallwork23!

crates/zizmor/src/audit/concurrency_limits.rs
joewallwork: Ah okay, good point.

PR #1229, opened 2025-10-12
Add Fixes for dependabot-cooldown and dependabot-execution audit rules

mostafa:
xref #876

### DIFF
--- a/docs/audits.md
+++ b/docs/audits.md
@@ -377,7 +377,7 @@ Some general pointers:
 
 | Type     | Examples                | Introduced in | Works offline  | Auto-fixes available | Configurable |
 |----------|-------------------------|---------------|----------------|--------------------| ---------------|
-| Dependabot  | [dependabot-cooldown/]       | v1.15.0       | ✅             | ❌                 | ❌  |
+| Dependabot  | [dependabot-cooldown/]       | v1.15.0       | ✅             | ✅                 | ❌  |
 
 [dependabot-cooldown/]: https://github.com/zizmorcore/zizmor/blob/main/crates/zizmor/tests/integration/test-data/dependabot-cooldown/
 
@@ -443,7 +443,7 @@ enforces the following minimums:
 
 | Type     | Examples                | Introduced in | Works offline  | Auto-fixes available | Configurable |
 |----------|-------------------------|---------------|----------------|--------------------| ---------------|
-| Dependabot  | [dependabot-execution/]       | v1.15.0       | ✅             | ❌                | ❌  |
+| Dependabot  | [dependabot-execution/]       | v1.15.0       | ✅             | ✅                | ❌  |
 
 [dependabot-execution/]: https://github.com/zizmorcore/zizmor/blob/main/crates/zizmor/tests/integration/test-data/dependabot-execution/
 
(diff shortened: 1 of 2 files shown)


crates/zizmor/src/audit/dependabot_execution.rs
```diff
@@ -13,6 +14,24 @@ audit_meta!(
 
 pub(crate) struct DependabotExecution;
 
+impl DependabotExecution {
+    /// Creates a fix that changes insecure-external-code-execution from allow to deny
+    fn create_set_deny_fix<'doc>(update: crate::models::dependabot::Update<'doc>) -> Fix<'doc> {
+        Fix {
+            title: "set insecure-external-code-execution to deny".to_string(),
+            key: update.location().key,
+            disposition: FixDisposition::Safe,
```
woodruffw: Nit: We should probably make this one as unsafe for now, since I imagine people often *do* have this set for load-bearing reasons 🙂

woodruffw:
~~This changes look great to me (besides a small nit), although I'm kind of confused as to why the snapshots didn't change -- I would have expected the Dependabot snapshots to get updated with "fixes available" renders.~~

Never mind, I forgot I had inlined these snapshots and was looking in the wrong place 🤦

crates/zizmor/src/audit/dependabot_cooldown.rs
```diff
@@ -85,3 +150,209 @@ impl Audit for DependabotCooldown {
...
+            DependabotCooldown,
+            "test_fix_multiple_updates.yml",
+            dependabot_content,
+            |_dependabot: &Dependabot, findings: Vec<crate::finding::Finding>| {
+                insta::assert_snapshot!(findings.len(), @"2");
+
+                // Verify both findings have fixes
+                for finding in &findings {
+                    assert!(!finding.fixes.is_empty(), "Expected fixes but got none");
+                }
```
woodruffw: Can we snapshot the full change here, rather than asserting the findings length/fixes status?

crates/zizmor/src/audit/dependabot_cooldown.rs
```diff
@@ -85,3 +150,209 @@ impl Audit for DependabotCooldown {
...
+    schedule:
+      interval: daily
+"#;
+
+        test_dependabot_audit!(
+            DependabotCooldown,
+            "test_no_fix_needed.yml",
+            dependabot_content,
+            |_dependabot: &Dependabot, findings: Vec<crate::finding::Finding>| {
+                insta::assert_snapshot!(findings.len(), @"0");
```
woodruffw: Same as above 🙂

crates/zizmor/src/audit/dependabot_execution.rs
```diff
@@ -50,3 +70,146 @@ impl Audit for DependabotExecution {
...
+      interval: daily
+    insecure-external-code-execution: deny
+"#;
+
+        test_dependabot_audit!(
+            DependabotExecution,
+            "test_no_fix_needed_for_deny.yml",
+            dependabot_content,
+            |_dependabot: &Dependabot, findings: Vec<crate::finding::Finding>| {
+                insta::assert_snapshot!(findings.len(), @"0");
```
woodruffw: Same as above 🙂 (etc. for the other tests below too)

mostafa:
@woodruffw Thanks for the review. I added snapshot testing for all tests.

PR #1232, opened 2025-10-13
feat: stabilize the auto-fix mode

woodruffw:
See #876.


woodruffw:
NTS: Need to update the audit metadata tables to indicate whether each audit supports auto-fixes.

PR #1242, opened 2025-10-14
Fix exit code to return 0 when all findings are auto-fixable

### LINKED ISSUE #1189: [BUG]: Does not return exit code 0 even if all findings are auto-fixable
### Pre-submission checks

### zizmor version

1.12.1

### Expected behavior

Exit code 0 when all findings are auto-fixable

### Actual behavior

Exit code 11-14

### Reproduction steps

Given a GitHub Action workflow with:
```yml
- uses: actions/checkout@v5
```
run zizmor against it with `--fix`

### Logs

```plain text

```

### Additional context

This causes tools that depend on zizmor to not know whether there are any findings that cannot be auto-fixed, e.g. [super-linter](https://github.com/super-linter/super-linter) (see https://github.com/super-linter/super-linter/issues/7067#issuecomment-3341834611)

cnaples79:
## Summary
- Implements exit code 0 when all findings matching the fix mode are successfully auto-fixed
- Adds tracking of fix application success/failure counts
- Adds method to check if all findings have applicable fixes for a given mode

### `crates/zizmor/src/output/fix.rs`
- Added `FixResult` struct to track applied and failed fix counts
- Modified `apply_fixes()` to return `FixResult` instead of `()`
- Track successful and failed fix applications

### `crates/zizmor/src/registry.rs`
- Added `all_findings_have_applicable_fixes()` method to `FindingRegistry`
- Checks if every finding has at least one fix matching the specified fix mode
- Returns true if no manual intervention would be required

### `crates/zizmor/src/main.rs`
- Updated exit code logic to return `ExitCode::SUCCESS` when all findings are auto-fixed
- Considers both fix availability and application success
- Maintains backward compatibility for cases without `--fix` or with manual fixes required

## Behavior

The new exit code logic works as follows:

- **`--fix` / `--fix=safe`**: Exit with 0 if all findings have safe fixes that apply successfully, but still exit with 11-14 if any findings only have unsafe fixes
- **`--fix=all`**: Exit with 0 if all findings have fixes (safe or unsafe) that apply successfully  
- **`--fix=unsafe-only`**: Exit with 0 if all findings have unsafe fixes that apply successfully, but still exit with 11-14 if any findings only have safe fixes
- **No `--fix`**: Maintains current behavior (exit code 11-14 based on severity)

In short: exit with 0 if there are no manual fixes required.

## Test Plan
- Verified code compiles with `cargo check --lib`
- Logic reviewed for correctness with fix mode combinations
- Handles edge cases: no findings, partial fix failures, mixed fix dispositions

Fixes #1189

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -840,7 +840,7 @@ dependencies = [
 
 [[package]]
 name = "github-actions-models"
-version = "0.33.0"
+version = "0.36.0"
 dependencies = [
  "indexmap",
  "serde",
@@ -3779,7 +3779,7 @@ dependencies = [
 
 [[package]]
 name = "zizmor"
-version = "1.15.0"
+version = "1.15.2"
 dependencies = [
  "annotate-snippets",
  "anstream",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,7 +20,7 @@ rust-version = "1.88.0"
 [workspace.dependencies]
 anyhow = "1.0.100"
 github-actions-expressions = { path = "crates/github-actions-expressions", version = "0.0.10" }
-github-actions-models = { path = "crates/github-actions-models", version = "0.33.0" }
+github-actions-models = { path = "crates/github-actions-models", version = "0.36.0" }
 itertools = "0.14.0"
 pest = "2.8.3"
 pest_derive = "2.8.3"
--- a/crates/github-actions-models/Cargo.toml
+++ b/crates/github-actions-models/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "github-actions-models"
-version = "0.33.0"
+version = "0.36.0"
 description = "Unofficial, high-quality data models for GitHub Actions workflows, actions, and related components"
 repository = "https://github.com/zizmorcore/zizmor/tree/main/crates/github-actions-models"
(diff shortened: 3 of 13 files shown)


woodruffw:
Hi @cnaples79, thanks for the PR!

Out of curiosity, did you use an LLM for all or part of this work? It's no problem if you did, but I'm trying to gauge their use by contributors as part of this potential change in policy: https://github.com/orgs/zizmorcore/discussions/1238

cnaples79:
@woodruffw you're welcome!

I use autocomplete inside of either Windsurf or VS Code usually when I'm coding. So, I think technically that would mean yes?

I've looked at the change policy and think it's a good idea to implement. If you'd like to discuss more let me know! I follow research on agentic tools very closely.

If I need to update anything with the PR itself jist let me know!

woodruffw:
> I use autocomplete inside of either Windsurf or VS Code usually when I'm coding. So, I think technically that would mean yes?

Yeah, I would count that as a 'yes' 🙂 

And thanks! The PR looks great to me overall, I'll do another review tonight.

woodruffw:
(If you feel like it, the usage docs on exit codes could also use updating to match the new behavior.)

woodruffw:
Thanks @cnaples79!

Issue #1246, opened 2025-10-14
Feature: Support expanded Dependabot schedule intervals


kingsword09:
### Pre-submission checks

- [x] I am **not** reporting a bug (crash, false positive/negative, etc). These must be filed via the bug report template.
- [x] I have looked through both the [open](https://github.com/zizmorcore/zizmor/issues?q=is%3Aissue+is%3Aopen+) and [closed](https://github.com/zizmorcore/zizmor/issues?q=is%3Aissue%20state%3Aclosed) issues for a duplicate request.

### What's the problem this feature will solve?

Dependabot now documents additional `schedule.interval` values (`quarterly`, `semiannually`, `yearly`, `cron`), but `github-actions-models::dependabot::v2::Interval` still only exposes `daily/weekly/monthly`. As a result, zizmor cannot parse repositories whose `dependabot.yml` uses the newer intervals or a cron expression, preventing audits on modern configurations.

### Describe the solution you'd like

Extend the `Interval` enum to include the new documented variants (`quarterly`, `semiannually`, `yearly`, `cron`).

### Additional context

GitHub documentation: https://docs.github.com/en/code-security/dependabot/working-with-dependabot/dependabot-options-reference#interval

woodruffw:
Hi @kingsword09, thanks for the report! I'll fix this today.

woodruffw:
(In the future, please file this with the bug template instead of the issue template -- it makes it easier for me to triage which versions have a bug in them.)

PR #1247, opened 2025-10-14
feat: Add validation for extended Dependabot schedule intervals

kingsword09:
Closes: #1246

- Extend the `Interval` enum to include the new documented variants (`quarterly`, `semiannually`, `yearly`, `cron`).
- Tighten `Schedule` deserialization to mirror the official schema requirements:
   - `interval = cron` must provide `cronjob`;
   - non-cron intervals must not set `cronjob`;
   - `day` is only valid for `weekly`.
- Add positive/negative YAML fixtures (mark invalid ones via `.invalid.yml`) and corresponding integration tests so these constraints remain covered in future changes.

### DIFF
--- a/docs/release-notes.md
+++ b/docs/release-notes.md
@@ -9,6 +9,13 @@ of `zizmor`.
 
 ## Next (UNRELEASED)
 
+## 1.15.2
+
+### Bug Fixes 🐛
+
+* Fixed a bug where `zizmor` would fail to parse some Dependabot configuration
+  files due to missing support for some schedule formats (#1247)
+
 ## 1.15.1
 
 ### Bug Fixes 🐛


kingsword09:
cc @woodruffw

crates/github-actions-models/tests/sample-dependabot/v2/cron-missing-cronjob.invalid.yml
woodruffw, line 1: @kingsword09 Could you remove these invalid test cases here? These "sample" directories in the models crate were only intended for valid cases; invalid cases should have their own dedicated tests.

(But in this case I'm OK with only having valid tests.)

woodruffw: Actually never mind, I guess it's okay. Thanks!

woodruffw:
This introduced an overly conservative check (the check is correct, but GitHub is permissive and so we have to be too); xref #1305 and #1308

Issue #1251, opened 2025-10-14
[BUG]: `use-trusted-publishing` complains about `cargo publish --dry-run` w/o authentication token


Holzhaus:
### Pre-submission checks

- [x] I am **not** filing a feature request. These should be filed via the feature request form instead.
- [x] I have looked through both the [open](https://github.com/zizmorcore/zizmor/issues?q=is%3Aissue+is%3Aopen+) and [closed](https://github.com/zizmorcore/zizmor/issues?q=is%3Aissue%20state%3Aclosed) issues for a duplicate report.

### zizmor version

1.15.2

### Expected behavior

When scanning a workflow containing the following step, zizmor should not raise the `use-trusted-publishing` audit.

```yaml
- name: Build Package and Check Publishing
  run: cargo publish --dry-run --locked --no-default-features --features "${CRATE_FEATURES}"
```

### Actual behavior

zizmor complains that trusted publishing should be used for authentication, even though no authentication is used at all because `cargo publish` is in `--dry-run` mode.

### Reproduction steps

Add the following YAML code to your workflow and lint it with zizmor:

```yaml
- name: Build Package and Check Publishing
  run: cargo publish --dry-run --locked --no-default-features --features "${CRATE_FEATURES}"
```

### Logs

```plain text
info[use-trusted-publishing]: prefer trusted publishing for authentication
  --> .github/workflows/build.yml:41:20
   |
41 |               run: cargo publish --dry-run --locked --no-default-features --features "${CRATE_FEATURES}"
   |               ---  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ this command
   |               |
   |               this step
   |
   = note: audit confidence → High

4 findings (3 suppressed): 1 informational, 0 low, 0 medium, 0 high
```

### Additional context

See https://doc.rust-lang.org/cargo/reference/publishing.html#packaging-a-crate for details on the `--dry-run` flag:

> ### Packaging a crate
>
> The next step is to package up your crate and upload it to [crates.io]. For
> this we’ll use the [`cargo publish`] subcommand. This command performs the following
> steps:
>
> 1. Perform some verification checks on your package.
> 2. Compress your source code into a `.crate` file.
> 3. Extract the `.crate` file into a temporary directory and verify that it
>    compiles.
> 4. Upload the `.crate` file to [crates.io].
[...]

PR #1253, opened 2025-10-15
fix: check for dry-run flags in use-trusted-publishing

woodruffw:
This is a tentative fix for #1251. However, I don't like it -- it's shell specific (needs `shlex` for bash, etc.) and essentially duplicates a lot of the work that tree-sitter can do under the hood.

Putting this up so I can think about it more.


woodruffw:
Not happy with this.

PR #1254, opened 2025-10-15
build: don't use jemalloc on any Windows-family targets

woodruffw:
The previous jemalloc check treated `msvc` as a proxy for Windows, but there are non-MSVC Windows targets like MinGW. This ensures that we never use the jemalloc allocator on Windows targets, regardless of compiler toolchain.

See: https://github.com/zizmorcore/zizmor/pull/1200#issuecomment-3404150574

+CC @striezel to sanity check this, since non-MSVC targets aren't currently included in the official build platforms for zizmor.

### DIFF
--- a/crates/zizmor/Cargo.toml
+++ b/crates/zizmor/Cargo.toml
@@ -71,7 +71,7 @@ tree-sitter-powershell.workspace = true
 yamlpath.workspace = true
 yamlpatch.workspace = true
 
-[target.'cfg(not(target_env = "msvc"))'.dependencies]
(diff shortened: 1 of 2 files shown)


woodruffw:
I guess `target_os` would also work here, and might be more idiomatic.

striezel:
> I guess `target_os` would also work here, and might be more idiomatic.

Probably, but I just ran builds with `target_family = "windows"` on MSYS2 (for ucrt64, mingw64, clang64 and clangarm64) and that works as it is now. So it's fine with me. :+1:

PR #1263, opened 2025-10-19
feat: track anchors in yamlpath

woodruffw:
This is an initial step towards supporting YAML anchors. The basic idea here is that a `yamlpath::Document` now tracks anchors with an anchor map, which we eagerly compute when the document is loaded. This map is a mapping of `anchor-name -> node`, which we'll then use to "teleport" around in pathing operations.

This also adds a new small support crate, `tree-sitter-iter`, to provide an iterator abstraction over `TreeCursor` from `tree-sitter`.

One intentional limitation here is that we currently require anchors to be unique: this isn't actually what YAML allows (...why?), but unique anchors are probably common enough in practice for this to be an acceptable constraint for now.

More on that here: https://til.agile.esm.co.jp/posts/yaml-anchors-do-not-have-to-be-unique/

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3173,7 +3173,7 @@ dependencies = [
 
 [[package]]
 name = "tree-sitter-iter"
-version = "0.0.1"
+version = "0.0.2"
 dependencies = [
  "tree-sitter",
  "tree-sitter-yaml",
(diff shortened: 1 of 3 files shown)


woodruffw:
Noting for myself: I can support multiple anchors with the same name by making it into a map of `name -> [node]`, and then finding the anchor node that's "closest" to the reference node. This will work better than trying to walk back up the tree to find the "most recent" anchor node (per the spec language), since that node could be in some arbitrary adjacent child of the CST.

woodruffw:
@kusari-inspector feedback this produced a false positive -- it can't tell that `tree-sitter-iter` is a workspace package, and that the missing version (0.0.2) is only missing because I was just about to publish it. I'm kind of confused as to why only the bump produced a false positive versus when I introduced the crate, though.

Issue #1269, opened 2025-10-21
Fix wheel builds


woodruffw:
These have been failing since #1257, which (re)introduced an OpenSSL dependency via `git2 -> openssl-sys`. 

https://github.com/rust-lang/git2-rs/issues/623 tracks the status of non-OpenSSL transport in `git2-rs`.

woodruffw:
I guess I could switch to gitoxide here, but it looks pretty hefty. It might be simpler to just make the ~~dumb~~ smart HTTP git connection for remote listing directly.

woodruffw:
This works well enough:

```
curl 'https://github.com/zizmorcore/zizmor.git/info/refs?service=git-upload-pack'
```

https://git-scm.com/docs/http-protocol

https://git-scm.com/docs/pack-protocol

https://git-scm.com/docs/protocol-common

woodruffw:
The v2 protocol might also be faster, although it seems to involve an additional round trip:

https://git-scm.com/docs/protocol-v2/2.21.0

```
curl 'https://github.com/zizmorcore/zizmor.git/info/refs?service=git-upload-pack'   -H 'Git-Protocol: version=2'

curl 'https://github.com/zizmorcore/zizmor.git/git-upload-pack' -d $'0014command=ls-refs\n0000' -H 'Git-Protocol: version=2'
```

I guess that first request technically isn't necessary.

Edit: with v2 we can also use the `ref-prefix <prefix>` option to filter down to just branch and tag refs, so that we don't have to pull down every GitHub-specific pull ref as well.

Bo98:
As a quick fix, `pyca/cryptography-manylinux*` images could work.

But protocol v2 seems decent for filtering - libgit2 still uses protocol v0. While the full gitoxide is hefty, `gix-packetline` might be a lightweight enough encoder/decoder for the git wire protocol and we do the actual transport ourselves.

The double roundtrip is mainly for backwards compatibility and for feature checks. The first response uses the same URL back to v0 so you can determine what the server supports based on it. If we limit it to just github.com and assume that GitHub will always support version 2 for at least as long as we care (since compatibility might not extend to the second request) then it's a worthwhile assumption

Bo98:
> If we limit it to just github.com

I assume GHES too given how long v2 has been a thing on github.com.

Speaking of, an error I may have made is neglecting GHES in the libgit2 implementation. I'm not familiar with how imported actions work in GHES. Maybe it's just as simple as github.com -> ghes host

[...]

PR #1273, opened 2025-10-22
feat: begin replacing git2 with direct remote listing

woodruffw:
See #1269.

### DIFF
--- a/docs/release-notes.md
+++ b/docs/release-notes.md
@@ -18,7 +18,7 @@ of `zizmor`.
 
 ### Performance Improvements 🚄
 
-* `zizmor`'s online mode is now significantly (40% to 90%) faster on
+* `zizmor`'s online mode is now significantly (40% to over 95%) faster on
   common workloads, thanks to a combination of caching improvements and
   conversion of GitHub API requests into Git remote lookups (#1257)
 


woodruffw:
In local testing this is still >95% faster on the hot cache path than 1.15.2. Which is egregious -- I haven't tested it yet, but I'm wondering if I introduced a significant network performance regression in the last few versions. Or maybe it's just the slow accretion of online audits.

woodruffw:
+CC @Bo98, I figure you might have opinions about the approach here 🙂 (I agree with your point about not needing to sniff for v2, FWICT both GitHub.com and GHES have supported it for years).

+CC @geofft as well, if I can snipe you into a review 😉

woodruffw:
On `main` with git2:

```
% GH_TOKEN=$(gh auth token) hyperfine -i -- './target/release/zizmor python/cpython'
Benchmark 1: ./target/release/zizmor python/cpython
  Time (mean ± σ):      4.927 s ±  1.261 s    [User: 0.699 s, System: 0.756 s]
  Range (min … max):    3.820 s …  6.776 s    10 runs
```

Here:

```
% GH_TOKEN=$(gh auth token) hyperfine -i -- './target/release/zizmor python/cpython'
Benchmark 1: ./target/release/zizmor python/cpython
  Time (mean ± σ):      3.347 s ±  0.871 s    [User: 0.700 s, System: 0.739 s]
  Range (min … max):    2.985 s …  5.818 s    10 runs
```

So that's a nice boost even over git2. I think that's probably from the v0 vs v2 listing for the most part, plus maybe connection reuse now that we're connecting through our reqwest pool.

crates/zizmor/src/github.rs
```diff
@@ -234,27 +234,27 @@ impl Client {
         token: &GitHubToken,
         cache_dir: &Utf8Path,
     ) -> Result<Self, ClientError> {
-        let mut headers = HeaderMap::new();
-        headers.insert(USER_AGENT, "zizmor".parse().unwrap());
-        headers.insert(AUTHORIZATION, token.to_header_value()?);
-        headers.insert("X-GitHub-Api-Version", "2022-11-28".parse().unwrap());
-        headers.insert(ACCEPT, "application/vnd.github+json".parse().unwrap());
+        // Base HTTP client for non-API requests, e.g. direct Git access.
+        let base_client = Self::default_middleware(
```
Bo98: Given this is a POST request, file-based caching will never kick in. Memory-based HTTP cache could kick in but we would need to change `ForceCache` to `IgnoreRules`. Doing so would work, though given I don't think it takes the body into consideration it would break if we ever do a second request with different parameters.

crates/zizmor/src/github.rs
```diff
@@ -279,73 +279,105 @@ impl Client {
...
-        let remote_refs: Vec<_> = tokio::task::spawn_blocking(move || {
-            let mut remote = Remote::create_detached(url)?;
+    async fn list_refs(&self, owner: &str, repo: &str) -> Result<Vec<RemoteHead>, ClientError> {
+        let url = format!("https://github.com/{owner}/{repo}.git/git-upload-pack");
 
-            let mut callbacks = RemoteCallbacks::new();
-            callbacks.credentials(|_url, _username_from_url, _allowed_types| token.to_cred());
+        if let Some(cached_entry) = self.ref_cache.get(&url).await {
+            return Ok(cached_entry);
+        }
```
Bo98: Linked to the previous comment, this obsoletes the HTTP cache entirely and is safer given we'd just roll a separate cache for separate requests (and also faster given we cache the parsing here too).

woodruffw: Ah yeah, good catch -- I factored all of the middleware building out into a little helper without thinking too hard about it.

woodruffw:
With `ref-prefix` filtering:

```
Benchmark 1: ./target/release/zizmor python/cpython
  Time (mean ± σ):      2.725 s ±  0.669 s    [User: 0.635 s, System: 0.714 s]
  Range (min … max):    2.457 s …  4.622 s    10 runs
```

I need to re-run without it though, with proper warmups.

Less noise, and warmups, on https://github.com/zizmorcore/zizmor/pull/1273/commits/2c4963dc5f04389a4b7a2a02b69db8bbb2affa68:

```
  Time (mean ± σ):      2.473 s ±  0.031 s    [User: 0.628 s, System: 0.678 s]
  Range (min … max):    2.417 s …  2.520 s    10 runs
```

woodruffw:
Without `ref-prefix` filtering: slightly slower, but I don't think the differences here are significant/escape the noise floor of my laptop:

```
  Time (mean ± σ):      2.879 s ±  0.040 s    [User: 0.678 s, System: 0.716 s]
  Range (min … max):    2.832 s …  2.947 s    10 runs
```

geofft:
The git-parsing code seems fine to me. I think you can avoid using regex but it's fine as is.

Do you want to send an `agent` capability in addition to the HTTP user-agent? The git CLI seems to send both.

geofft:
Also I was going to say you could extend your `list_refs` API to take a prefix, but the change you made to list both heads and tags avoids listing pulls which is the major benefit.

woodruffw:
> The git-parsing code seems fine to me. I think you can avoid using regex but it's fine as is.

Yeah, I was thinking I could hand-write this (or use `nom`) but I got lazy. It's not a very complicated pattern regardless, so I think `regex` will perform about as well as anything I handwrite 😅 

> Do you want to send an `agent` capability in addition to the HTTP user-agent? The git CLI seems to send both.

Ah yeah, good call. I should probably add that.

(Right now I'm not sniffing the server to make sure it actually advertises `agent` support, but given that GitHub is the only intended target here that's probably fine...)

PR #1323, opened 2025-11-07
Fix trailing newline bug

### LINKED ISSUE #1315: [BUG]: zizmor --fix=safe does not preserve trailing newlines when adding cooldown config to dependabot config
### Pre-submission checks

### zizmor version

zizmor 1.16.2

### Expected behavior

when running `zizmor --fix=safe .` with a dependabot config that ends in a trailing newline (as vim and prettier would add and yamllint requires), adding cooldown config at the end of an ecosystem stanza would preserve that trailing newline.

### Actual behavior

```diff
commit 05921a95a6f07981fa32537c69f8bed8a4cba9c4
Author: Ryan Lopopolo <rjl@hyperbo.la>
Date:   Sun Nov 2 19:27:48 2025 -0800

    zizmor

diff --git a/.github/dependabot.yml b/.github/dependabot.yml
index daa8456..e3d11eb 100644
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -13,6 +13,8 @@ updates:
       - lopopolo
     labels:
       - A-deps
+    cooldown:
+      default-days: 7
   - package-ecosystem: npm
     directory: "/"
     schedule:
@@ -28,6 +30,8 @@ updates:
     ignore:
       - dependency-name: "marked"
         update-types: ["version-update:semver-major"]
+    cooldown:
+      default-days: 7
   - package-ecosystem: bundler
     directory: "/"
     schedule:
@@ -40,3 +44,5 @@ updates:
       - lopopolo
     labels:
       - A-deps
+    cooldown:
+      default-days: 7
\ No newline at end of file
```

### Reproduction steps

zizmor --fix=safe .

### Logs

```plain text

```

### Additional context

_No response_

mostafa:
## Pre-submission checks

Please check these boxes:

      one first).

## Summary

Fixed trailing newline removal when adding cooldown config to dependabot files by preserving the document's trailing newline in yamlpatch's `Op::Add` handler if the original document ended with one. This fixes #1315.

## Test Plan

Added a test case to the yamlpatch unit-tests.

### DIFF
--- a/docs/release-notes.md
+++ b/docs/release-notes.md
@@ -20,15 +20,17 @@ of `zizmor`.
   and I/O-heavy audits significantly faster. Typical user workloads
   should see speedups of 40% to 70% (#1314)
 
+### Bug Fixes 🐛
+
+* Fixed a bug where auto-fixes would fail to preserve a document's final
+  newline (#1323)
+
 ## 1.16.3
 
 ### Bug Fixes 🐛
 
 * Fixed a bug where `zizmor` would crash on an unexpected caching middleware
   state. `zizmor` will now exit with a controlled error instead (#1319)
-
-* Fixed a bug where auto-fixes would fail to preserve a document's final
-  newline (#1323)
 
 ## 1.16.2
 


crates/yamlpatch/src/lib.rs
```diff
@@ -354,6 +354,13 @@ fn apply_single_patch(
                 feature.location.byte_span.0..feature.location.byte_span.1,
                 &updated_feature,
             );
+
+            // Preserve trailing newline if original document had one and result doesn't
+            // This handles the case where adding content at the end removes the trailing newline
+            if content.ends_with('\n') && !result.ends_with('\n') {
+                result.push('\n');
+            }
```
woodruffw: Huh, I'm kind of confused as to _how_ this happens -- my understanding of tree-sitter-yaml is that our parse is "greedy" i.e. always includes the trailing newline, so I don't understand how we end up losing it.

(Also, do we need to do this newline check more generally, or only for the `Add` op?)

mostafa: Tree-sitter's feature byte span (`node.start_byte()` to `node.end_byte()`) ends at the feature's content, so a trailing newline after the feature is outside the span. Replacing only that span drops the trailing newline if the replacement doesn't include it. I'll update the code to fix it in all places along with more tests.

mostafa: I tried to reimplement the fix in a better way and fix it for `Replace`, `ReplaceComment`, `RewriteFragment` and `Add` operations. Also, I added more unit-tests.

crates/yamlpatch/src/lib.rs
```diff
@@ -209,6 +209,43 @@ pub fn apply_yaml_patches(
...
+    document: &yamlpath::Document,
+) -> std::ops::Range<usize> {
+    let start = byte_span.0;
+    let mut end = byte_span.1;
+    let content = document.source();
+
+    // Only include trailing newline if it's the last character in the document.
+    // This preserves document-level trailing newlines without interfering with
+    // newlines that are part of the document structure.
+    if end < content.len() && content.as_bytes()[end] == b'\n' && end + 1 == content.len() {
```
woodruffw: I might be underthinking this, but I think this is shorter and equivalent:

```suggestion
    if content.as_bytes()[end] == b'\n' && end + 1 == content.len() {
```

Since `end + 1 == content.len()` implies `env < content.len()` 🙂

crates/yamlpatch/src/lib.rs
```diff
@@ -244,10 +273,8 @@ fn apply_single_patch(
 
             // Finally, put our patch back into the overall content.
             let mut patched_content = content.to_string();
-            patched_content.replace_range(
-                feature.location.byte_span.0..feature.location.byte_span.1,
-                &patched_feature,
-            );
+            let replacement_range = compute_replacement_range(&feature, document);
+            patched_content.replace_range(replacement_range, &patched_feature);
```
woodruffw, line 277: Not a blocker, just thinking out loud: it's not super ideal that we need these helpers now, since I could easily see us forgetting to call them in relevant places in the future. We might want to do more type state encoding here, i.e. have a `RawSpan` and `AdjustedSpan` or whatever and only allow replacements on ranges with the latter.

(That isn't necessarily the right approach either, just something that might help.)

woodruffw:
Thanks @mostafa!

PR #1347, opened 2025-11-18
Check for `timeout-minutes` in jobs and steps

### LINKED ISSUE #1023: Feature: Require or recommend the `timeout-minutes` property for all jobs

joewallwork:
## Pre-submission checks

Please check these boxes:

      one first).

## Summary

Fixes #1023
Towards #1225

This PR builds on the work of @EmmanuelBerkowicz in #1067 and gets the `timeout-minutes` check working.

I'm still quite new to Rust so I reimplemented it using what I did in #1227 as a template.

I've included checks for steps as well as jobs. There's an initial check whether any of the steps in a job set `timeout-minutes`. If so, we check the other steps to ensure they do, too. If not, we check if the job itself sets `timeout-minutes`.

## Test Plan

A test workflow is provided and the test data and snapshots are updated for the existing tests.


crates/zizmor/src/audit/timeout_minutes.rs
```diff
@@ -0,0 +1,86 @@
...
+    "missing timeout-minutes"
+);
+
+#[async_trait::async_trait]
+impl Audit for TimeoutMinutes {
+    fn new(_state: &AuditState) -> Result<Self, AuditLoadError> {
+        Ok(Self)
+    }
+
+    async fn audit_normal_job<'doc>(
```
woodruffw, line 27: The logic here seems pretty deeply nested 🙂 -- maybe we could do this in a "bubble up" style instead, i.e. use `audit_step` and then check the step's parent (the job) if a timeout is missing?

joewallwork: Good suggestion. I separated out `audit_step` in https://github.com/zizmorcore/zizmor/pull/1347/commits/f9a4803f12b6c9600068381babc96084af84d7f1 but the logic is still quite complex. I wanted to make sure that we don't get reports for both the job and the step when neither set `timeout-minutes`. Is there a simpler way?

woodruffw:
Closing in favor of #2166.

PR #1351, opened 2025-11-21
feat: downgrade artipacked severity for checkout@v6+

### LINKED ISSUE #1334: Feature: Update rules for persist-credentials in actions/checkout@v6
### Pre-submission checks

Starting with `actions/checkout@v6` (see [release notes](https://github.com/actions/checkout/releases/tag/v6-beta)), the behavior of `persist-credentials` will change:
credentials are now stored under `$RUNNER_TEMP` instead of the local Git config.
This significantly reduces the risk of accidentally leaking credentials through workflow artifacts, Docker builds, or uploaded files.

Zizmor currently recommends setting `persist-credentials: false` for all checkout actions to avoid credential leakage.
With this change, that recommendation might be less critical for workflows using `checkout@v6`, since the credentials are no longer stored inside the repository itself.

[*Quoted from GitHub Actions docs:*](https://docs.github.com/en/actions/reference/workflows-and-actions/variables)

> **RUNNER_TEMP**:
> The path to a temporary directory on the runner. This directory is emptied at the beginning and end of each job. Note that files will not be removed if the runner's user account does not have permission to delete them. For example, `D:\a\_temp`.

It might make sense to revisit Zizmor’s `artipacked` rule or documentation around `persist-credentials` to reflect this new behavior in `actions/checkout@v6`.

### Additional context

_No response_

ManuelLerchnerQC:
## Pre-submission checks

Please check these boxes:

      one first).

## Summary

Closes #1334

This PR downgrades the severity of `artipacked` findings for `actions/checkout@v6+`.

Starting with `actions/checkout@v6`, credentials are stored in `$RUNNER_TEMP` instead of `.git/config`, significantly reducing the risk of credential leakage. This allows us to lower the severity of such findings.

## Changes

- Added version detection for `actions/checkout@v6+` and severity downgrade logic
- Severity changes: `Medium` → `Low` (no uploads) and `High` → `Medium` (with uploads) for v6+

## Test Plan

- Added unit tests for version detection and severity determination

### DIFF
--- a/crates/zizmor/src/audit/artipacked.rs
+++ b/crates/zizmor/src/audit/artipacked.rs
@@ -5,7 +5,7 @@ use super::{Audit, AuditLoadError, audit_meta};
 use crate::{
     audit::AuditError,
     finding::{Confidence, Finding, Fix, Persona, Severity, location::Routable as _},
-    models::{StepBodyCommon, StepCommon, uses::RepositoryUsesExt as _},
+    models::{StepBodyCommon, StepCommon, uses::RepositoryUsesExt as _, version::Version},
     state::AuditState,
     utils::split_patterns,
 };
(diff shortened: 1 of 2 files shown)


docs/audits.md
```diff
@@ -81,6 +81,14 @@ including it in a publicly accessible artifact via @actions/upload-artifact.
 However, even without this, persisting the credential in the `.git/config`
 is non-ideal unless actually needed.
 
+!!! note "checkout@v6+ behavior change"
+
+    Starting with `actions/checkout@v6`, credentials are stored under `$RUNNER_TEMP`
```
ytausch, line 86: Maybe link the [release](https://github.com/actions/checkout/releases/tag/v6.0.0) notes?

crates/zizmor/src/audit/artipacked.rs
```diff
@@ -20,6 +20,52 @@ audit_meta!(
 );
 
 impl Artipacked {
+    /// Check if the checkout action is version 6 or higher.
+    /// Returns None if the version cannot be determined (e.g., for commit SHAs).
+    fn is_checkout_v6_or_higher(
+        uses: &github_actions_models::common::RepositoryUses,
+    ) -> Option<bool> {
+        // Only check symbolic refs (tags/branches), not commit SHAs
+        let Some(ref_str) = uses.symbolic_ref() else {
```
ytausch, line 29: Pinning actions to commit SHAs is actually the recommended way to do things, so I think this doesn't help us really

crates/zizmor/src/audit/artipacked.rs
```diff
@@ -20,6 +20,52 @@ audit_meta!(
...
+        };
+
+        // Try to parse the ref as a version
+        let Ok(version) = Version::parse(ref_str) else {
+            // If we can't parse it as a version, assume it's not v6+
+            return Some(false);
+        };
+
+        // Check if version is >= v6
+        let v6 = Version::parse("v6").ok()?;
```
ytausch, line 41:
```suggestion
        let v6 = Version::parse("v6").unwrap();
```

An error here would be an implementation error that we want to know about

ytausch (changes requested):

crates/zizmor/src/audit/artipacked.rs
woodruffw: Yeah, there's a bigger issue here: the easiest way to check this is the version tag, but zizmor explicitly encourages people (correctly IMO) to hash-pin their actions.

So, for full generality, what we actually need here is:

1. If it's a symbolic ref, perform version comparison directly
2. If it's a hash ref, we need perform an online ref lookup (we have a fast path for this in the GitHub client) and map it back to a version, or fall back to assuming `< 6` if we can't map it.

(2) requires online access, which would be a significant change for this audit (it's currently offline only). So what we'll probably want to do is progressively enhance: only perform the online ref lookup if zizmor is online, otherwise fall back to the less precise finding.

woodruffw:
https://github.com/zizmorcore/zizmor/pull/1351#discussion_r2550482176 is going to involve a larger change that I'll need to iterate on, so I'm going to branch off of this and start a new PR. Thanks for starting this @ManuelLerchnerQC!

(Your contributions will be preserved in the commit history + I'll credit your work in the changelog for this 🙂)

woodruffw:
Continuing with #1353.

PR #1353, opened 2025-11-23
feat: downgrade artipacked severity for checkout@v6+

### LINKED ISSUE #1334: Feature: Update rules for persist-credentials in actions/checkout@v6

woodruffw:
Closes #1334.

Continues #1351.

### DIFF
--- a/docs/release-notes.md
+++ b/docs/release-notes.md
@@ -13,6 +13,11 @@ of `zizmor`.
 
 * `zizmor` now produces a more useful error message when asked to
   collect only workflows from a remote input that contains no workflows (#1324)
+  
+* `zizmor` now produces more precise severities on @actions/checkout versions
+  that have more misuse-resistant credentials persistence behavior (#1353)
+  
+    Many thanks to @ManuelLerchnerQC for proposing and implementing this improvement!
 
 ### Performance Improvements 🚄
 


ManuelLerchnerQC:
Thanks a lot for picking this up and finishing it over the weekend! @woodruffw

PR #1354, opened 2025-11-23
docs: switch to zensical

woodruffw:
It seems to Just Work (tm), and doesn't cause weird libcairo linkage issues like mkdocs-material sometimes did.


woodruffw:
The only thing that seems to be currently missing is social cards, and I'm okay with not having those for now.

Issue #1363, opened 2025-11-25
Feature: adding new elements to existing YAML block sequences


omar-s-ta:
### Pre-submission checks

- [x] I am **not** reporting a bug (crash, false positive/negative, etc). These must be filed via the bug report template.
- [x] I have looked through both the [open](https://github.com/zizmorcore/zizmor/issues?q=is%3Aissue+is%3Aopen+) and [closed](https://github.com/zizmorcore/zizmor/issues?q=is%3Aissue%20state%3Aclosed) issues for a duplicate request.

### What's the problem this feature will solve?

The `yamlpatch` library currently lacks support for adding elements to YAML block sequences. Using `Op::Add` to insert a new mapping into a list results in the value being serialized as an unwanted single-line flow style, for example:
{ subsetName: lego-quarantine, quarantine: true, ... }. This breaks the expected multi-line block formatting that YAML uses for readability and structural clarity.

There is no operation for appending items to a block sequence, to manage list-based configurations.

When we were trying to use the library we wanted to be able to add block sequences with also preserving the structure and the comments.

### Describe the solution you'd like

Introduce an `Op::Append` operation to the `yamlpatch` library, to enable adding YAML block sequences.

woodruffw:
Thanks for opening an issue @omar-s-ta! I'll be able to give #1364 a review in the coming days.

Out of curiosity, did you have an audit in mind for this new operation? I couldn't think of a place off the top of my head where we append currently, but I might not be remembering one.

omar-s-ta:
Hello @woodruffw, thanks for the quick reply.
No, I do not have an audit in mind for this operation.
We are using the crate to automate some tasks by modifying our existing yaml files, and using the `Op::Add` creates a flow-style instead of block-style, and we didn't find another way to do it. That was the motivation behind creating the issue, hence the PR.

woodruffw:
> We are using the crate to automate some tasks by modifying our existing yaml files, and using the `Op::Add` creates a flow-style instead of block-style, and we didn't find another way to do it. That was the motivation behind creating the issue, hence the PR.

Okay, thanks for explaining. FWIW, this crate is an internal implementation detail of zizmor, so there are no strong stability/correctness guarantees with it -- I try to keep it stable for my own purposes, but it may change significantly between releases. I figure you're probably already aware of that, but just to make it explicit!

mostafa:
@woodruffw I created a tiny PR to update the README: #1389.

[...]

PR #1364, opened 2025-11-25
Support adding block sequence in `yamlpatch`

omar-s-ta:
## Pre-submission checks

Please check these boxes:

      one first).

Used AI assistant to create the tests.

## Summary

Towards: https://github.com/zizmorcore/zizmor/issues/1363

Add `Op::Append` operation to the `yamlpatch` library, to add elements to YAML block sequences. 

Implement `handle_block_sequence_append`. The function focuses only on applying the correct list-item formatting and indentation.

## Test Plan

Tests are add in `crates/yamlpatch/tests/unit_tests.rs`


crates/yamlpatch/tests/unit_tests.rs
```diff
@@ -2547,3 +2547,430 @@ key: value"#;
...
+    let result =
+        apply_yaml_patches(&yamlpath::Document::new(original).unwrap(), &operations).unwrap();
+
+    insta::assert_snapshot!(result.source(), @r"
+        foo:
+          - abc
+          - - def
+            - ghi
+        ");
+}
```
woodruffw, line 3184: @omar-s-ta I added this test to confirm my understanding, and I'm curious what you think about it -- it's valid YAML, but I think this kind of nested block sequence is a little less idiomatic than:

```yaml
foo:
  - abc
  - [def, ghi]
```

(Not necessarily a blocker, just something I wasn't sure about.)

omar-s-ta: I agree with you, I'd prefer to create a follow-up enhancement issue to improve this one, if that is okay with you. Let me know WYT.

woodruffw: Yep, a follow-up is fine. Thank you!

PR #1381, opened 2025-11-29
feat: improve forbidden-uses error message on invalid config

woodruffw:
Our previous use of `untagged` meant that we'd swallow our actual underlying error and replace it with a generic "no variants matched" error, which is not super useful to users.

We now employ a minor hack to get JSON style tagged enums to work with serge-yaml, which changed its behavior on 0.9 to default to YAML style explicit tagging (which we do not want).


woodruffw:
This is also probably slightly faster, since we no longer have to perform exhaustive enum matching. But the performance difference doesn't matter at all given that config loading isn't a hot path at all + there are only two variants here.

PR #1396, opened 2025-11-30
feat: improve error reporting for invalid YAML inputs

### LINKED ISSUE #1395: Poor user experience on invalid YAML inputs
### Pre-submission checks

### zizmor version

1.18.0

### Expected behavior

Zizmor should handle common syntax errors like duplicate keys more gracefully, for example by giving an error message that highlights the location of the syntax error.

### Actual behavior

Zizmor gives a generic "failed to load file as workflow" error with a message that suggests reporting it as a bug.

### Reproduction steps

Run zizmor on the following YAML file:
```yml
name: Demonstration of Zizmor bug
on: workflow_dispatch
jobs:
  demo:
    name: Demonstration
    runs-on: ubuntu-latest
    steps:
      - name: Duplicate env block breaks Zizmor
        env:
          FOO: foo
        run: echo ...
        env:
          BAR: bar
```

### Logs

```plain text
$ zizmor --verbose example.yml
🌈 zizmor v1.18.0
DEBUG collect_inputs: zizmor::config: discovering config for local input `example.yml`
DEBUG collect_inputs: zizmor::config: attempting config discovery in `.`
DEBUG collect_inputs: zizmor::config: reached filesystem root without finding a config
DEBUG collect_inputs: zizmor::registry::input: registering workflow input as with key file://example.yml
fatal: no audit was performed
failed to load file://example.yml as workflow

Caused by:
    0: couldn't turn input into a an appropriate model
    1: this suggests a bug in zizmor; please report it!
    2: jobs: data did not match any variant of untagged enum Job at line 4 column 3
```

### Additional context

_No response_

woodruffw:
We previously produced suboptimal error messages on a subset of invalid inputs, in part because we would deserialize into a `serde_json::Value` directly rather than going through `serde_yaml::Mapping` first. This meant that YAML-specific validity rules (like no duplicate keys in mappings) were not enforced, leading to confusing error messages from the JSON schema validator.

With the specific case in #1395:

```console
🌈 zizmor v1.18.0
 WARN collect_inputs: zizmor::registry::input: failed to parse input: jobs.demo.steps[0]: duplicate entry with key "env" at line 8 column 9
fatal: no audit was performed
no inputs collected
```

and with `--strict-collection`:

```console
🌈 zizmor v1.18.0
fatal: no audit was performed
failed to load file:///Users/william/personal/devel/zizmorcore/zizmor/crates/zizmor/tests/integration/test-data/issue-1394-repro.yml as workflow

Caused by:
    0: invalid YAML syntax: jobs.demo.steps[0]: duplicate entry with key "env" at line 8 column 9
    1: jobs.demo.steps[0]: duplicate entry with key "env" at line 8 column 9
```

Fixes #1395.

### DIFF
--- a/crates/zizmor/src/main.rs
+++ b/crates/zizmor/src/main.rs
@@ -955,7 +955,8 @@ async fn main() -> ExitCode {
 
                         Some(report)
                     }
-                    CollectionError::Yamlpath(..) => {
+                    // These errors only happen if something is wrong with zizmor itself.
+                    CollectionError::Yamlpath(..) | CollectionError::Model(..) => {
                         let group = Group::with_title(Level::ERROR.primary_title(err.to_string())).elements([
                             Level::HELP.message("this typically indicates a bug in zizmor; please report it"),
                             Level::HELP.message(
--- a/crates/zizmor/src/registry/input.rs
+++ b/crates/zizmor/src/registry/input.rs
@@ -34,7 +34,7 @@ pub(crate) enum CollectionError {
     /// The input couldn't be converted into the expected model.
     /// This typically indicates a bug in `github-actions-models`.
     #[error("couldn't turn input into a an appropriate model")]
-    Model(#[source] anyhow::Error),
+    Model(#[from] serde_yaml::Error),
 
     /// The input couldn't be loaded into an internal yamlpath document.
     /// This typically indicates a bug in `yamlpath`.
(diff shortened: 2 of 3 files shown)


crates/zizmor/src/main.rs
```diff
@@ -955,7 +955,8 @@ async fn main() -> ExitCode {
 
                         Some(report)
                     }
-                    CollectionError::Yamlpath(..) => {
+                    // These errors only happen if something is wrong with zizmor itself.
+                    CollectionError::Yamlpath(..) | CollectionError::Model(..) => {
```
woodruffw, line 959: Noting: this turns the somewhat confusing render in #1395 into a nice tip render, but I don't have a great way to test it (since these kinds of errors are exceptional by nature/indicate bugs that I try to fix).

PR #1397, opened 2025-11-30
chore: bump zensical, site deps

woodruffw:
This brings us to the latest zensical release. I had to add `--clean` to the build to prevent a broken local site build (seemingly due to caching of the old release's build).

See: https://github.com/zensical/zensical/issues/149

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,7 +4,7 @@ all:
 
 .PHONY: site
 site:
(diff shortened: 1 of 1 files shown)


woodruffw:
Also: https://github.com/zensical/zensical/issues/103

PR #1410, opened 2025-12-04
docs: trigger only if there are changes to the workflow files

yeikel:
Please check these boxes:

      one first).

This is my first contribution to the project. It is unclear if documentation changes also need an issue, but I'd be happy to create if it is mandatory for every change regardless of type.

Also, if this change is accepted, I'd be happy to send a similar change to https://github.com/zizmorcore/zizmor-action/blob/main/README.md

Thank you for reviewing my change!  

## Summary

Updates the GitHub Action workflow example so that it only triggers if there are changes to the workflow files. 

## Test Plan

`paths` is a known GitHub action feature: https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#onpushpull_requestpull_request_targetpathspaths-ignore

### DIFF
--- a/docs/integrations.md
+++ b/docs/integrations.md
@@ -17,9 +17,12 @@ name: GitHub Actions Security Analysis with zizmor 🌈
 
 on:
   push:
-    branches: ["main"]
+    branches: main
+    paths:
+      - ".github/workflows/**"
   pull_request:
-    branches: ["**"]
+    paths:
+      - ".github/workflows/**"
 
 permissions: {}
 
@@ -79,8 +82,11 @@ GitHub Actions setup:
     on:
       push:
         branches: ["main"]
+        paths:
+          - ".github/workflows/**"
       pull_request:
-        branches: ["**"]
+        paths:
+          - ".github/workflows/**"
 
     permissions: {}
 


docs/integrations.md
```diff
@@ -17,9 +17,12 @@ name: GitHub Actions Security Analysis with zizmor 🌈
 
 on:
   push:
-    branches: ["main"]
+    branches: main
+    paths:
+      - ".github/workflows/**"
   pull_request:
-    branches: ["**"]
```
yeikel, line 22: I removed this because this is the default behavior. I can revert it if this was intentional

woodruffw:
Hi @yeikel, thanks for the PR.

Unfortunately I don't think this can be the blanket filter: zizmor also audits action definitions and Dependabot configs, which wouldn't appear under `.github/workflows/`. The latter is easy since they have a static pattern but the former can appear anywhere in the Git repository, meaning that we can't define a generic pattern for them that'll work for all users.

(There's another edge case, which is template-style repositories with multiple interior "repository" layouts. zizmor also supports these!)


As a result, my concern with any pattern we'd suggest here is that it'll end up accidentally excluding files that *should* be analyzed by zizmor. 

Instead of changing the default, maybe we can add a tip or other callout explaining to the user that they can add a path filter _if_ they're confident that it's correct for their particular repo layout?

woodruffw:
There actually might be a way to do a generic `action.yml` match, now that I think of it: `**/action.ya?ml` might work.

However I'm kind of hesitant to use that because it still means that anybody who configures zizmor at a point in time wouldn't get access to new input support automatically -- for example we added Dependabot support pretty recently, so if we recommended a pattern prior to that then a lot of users would never have picked those changes up.

yeikel:
I see, thank you for the feedback.

I wasn’t aware that the scan covered more than just the workflow folders. The additional context was very helpful

> However I'm kind of hesitant to use that because it still means that anybody who configures zizmor at a point in time wouldn't get access to new input support automatically 

With this in mind, the potential risks seem to outweigh the benefits and may lead to more long-term problems than the CI time savings justify

In fact, we may want to scan for this proposal and consider it an anti-pattern given the note above! 😄

woodruffw:
Yeah, it's kind of unfortunate that there's no way to have zizmor and the workflow path patterns "communicate" more tightly -- some tools solve that problem by having a codegen option that generates the workflow itself, but in my experience users never actually regenerate it (and get new changes) once it appears to work.

yeikel:
> Yeah, it's kind of unfortunate that there's no way to have zizmor and the workflow path patterns "communicate" more tightly -- some tools solve that problem by having a codegen option that generates the workflow itself, but in my experience users never actually regenerate it (and get new changes) once it appears to work.

Yeah, I agree with this sentiment. I think that unless the action fails if the generation is behind, users will simply choose not to generate it which would cause issues as you suggested. 

Is [caching](https://github.com/actions/cache) something we can consider instead? 

I'd rather not make this more complicated than it needs to be. I am just thinking out loud assuming that it is not too complicated to determine a caching key

Issue #1415, opened 2025-12-07
Feature: Clarify why zizmor itself needs `actions: read` permission


johnbillion:
### Pre-submission checks

- [x] I am **not** reporting a bug (crash, false positive/negative, etc). These must be filed via the bug report template.
- [x] I have looked through both the [open](https://github.com/zizmorcore/zizmor/issues?q=is%3Aissue+is%3Aopen+) and [closed](https://github.com/zizmorcore/zizmor/issues?q=is%3Aissue%20state%3Aclosed) issues for a duplicate request.

### What's the problem this feature will solve?

I'm auditing the permissions in my workflow files and I haven't been able to identify why the `actions: read` permission is required by the `GH_TOKEN` that zizmor itself uses for its checks when it runs on GitHub Actions.

I looked at the list of audit rules that don't work offline to see if I could figure out which one requires it, but with no luck.

Candidates:

- `impostor-commit`
- `known-vulnerable-actions`
- `ref-confusion`
- `stale-action-refs`

Previous issue: #608

### Describe the solution you'd like

Could you confirm why this permission is required by zizmor and which audit rule uses it?

### Additional context

Big fan of zizmor!

woodruffw:
Hey @johnbillion, thanks for opening an issue (and for the kind words!)

Yeah, this is a source of confusion: zizmor *itself* doesn't need `actions: read`, but the integrations that `zizmor-action` uses by default (namely SARIF uploads) needs it (or at least, GitHub's own docs claim that it's needed).

For reference, here's GitHub's own docs on SARIF uploads:

https://docs.github.com/en/code-security/code-scanning/integrating-with-code-scanning/uploading-a-sarif-file-to-github

The actual permissions needed by zizmor *itself* are documented here: https://docs.zizmor.sh/usage/#github-api-token-permissions

TL;DR: None of the audits need that permission, and if you're not using the SARIF integration (which is enabled by default) you can remove it. It's documented by default because most people use it (at least I think they do 😅), but I'm open to suggestions on how to more clearly document/distinguish between permissions zizmor needs and permissions that zizmor-action needs for various integration points.

woodruffw:
#1423 addressed this, I believe 🙂

PR #1423, opened 2025-12-08
Update permissions documentation

johnbillion:
## Pre-submission checks

Please check these boxes:

      one first).

## Summary

See #1415.

This updates inline docs to clarify why each permission is required, and removes what I believe to be an unnecessary `actions: read` permission from the documentation on using GitHub annotations instead of code scanning. Would appreciate a sanity check on the latter.

### DIFF
--- a/docs/integrations.md
+++ b/docs/integrations.md
@@ -162,7 +162,6 @@ GitHub Actions setup:
         runs-on: ubuntu-latest
         permissions:
           contents: read # Only needed for private repos. Needed to clone the repo.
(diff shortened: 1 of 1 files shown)


johnbillion:
If you're interested, [this appears to be the API call](https://github.com/github/codeql-action/blob/149d184a5153ea45e6fbcef5588ac7b8c7af9835/src/api-client.ts#L159-L168) in upload-sarif which needs the `actions: read` permission.

woodruffw:
Thanks @johnbillion!

Issue #1138, opened 2025-09-04
Feature: Publish the schema of zizmor.yml to https://www.schemastore.org


vivodi:
### Pre-submission checks

- [x] I am **not** reporting a bug (crash, false positive/negative, etc). These must be filed via the bug report template.
- [x] I have looked through both the [open](https://github.com/zizmorcore/zizmor/issues?q=is%3Aissue+is%3Aopen+) and [closed](https://github.com/zizmorcore/zizmor/issues?q=is%3Aissue%20state%3Aclosed) issues for a duplicate request.

### What's the problem this feature will solve?

Sometimes I can’t be sure whether the zizmor.yml I wrote is valid.

### Describe the solution you'd like

Publish the schema of zizmor.yml to https://www.schemastore.org so that other tools can check whether zizmor.yml contains any errors.

That way, the pre-commit hook provided by https://github.com/python-jsonschema/check-jsonschema can check the validity of zizmor.yml.

woodruffw:
Hi @vivodi, thanks for opening an issue!

I agree that a schema for `zizmor.yml` should be published to SchemaStore. However, it's worth noting that a schema for `zizmor.yml` doesn't even *exist* yet, so that would need to be added first 🙂 

TL;DR: If someone were to contribute a JSON schema for `zizmor.yml` to this repo, I'd happily review it and then submit it upstream to SchemaStore. Otherwise I'll likely get to this in the medium term.

woodruffw:
Misc notes:

* SchemaStore's contributing guide: https://github.com/SchemaStore/schemastore/blob/master/CONTRIBUTING.md#recommended-extensions
* It seems like VSCode's YAML LSP doesn't actually pick up *new* schemas in SchemaStore by default; they need to be explicitly enabled on VSCode's side: https://github.com/microsoft/vscode/issues/26289
* It might also make sense for `zizmor-lsp` to ship this schema directly, although the LSP itself doesn't (yet) honor configs

woodruffw:
Also noting: we'll probably want to use schemars for this: https://docs.rs/schemars/latest/schemars/index.html

woodruffw:
Also noting: one of the things that makes this annoying/hard in zizmor's context is that each audit's config is "polymorphic," i.e. not directly described by the types within `RawConfig` itself.

kiwamizamurai:
@woodruffw 

Hello, I'm kiwamizmaurai from Japan. nice to meet you.

[...]

PR #1467, opened 2025-12-24
feat: Add JSON Schema generation for zizmor.yml using schemars

kiwamizamurai:
## Pre-submission checks

Please check these boxes:
      one first).
		https://github.com/zizmorcore/zizmor/issues/1138

## Summary

This PR implements JSON Schema generation for `zizmor.yml` configuration files using the `schemars` crate, as suggested in #1138

## Test Plan

```bash
# Build with schema feature
cargo build --features schema
# Build without schema feature (ensure no regression)
cargo build
# Generate and verify schema
cargo run --features schema -- --generate-schema | python3 -c "import json, sys; json.load(sys.stdin); print('Valid JSON')"
```

and VscodeIDE

### DIFF
--- a/support/zizmor.schema.json
+++ b/support/zizmor.schema.json
@@ -41,7 +41,7 @@
           "default": 7,
           "type": "integer",
           "format": "uint64",
-          "minimum": 1
+          "minimum": 1.0
         }
       },
       "additionalProperties": false


woodruffw:
Thanks @kiwamizamurai, I'll try and give this a review later today!

.github/workflows/ci.yml
```diff
@@ -76,10 +76,37 @@ jobs:
...
+            exit 1
+          fi
+          echo "Schema covers all $REGISTRY_COUNT audits"
+
+      - uses: Swatinem/rust-cache@779680da715d629ac1d338a641029a2f4372abb5 # v2.8.2
+
+      - name: Verify committed schema is up-to-date
+        run: |
+          cargo run --features schema -- --generate-schema > /tmp/generated-schema.json
+          diff -u crates/zizmor/src/data/zizmor.schema.json /tmp/generated-schema.json
```
woodruffw: Nitpick: maybe we could add this as a Makefile target like the other support tasks, e.g. `make generate-schema`? We could then do `git diff --exit-code` to use git's own diff without an extra tempfile 🙂

crates/zizmor/src/data/zizmor.schema.json
woodruffw, line 1: Can we generate this somewhere else, maybe under `support/`? I think that would be consistent with where we put other kinds of generated assets.

(I think if put here, it'll be included in the crate distribution by default, which is _probably_ not what we want since it doesn't actually get used there.)

crates/zizmor/src/config/schema.rs
woodruffw, line 1: Nit: I think we should make this a sub-module of `config`, i.e. `config::schema` instead of an adjacent module. In other words, I think this should be under `config/schema.rs` 🙂

crates/zizmor/src/config_schema.rs
```diff
@@ -0,0 +1,202 @@
+use std::collections::HashMap;
+
+use schemars::JsonSchema;
+use serde::{Deserialize, Serialize};
+
+use crate::config::{UsesPolicy, WorkflowRule};
+use crate::models::uses::RepositoryUsesPattern;
+
+#[derive(Clone, Debug, Default, Serialize, Deserialize, JsonSchema)]
```
woodruffw: Q: is there a reason we need the `Serialize` trait on these types? FWICT we don't actually use that functionality.

crates/zizmor/src/config_schema.rs
```diff
@@ -0,0 +1,202 @@
...
+#[serde(deny_unknown_fields)]
+pub struct DependabotCooldownRuleConfig {
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub disable: Option<bool>,
+
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    #[schemars(schema_with = "workflow_rule_vec_schema")]
+    pub ignore: Option<Vec<String>>,
```
woodruffw: Q: Can we use serde's `flatten` here to dedupe this with `BaseRuleConfig`? I think that would be a little DRYer, but I'm not sure if schemars supports that 🙂

crates/zizmor/src/config_schema.rs
```diff
@@ -0,0 +1,202 @@
...
+
+#[derive(Clone, Debug, Default, Serialize, Deserialize, JsonSchema)]
+#[serde(deny_unknown_fields)]
+pub struct BaseRuleConfig {
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub disable: Option<bool>,
+
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    #[schemars(schema_with = "workflow_rule_vec_schema")]
+    pub ignore: Option<Vec<String>>,
```
woodruffw: Q: do these need to be `Option<...>`? I think we'd ideally re-use the type definitions in `config.rs`, which are non-optional but use `default` to ensure the right default value is populated.

woodruffw: (And if we can make these non-optional, then ideally we'd remove this type and reuse the ones in `config.rs`!)

.github/workflows/ci.yml
kiwamizamurai: Exactly! I've added a `make generate-schema` target and updated the CI to use `git diff --exit-code` instead of the tempfile approach. Much cleaner!

crates/zizmor/src/data/zizmor.schema.json
kiwamizamurai: Makes sense! Moved the schema to `support/zizmor.schema.json` to be consistent with other generated assets and avoid including it in the crate distribution.

crates/zizmor/src/config/schema.rs
kiwamizamurai: Agreed! Restructured as `config::schema` - moved `config.rs` to `config/mod.rs` and `config_schema.rs` to `config/schema.rs`.

crates/zizmor/src/config_schema.rs
kiwamizamurai: Exactly, `Serialize` isn't needed here - schemars doesn't require it for schema generation. Removed

But this caused `dead_code` warnings since the compiler can't see that schemars accesses these types via reflection. Added a module-level `#![allow(dead_code)]` with documentation explaining why these types appear unused but are actually accessed through schemars' procedural macros.

crates/zizmor/src/config_schema.rs
kiwamizamurai: Yes. schemars does support `flatten`, so I've refactored to use `#[serde(flatten)]` with `BaseRuleConfig`.

crates/zizmor/src/config_schema.rs
kiwamizamurai: I've looked into this - the challenge is that `config.rs` uses a dynamic `HashMap<String, AuditRuleConfig>` structure with generic `serde_yaml::Mapping` for audit-specific configs, while the schema needs static type definitions for each audit's configuration.

However, I've made two improvements:
1. The common types (`WorkflowRule`, `UsesPolicy`) in `config.rs` already derive `JsonSchema` conditionally
2. Changed the schema types to use `#[serde(flatten)]` with `BaseRuleConfig` to reduce duplication


If you have ideas on how to further reduce duplication while maintaining the strongly-typed schema, I'm happy to iterate!

woodruffw: Thanks, makes sense. A small amount of duplication seems acceptable to me here.

crates/zizmor/src/config_schema.rs
woodruffw: Oh huh, I'm kind of surprised it causes a dead code warning even though schemars accesses these. I can look into that. But thanks for removing the `Serialize` impls!

crates/zizmor/src/config/schema.rs
woodruffw: Thanks! I actually kind of prefer the `config.rs + config/schema.rs` pattern instead of `config/{mod.rs,schema.rs}`, but that can be a future refactor.

crates/zizmor/src/config_schema.rs
woodruffw: Thanks!

crates/zizmor/src/config_schema.rs
woodruffw: Oh, just realized -- even if we can't dedupe these types, can we remove the `Optional<...>` wrappers? I think that's actually somewhat misleading in terms of the generated schema, since it implies that `key: null` is valid when really we don't allow that.

For example:

```
    "DependabotCooldownConfig": {Add a comment on  line R43Add diff commentMarkdown input:  edit mode selected.WritePreviewAdd a suggestionHeadingBoldItalicQuoteCodeLinkUnordered listNumbered listTask listMentionReferenceSaved repliesAdd FilesPaste, drop, or click to add filesCancelCommentStart a reviewReturn to code
      "type": "object",
      "properties": {
        "days": {
          "default": null,
          "type": [
            "integer",
            "null"
          ],
          "format": "uint64",
          "minimum": 0.0
        }
      },
      "additionalProperties": false
    },
```

This implies that `days: null` is valid, but in reality that's an invalid config. If the user wants the default, they need to leave `days` unset 🙂

kiwamizamurai: Sorry I fixed it

[`34276aa` (#1467)](https://github.com/zizmorcore/zizmor/pull/1467/commits/34276aab1c2f4465b57d39871fbb5468bf67f92c)

crates/zizmor/src/config_schema.rs
kiwamizamurai: I investigated this a little bit. Here's what I found:

### Root Cause
The `dead_code` warning occurs because of how `JsonSchema` derive generates code vs `Serialize`:
**JsonSchema generates:**
```rust
schemars::_private::insert_object_property(
    &mut schema,
    "field1",                              // Field name as string literal
    generator.subschema_for::<String>(),   // Only TYPE information
);
```

**Serialize generates:**
```rust
serialize_field(&mut state, "field1", &self.field1)?;
//                                    ^^^^^^^^^^^
//                                    Actually reads the field value
```

JsonSchema only uses type reflection - it never accesses `self.field1`, so the compiler correctly identifies the fields as "never read". This is expected behavior, not a bug.

Other projects using `#![allow(dead_code)]` for the same reason (schema-only types without Serialize):

- [radicle-dev/heartwood](https://github.com/radicle-dev/heartwood/blob/main/crates/radicle/src/schemars_ext.rs) - `schemars_ext.rs`: JsonSchema-only types for remote derive pattern
- [modelcontextprotocol/rust-sdk](https://github.com/modelcontextprotocol/rust-sdk/blob/main/examples/servers/src/common/calculator.rs) - `calculator.rs`: Deserialize+JsonSchema (no Serialize)

crates/zizmor/src/config/mod.rs
```diff
@@ -640,47 +691,3 @@ impl Config {
...
-        );
-
-        assert_eq!(
-            WorkflowRule::from_str("foo.yml:123")?,
-            WorkflowRule {
-                filename: "foo.yml".into(),
-                line: Some(123),
-                column: None
-            }
-        );
```
woodruffw, line 668: Q: why were these tests deleted? These would be nice to keep I think.

support/zizmor.schema.json
```diff
@@ -0,0 +1,567 @@
...
+      "additionalProperties": false
+    },
+    "DependabotCooldownConfig": {
+      "type": "object",
+      "properties": {
+        "days": {
+          "default": 0,
+          "type": "integer",
+          "format": "uint64",
+          "minimum": 0.0
```
woodruffw: I'm not sure how best to solve this, but there's an `imply Default for DependabotCooldownConfig` that defines `days = 7`. So we'd ideally reflect that here, rather than implying that the default is `0`.

woodruffw: Maybe defining the same default on `schema::DependabotCooldownConfig` would work? But ideally we'd deduplicate this type.

crates/zizmor/src/config_schema.rs
woodruffw: No problem at all, thanks for fixing it!

crates/zizmor/src/config/mod.rs
kiwamizamurai: I just accidentally removed it, sorry. restored

support/zizmor.schema.json
kiwamizamurai: This was my careless mistake.

Added `#[schemars(default = "default_cooldown_days")]` to specify the correct default value of 7, and `#[schemars(range(min = 1))]` to match the NonZeroUsize constraint. The schema now shows `"default": 7` and `"minimum": 1`.

crates/zizmor/src/config/mod.rs
woodruffw: All good, thanks!

support/zizmor.schema.json
woodruffw: No worries, thanks!

woodruffw:
Thanks @kiwamizamurai! I'm going to branch off of this to experiment a bit -- I'll preserve your contributions in my branch.

Thanks again for your hard work here.

kiwamizamurai:
@woodruffw 
Thank you as well. I’m still figuring out many things from your reviews and commits, but I’m learning a lot along the way and really appreciate your feedback.


After that, I'll make a pr for https://github.com/SchemaStore/schemastore

```json
{
  "name": "zizmor",
  "description": "Configuration file for zizmor, a static analysis tool for GitHub Actions",
  "fileMatch": [
    "zizmor.yml",
    "zizmor.yaml",
    ".zizmor.yml",
    ".zizmor.yaml",
    ".github/zizmor.yml",
    ".github/zizmor.yaml"
  ],
  "url": "https://raw.githubusercontent.com/woodruffw/zizmor/main/support/zizmor.schema.json"
}
```

PR #1469, opened 2025-12-25
perf: avoid deserializing into a String

woodruffw:
We can generally avoid allocations by deserializing into a borrowed string instead. This has no effect in some cases, but in others it's probably a very small perf boost.


woodruffw:
This ends up being subtly incorrect in a way that only fails at runtime: serde will return an error if it can't produce a borrowed form, and serde-yaml only supports owned forms: https://github.com/dtolnay/serde-yaml/issues/94#issuecomment-397726097

PR #1482, opened 2025-12-30
feat: handle matrix expressions in container image clauses

woodruffw:
~~WIP.~~

I noticed this while messing around with random repos: we don't currently handle `image: ${{ matrix.image }}` gracefully, despite that being a somewhat common and reasonable thing for people to do.

What we should do is expand all values of the requested matrix context and see if any of them are an unpinned image. If we can't expand the matrix (e.g. because it's dynamic or has dynamic components), then we can issue a blanket lower-confidence finding.

### DIFF
--- a/docs/release-notes.md
+++ b/docs/release-notes.md
@@ -17,6 +17,9 @@ of `zizmor`.
 * The [cache-poisoning] audit is now aware of the @ramsey/composer-install
   action (#1489)
 
+* The [unpinned-images] audit is now significantly more precise in the presence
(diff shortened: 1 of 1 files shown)


woodruffw:
Leaving notes for myself:

- The current matrix APIs predate the context pattern/expression refactor, so they're not as precise as they could potentially be.

woodruffw:
This revealed some other gaps with the current matrix APIs, which I've marked with `BUG`. But otherwise this is a lot better and correctly propagates the matrix expansions + locations.

PR #1488, opened 2026-01-02
fix: improve setup-node cache precision

### LINKED ISSUE #1485: [BUG]: cache-poisoning rule doesn't support breaking change of actions/setup-node v6
### Pre-submission checks

### zizmor version

1.19.0

### Expected behavior

Similar to #1152, I'd expect the step in the below posted reproduction sample to NOT be flagged as all caching in `setup-node` is explicitly turned off.

Basically, if `cache: ''` is set, package caching is (explicitly) disabled.
Similarly, if `package-manager-cache: false`, caching is also explicitly disabled.

In both cases, this sets the input to the default value.

Setting both ensures that if `setup-node` changes the defaults, the workflow will not be affected and should - if I understand things correctly - not be flagged as this is explicitly preventing cache poisoning.

Of course, this will need a check that `actions/setup-node` v6 is used.

### Actual behavior

Instead it gets flagged:

```
error[cache-poisoning]: runtime artifacts potentially vulnerable to a cache poisoning attack
  --> .\.github\workflows\deploy.yml:73:9
   |
73 | /         with:
74 | |           node-version: '14'
75 | |           # Explicitly NOT using the cache functionality to prevent cache poisoning for deployments.
76 | |           cache: ''
77 | |           package-manager-cache: false
   | |______________________________________^ opt-in for caching here
```

### Reproduction steps

1. Save the below minimal reproduction sample to a file in a `.github/workflows` directory.
```yml
name: Deploy

on:
  push:
    tags:
      - '**'

jobs:
  prepare:
    runs-on: ubuntu-latest

    steps:
      - uses: actions/checkout@v6

      - name: Set up node and disable caching of dependencies
        uses: actions/setup-node@v6
        with:
          node-version-file: './.nvmrc'
          # Explicitly NOT using the cache functionality to prevent cache poisoning for deployments.
          cache: ''
          package-manager-cache: false
```
2. Run `zizmor . --pedantic` from the project root
3. Removing either the `cache` or `package-manager-cache` settings doesn't make any difference.

### Logs

```plain text

```

### Additional context

Related links:
* https://github.com/actions/setup-node/releases/tag/v6.0.0
* https://github.com/actions/setup-node#breaking-changes-in-v6
* https://github.com/actions/setup-node/pull/1374

woodruffw:
This actually has two fixes in it: 

1. We now use `ControlFieldType::Exact` for `setup-node`'s `cache:` input, since we know the exact set of inputs that enables it;
2. We fix `ControlFieldType::FreeString` more generally to not flag empty strings as satisfying.

Fixes #1485.

### DIFF
--- a/docs/release-notes.md
+++ b/docs/release-notes.md
@@ -39,6 +39,10 @@ of `zizmor`.
 * The [dependabot-cooldown] audit no longer flags missing cooldowns on
   ecosystems that don't (yet) support cooldowns, such as `opentofu` (#1480)
 
+* Fixed a false positive in the [cache-poisoning] audit where `zizmor` would
+  treat empty strings (e.g. `cache: ''`) as enabling rather than disabling
+  caching (#1482)
+
 ### Miscellaneous 🛠
 
 * zizmor's configuration now has an official JSON schema that will


jrfnl:
Thanks @woodruffw !

PR #1489, opened 2026-01-02
feat: support ramsey/composer-install in cache-poisoning audit

### LINKED ISSUE #1487: Feature: expand cache-poisoning detection for PHP ecosystem
### Pre-submission checks

The cache poisoning rule warns about the use of caching in release workflows.

In the PHP ecosystem there is a much used action runner which handles the caching of the Composer/Packagist dependencies between workflows.

I believe it would be prudent to flag the use of that action runner.

**IF** a workflow contain the following trigger:
```yaml
on:
  push:
    tags:
      - '**'
```
**AND** contains a step containing the following:
```yaml
uses: ramsey/composer-install@...
```

... this should be flagged for the potential of cache poisoning.

This could be auto-fixed like this:
```diff
-uses: ramsey/composer-install@...
+run: composer install
```

### Additional context

_No response_

woodruffw:
Closes #1487. Also improves the quality of the actual finding spans a bit.

### DIFF
--- a/docs/release-notes.md
+++ b/docs/release-notes.md
@@ -14,6 +14,9 @@ of `zizmor`.
 * The [excessive-permissions] audit is now aware of the `artifact-metadata`
   and `models` permissions (#1461)
 
+* The [cache-poisoning] audit is now aware of the @ramsey/composer-install
+  action (#1489)
+
 ### Changes ⚠️
 
 * The default policy for the [unpinned-uses] audit has changed from allowing


crates/zizmor/tests/integration/test-data/cache-poisoning/ramsey-composer-install.yml
```diff
@@ -0,0 +1,33 @@
...
+
+      - uses: ramsey/composer-install@3cf229dc2919194e9e36783941438d17239e8520 # 3.1.1
+        with:
+          # OK: caching explicitly disabled
+          ignore-cache: yes
+
+      - uses: ramsey/composer-install@3cf229dc2919194e9e36783941438d17239e8520 # 3.1.1
+        with:
+          # OK: caching explicitly disabled
+          ignore-cache: 1
```
jrfnl, line 38: Would it make sense to also test/safeguard the case where the `yes` text string is in quotes ?

Asking as the [docs of the action runner explicitly suggest using quotes](https://github.com/ramsey/composer-install#ignore-cache), so I imagine this is something which will be seen in the real world too.

```yaml
      - uses: ramsey/composer-install@3cf229dc2919194e9e36783941438d17239e8520 # 3.1.1
        with:
          # OK: caching explicitly disabled
          ignore-cache: "yes"          

      - uses: ramsey/composer-install@3cf229dc2919194e9e36783941438d17239e8520 # 3.1.1
        with:
          # OK: caching explicitly disabled
          ignore-cache: 'yes'          
```

jrfnl (commented):
Thanks @woodruffw for setting this up! Looks good to me (as a non-~~Ruby~~Rust dev). Just left one question about the tests.

crates/zizmor/tests/integration/test-data/cache-poisoning/ramsey-composer-install.yml
woodruffw: Thanks for calling this out -- both `yes` and `'yes'` should be handled equivalently transparently, since we're taking both from a YAML string and YAML should strip the quotes for us. But I'll add a testcase that demonstrates that.

woodruffw: 75d9c7e

jrfnl: Thanks & yes, I figured something like that was probably happening. Still never a bad thing to safeguard "expected" behaviour via tests ;-)

PR #1517, opened 2026-01-12
Add a 'misfeature' audit

### LINKED ISSUE #1494: New audit: "misfeature" for mis-feature usage
I've been thinking about this as a new blanket audit recently. This should cover a few things:

- The current CMD shell check in `obfuscation` is miscategorized (I just put it there for expedience). It should be under `misfeature` instead.
- `setup-python` has a `pip-install` input that will almost certainly do the wrong thing if users actually use it. This should be flagged.
- Maybe `shell:` values other than the default ones? This might have to be a pedantic finding, since `shell: python` is common and reasonable enough.

woodruffw:
Per #1494, this steals the `shell: cmd` check from `obfuscation` and adds a few other mis-feature checks.

Closes #1494.

### DIFF
--- a/crates/zizmor/src/audit/misfeature.rs
+++ b/crates/zizmor/src/audit/misfeature.rs
@@ -52,7 +52,7 @@ impl Misfeature {
                                     .annotated("installs packages in a brittle manner"),
                             )
                             .tip("always use a virtual environment to manage Python packages")
-                            .severity(Severity::Medium)
+                            .severity(Severity::Low)
                             .confidence(Confidence::High)
                             .build(step)?,
                     )
--- a/crates/zizmor/tests/integration/audit/misfeature.rs
+++ b/crates/zizmor/tests/integration/audit/misfeature.rs
@@ -7,7 +7,7 @@ fn test_setup_python_pip_install() -> anyhow::Result<()> {
             .input(input_under_test("misfeature/setup-python-pip-install.yml"))
             .run()?,
         @r"
-    warning[misfeature]: usage of GitHub Actions misfeatures
+    help[misfeature]: usage of GitHub Actions misfeatures
       --> @@INPUT@@:14:11
        |
     12 |       - uses: actions/setup-python@83679a892e2d95755f2dac6acb0bfd1e9ac5d548 # v6.1.0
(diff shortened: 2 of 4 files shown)


vszakats:
After this, projects using MSYS2 or Cygwin, with their documented/necessary
bash shell setting, are flagged "misfeature" and breaking CI.

Examples:
curl: https://github.com/curl/curl/actions/runs/21062602276/job/60572209587
libssh2: https://github.com/libssh2/libssh2/actions/runs/21064364347/job/60578038252

```yaml
shell: msys2 {0}  # MSYS2
shell: D:\cygwin\bin\bash.exe '{0}'  # Cygwin
```

Refs:
https://github.com/msys2/setup-msys2
https://github.com/cygwin/cygwin-install-action

Is there a recommended way to fix this? or to silence the warning?

vszakats:
Managed with `  # zizmor: ignore[misfeature]`. But with no chance for a proper local fix, this is annoying.

Could zizmor recognize these shells (in conjunction with these specific actions maybe), and
perhaps also make it look into the `run:` commands, which is regular bash in these cases.
(They do pass shellcheck, as verified in CI separately.)

woodruffw:
Thanks @vszakats, I agree that's much noisier than intended. I'm going to likely drop that to the 'auditor' persona in a bugfix release.

woodruffw:
Opened #1531 from your comment so I don't forget it/lose it.

vszakats:
Thank you @woodruffw!

woodruffw:
No problem! Sorry for any inconvenience this caused you; I'm also probably going to add some of those projects to zizmor's integration tests so I can observe when changes cause big differences like this 🙂

vszakats:
> No problem! Sorry for any inconvenience this caused you; I'm also probably going to add some of those projects to zizmor's integration tests so I can observe when changes cause big differences like this 🙂

No worries at all. I understand the need for more correctness,
and hitting the mark can be tricky. We had quite a few silent zizmor
updates in the recent months. Meanwhile also setting up zizmor in
more curl repos.

Feel free to use anything from curl (or libssh2) that seems useful for tests.

Also happy to help if needed!

Issue #1531, opened 2026-01-16
Noise/low value findings from 'misfeature' audit


woodruffw:
After this, projects using MSYS2 or Cygwin, with their documented/necessary
bash shell setting, are flagged "misfeature" and breaking CI.

Examples:
curl: https://github.com/curl/curl/actions/runs/21062602276/job/60572209587
libssh2: https://github.com/libssh2/libssh2/actions/runs/21064364347/job/60578038252

```yaml
shell: msys2 {0}  # MSYS2
shell: D:\cygwin\bin\bash.exe '{0}'  # Cygwin
```

Refs:
https://github.com/msys2/setup-msys2
https://github.com/cygwin/cygwin-install-action

Is there a recommended way to fix this? or to silence the warning?

_Originally posted by @vszakats in https://github.com/zizmorcore/zizmor/pull/1517#issuecomment-3759740853_

woodruffw:
Noting for myself: I need to also add some of these to the integration tests, to catch these kinds of disruptive changes more preemptively.

PR #1532, opened 2026-01-16
Drop 'custom shell' finding to auditor persona

woodruffw:
See #1531.

CC @vszakats

### DIFF
--- a/crates/zizmor/tests/integration/audit/misfeature.rs
+++ b/crates/zizmor/tests/integration/audit/misfeature.rs
@@ -31,7 +31,7 @@ fn test_non_well_known_shell() -> anyhow::Result<()> {
     insta::assert_snapshot!(
         zizmor()
             .input(input_under_test("misfeature/non-well-known-shell.yml"))
-            .args(["--persona=pedantic"])
+            .args(["--persona=auditor"])
             .run()?,
         @r"
     help[misfeature]: usage of GitHub Actions misfeatures


vszakats:
LGTM, though that makes we want to try the auditor persona! (curl/libssh2 passes)

Do you see a possibility to accept known shells outside the offical ones in GH docs?

To avoid the rabbit hole of maintaining an allow-list, would it work to allow a markup
telling zizmor to consider a special shell value as an alias of a well-known one?
E.g. `# zizmor: shell[bash]`

It'd would also help with this special case:
```yaml
shell: ${{ contains(matrix.image, 'windows') && 'msys2 {0}' || 'bash' }}  # zizmor: shell[bash]
```
it's not triggering a warning as of 1.21.0, but prevents analyzing run commands.

woodruffw:
> LGTM, though that makes we want to try the auditor persona! (curl/libssh2 passes)

Please do, although I definitely recommend not using it to gate CI runs 😅 -- it's intentionally used to isolate incredibly noisy/low-impact findings (and it can change a lot between releases), so it might be pretty frustrating to have as a required check.

The way I personally use it is for manual periodic check-ups: CI does a regular persona run, and then I periodically run a pedantic or auditor run locally (every few months) to shake smaller things out. 

> Do you see a possibility to accept known shells outside the offical ones in GH docs?

My default would be to do this with an allow-list, although I think you're right that this might be a rabbit-hole/too-intrusive for most people 😅 

```yaml
rules:
  misfeature:
    config:
      extend-well-known-shells:
        - msys2
        - something-else
```

Do you think that would be too intrusive? I don't have a super strong intuition here.

> E.g. `# zizmor: shell[bash]`

I'm a *little* hesitant to add more special comment forms -- they add another layer on top of config that's harder to constrain later on. But I do agree this is pretty clean/obvious compared to a config section.

vszakats:
> > LGTM, though that makes we want to try the auditor persona! (curl/libssh2 passes)
> 
> Please do, although I definitely recommend not using it to gate CI runs 😅 -- it's intentionally used to isolate incredibly noisy/low-impact findings (and it can change a lot between releases), so it might be pretty frustrating to have as a required check.
>
> The way I personally use it is for manual periodic check-ups: CI does a regular persona run, and then I periodically run a pedantic or auditor run locally (every few months) to shake smaller things out.
> 
> > Do you see a possibility to accept known shells outside the offical ones in GH docs?

Sounds like a plan, and maybe pedantic is too much for CI, yes. Though so far
I managed. I've been thinking if the occasional new fallout is the bigger "cost",
or Renovate bugging for a manual bump per release. So far I think the former is
more practical. Also simpler.
 
> My default would be to do this with an allow-list, although I think you're right that this might be a rabbit-hole/too-intrusive for most people 😅

It certainly is an option. The landscape may not be wide;
the only outliers I know are these two. It's perhaps all there is.
Either way it's always an option to make it more complex if it
becomes necessary.

> ```yaml
> rules:
>   misfeature:
>     config:
>       extend-well-known-shells:
>         - msys2
>         - something-else
> ```
> 
> Do you think that would be too intrusive? I don't have a super strong intuition here.

No, it feels fine to me.

The trickier bit is Cygwin (also MSYS2 supports it), because
the shell can be at any path. (e.g. `C:\cygwin\bin\bash.exe {0}`)

If there is an option for a wildcard, `*bash.exe*` (or similar)
would catch most combinations:

<https://github.com/search?q=shell%3A+bash.exe+language%3Ayaml&type=code>

(`.exe` may be optional, but I think it's safe to count on it
for majority, if not all, cases.)

> > E.g. `# zizmor: shell[bash]`
> 
> I'm a _little_ hesitant to add more special comment forms -- they add another layer on top of config that's harder to constrain later on. But I do agree this is pretty clean/obvious compared to a config section.

Agreed, this route adds more complexity, and perhaps overkill for this.

woodruffw:
> The trickier bit is Cygwin (also MSYS2 supports it), because the shell can be at any path. (e.g. `C:\cygwin\bin\bash.exe {0}`)
> 
> If there is an option for a wildcard, `*bash.exe*` (or similar) would catch most combinations:

Ah yeah, I think this is actually a separate bug 🙂 -- zizmor _should_ be normalizing paths in shells to turn `...\bash.exe` into just `bash`, but I strongly suspect I forgot to handle Windows-style paths.

(That normalization is technically wrong anyways too, since `shell: /custom/bash` *should* be an auditor finding but isn't currently since we consider it the same as the "well-known" bash.)

PR #1537, opened 2026-01-17
Read `GITHUB_TOKEN`

branchv:
## Pre-submission checks

Please check these boxes:

      one first).

## Summary

Forgive me for not creating an issue first 😃. But currently, only `GH_TOKEN` is respected while `GITHUB_TOKEN` is (if not more) standard (and read by `gh`: https://cli.github.com/manual/gh_help_environment)

### DIFF
--- a/crates/zizmor/src/main.rs
+++ b/crates/zizmor/src/main.rs
@@ -730,10 +730,17 @@ async fn run(app: &mut App) -> Result<ExitCode, Error> {
         app.persona = Persona::Pedantic;
     }
 
+    // Respect GITHUB_TOKEN in addition to GH_TOKEN.
+    if app.gh_token.is_none()
+        && let Ok(token) = env::var("GITHUB_TOKEN")
+    {
+        app.gh_token = GitHubToken::new(&token).ok();
+    }
+
     // Unset the GitHub token if we're in offline mode.
     // We do this manually instead of with clap's `conflicts_with` because
     // we want to support explicitly enabling offline mode while still
-    // having `GH_TOKEN` present in the environment.
+    // having `GH_TOKEN` or `GITHUB_TOKEN` present in the environment.
     if app.offline {
         app.gh_token = None;
     }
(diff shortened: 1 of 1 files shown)


woodruffw:
Thanks @branchv! I'll give this a look tomorrow.

(From memory, `gh` uses `GH_TOKEN` as the primary variable and `GITHUB_TOKEN` as the fallback. But I agree zizmor should accept both, since they're both common 🙂)

branchv:
Yup that's right, this follows that same precedence. And no rush

crates/zizmor/src/main.rs
```diff
@@ -730,10 +730,17 @@ async fn run(app: &mut App) -> Result<ExitCode, Error> {
         app.persona = Persona::Pedantic;
     }
 
+    // Respect GITHUB_TOKEN in addition to GH_TOKEN.
+    if app.gh_token.is_none()
+        && let Ok(token) = env::var("GITHUB_TOKEN")
+    {
+        app.gh_token = GitHubToken::new(&token).ok();
+    }
```
woodruffw, line 738: Hmm, I think my main concern with this approach is that all other envvars are handled via clap, which ensures that clap considers them for satisfaction/conflict resolution, etc. And naturally clap doesn't support multiple environment variables per argument 😞 

Another idea: maybe we could add a new `--github-token` option that defines `GITHUB_TOKEN` as its environment variable? That would allow us to ensure conflicts are handled correctly, plus we could mark `--github-token` as a hidden option so that people still mostly use `--gh-token` via the CLI.

woodruffw:
Thanks @branchv! I agree this would be good to have, but I'm hesitant to do it as a layer atop clap -- I left a comment on the diff itself with an idea for how we could do this directly within clap instead.

woodruffw:
Opened #1566 for doing this within clap's derive API. Thanks again @branchv!

branchv:
awesome, thank you @woodruffw!

PR #1555, opened 2026-01-22
LSP: config support

### LINKED ISSUE #1002: [BUG]: LSP mode: violation reported despite configuration for action
### Pre-submission checks

### zizmor version

1.11.0

### Expected behavior

I tried out the new VSCode extension with the latest zizmor version. Thanks a lot for that, very helpful!

I have a `zizmor.yml` where `unpinned-uses` is configured for certain actions to be allow `ref-pin`. I have a file in the repo `.github/actions/some-action/action.yaml` that uses one of those actions but the zizmor vscode extension still reports a violation for each action use with a ref-pin.

`zizmor` itself does not:

```shell
$ zizmor .github/actions/docker-build/action.yaml
 INFO zizmor::registry: skipping impostor-commit: can't run without a GitHub API token
 INFO zizmor::registry: skipping ref-confusion: can't run without a GitHub API token
 INFO zizmor::registry: skipping known-vulnerable-actions: can't run without a GitHub API token
 INFO zizmor::registry: skipping forbidden-uses: audit not configured
 INFO zizmor::registry: skipping stale-action-refs: can't run without a GitHub API token
 INFO audit: zizmor: 🌈 completed .github/actions/docker-build/action.yaml
No findings to report. Good job! (2 ignored, 2 suppressed)
```

The relevant file is here: https://github.com/opalmedapps/.github/blob/main/.github/actions/docker-build/action.yaml

### Actual behavior

Match the report of `zizmor`, i.e., no violation in this particular use case.

### Reproduction steps

1. Clone repo: https://github.com/opalmedapps/.github
2. Open in vscode with the zizmor extension installed and enabled
3. Open `.github/actions/docker-build/action.yaml`

### Logs

```plain text

```

### Additional context

_No response_

woodruffw:
~~WIP.~~

The basic idea here:

1. Register document monitoring events for zizmor config file patterns
2. Differentiate those events from normal auditable input events
3. Keep a config loaded/updated in `lsp::Backend`
4. Use that config in audits

Another piece of complexity is that (1) isn't enough -- we also need to proactively load the config on LSP initialization. I'm pretty sure the right way to do this is to get the `workspace_folders` from the LSP initialization parameters and walk those for initial configuration.

Closes #1002.

### DIFF
--- a/crates/zizmor/src/lsp.rs
+++ b/crates/zizmor/src/lsp.rs
@@ -276,8 +276,8 @@ impl Backend {
         // Try to find a configuration file for this audit.
         // The approach below is probably wrong: we scan each workspace directory
         // in order and use the first configuration we find. Instead, we should
-        // probably find the configuration file that is "closest" to the input
-        // being audited.
+        // probably find the configuration file that is in the "closest"
+        // workspace to the file being audited.
         let config = {
             let mut config = Config::default();
             let workspace_dirs = self.workspace_dirs.read().await;


woodruffw:
I ended up going a pretty different (much simpler, but also incorrect in some ways that may matter later) approach: instead of monitoring for the config files, we stash the set of workspace directories at initialization, then use those to discover a config file on each audit. This is really simple, but it has some downsides:

1. It's slower than necessary, since we're reloading the config on every single change. Ideally we'd cache this and only reload it when it changes, which brings things back to needing to monitor/watch for file events. In practice it's not meaningfully slow now and probably won't be unless zizmor becomes a lot slower in general.
2. It's definitely wrong in a whole slew of scenarios where there's more than one workspace directory. I don't really understand these scenarios super well (I never use VS Code with more than one workspace at a time), so we'll cross this bridge when we come to it.
3. Changes to the config don't invalidate existing diagnostics -- the user has to save or otherwise induce a change on their input to force the zizmor LSP to refresh. This would also be addressed by monitoring for configs.

PR #1559, opened 2026-01-22
feat(uses): add glob pattern support for repository matching

ryandelano:
## Pre-submission checks

Please check these boxes:

      one first) - discussed with @woodruffw on discord

      *will* be rejected for undisclosed use. - code review, slight optimization regarding string allocation

## Summary
Add support for glob wildcards (e.g., `foo-*`, `*-bar`, `foo-*-bar`) in repository and subpath pattern matching for `uses:` clauses. This allows more flexible policy configuration for any audit that uses repository patterns, such as `unpinned-uses` and `forbidden-uses`. Used intuition for each of the possible cases and how they should behave, so I'm happy to adjust based on other perspectives!

Changes:
- Add `Segment` enum to represent exact or glob pattern matches
- Add `ParsedSegment` enum for parsing patterns with `*` wildcards
- Update `RepositoryUsesPattern` to use `Segment` for repo/subpath fields
- Implement specificity ordering so exact patterns take precedence over globs
- Optimize case-insensitive matching to avoid string allocations
- Add comprehensive tests for glob pattern parsing and matching
- Document glob patterns in configuration.md

## New Patterns Supported

- `owner/prefix-*` - match repos starting with `prefix-`
- `owner/*-suffix` - match repos ending with `-suffix`
- `owner/prefix-*-suffix` - match repos with both prefix and suffix
- `owner/prefix-*/*` - match repos with prefix, any subpath
- `owner/repo/subpath-*` - match subpaths with glob
- Combined: `owner/prefix-*/subpath-*`

### unpinned-uses:

```yaml
rules:
  unpinned-uses:
    config:
      policies:
        "github/codeql-*/*": "hash-pin"  # All codeql actions
        "myorg/action-*": "ref-pin"       # All myorg action-* repos
```

### forbidden-uses:

```yaml
rules:
  forbidden-uses:
    config:
      allow:
        - "actions/*"           # Allow all actions org repos
        - "myorg/trusted-*"     # Allow myorg repos starting with "trusted-"
      # OR
      deny:
        - "untrusted-org/*-exploit"  # Deny repos ending with "-exploit"
 ```

## Implementation

- New `Segment` enum handles exact vs glob matching
- Glob patterns with single * are supported while multiple wildcards per segment are rejected
- Specificity ordering ensures exact patterns take precedence over globs
- Case-insensitive matching optimized to avoid heap allocations

## Test Plan

- Added unit tests for `Segment` and `ParsedSegment` parsing
- Added pattern matching tests covering all glob variants
- Added ordering tests to verify specificity
- Added integration tests for `unpinned-uses` with invalid multi-wildcard patterns and `forbidden-uses` with glob patterns in allow/deny lists
- All existing tests continue to pass

### DIFF
--- a/crates/zizmor/src/models/uses.rs
+++ b/crates/zizmor/src/models/uses.rs
@@ -1,6 +1,6 @@
 //! Extension traits for the `Uses` APIs.
 
-use std::{cmp::Ordering, str::FromStr, sync::LazyLock};
+use std::{cmp::Ordering, cmp::Reverse, str::FromStr, sync::LazyLock};
 
 use github_actions_models::common::{RepositoryUses, Uses};
 use regex::Regex;
@@ -144,12 +144,13 @@ impl Segment {
     /// Lower values are more specific.
     /// Exact matches are more specific than globs.
     /// For globs, longer prefix+suffix means more specific.
-    fn specificity(&self) -> (u8, usize) {
+    fn specificity(&self) -> (u8, Reverse<usize>) {
         match self {
             // Exact is most specific (0), with no length consideration
-            Segment::Exact(_) => (0, 0),
-            // Glob is less specific (1), but longer literals are more specific (inverted)
-            Segment::Glob { prefix, suffix } => (1, usize::MAX - (prefix.len() + suffix.len())),
+            Segment::Exact(_) => (0, Reverse(0)),
+            // Glob is less specific (1), but longer literals are more specific
+            // Using Reverse so that longer lengths sort first (more specific)
+            Segment::Glob { prefix, suffix } => (1, Reverse(prefix.len() + suffix.len())),
         }
     }
 }
@@ -271,21 +272,18 @@ impl RepositoryUsesPattern {
 
     /// Returns a tuple used for ordering patterns by specificity.
     /// Lower values are more specific and should come first when sorted.
-    fn specificity(&self) -> (u8, Segment, Option<Segment>) {
-        // Create a dummy segment for comparison purposes
-        let no_segment = Segment::Exact(String::new());
-
+    fn specificity(&self) -> (u8, Option<Segment>, Option<Segment>) {
         match self {
             RepositoryUsesPattern::ExactWithRef { repo, subpath, .. } => {
-                (0, repo.clone(), subpath.clone())
+                (0, Some(repo.clone()), subpath.clone())
             }
             RepositoryUsesPattern::ExactPath { repo, subpath, .. } => {
-                (1, repo.clone(), Some(subpath.clone()))
+                (1, Some(repo.clone()), Some(subpath.clone()))
             }
-            RepositoryUsesPattern::ExactRepo { repo, .. } => (2, repo.clone(), None),
-            RepositoryUsesPattern::InRepo { repo, .. } => (3, repo.clone(), None),
-            RepositoryUsesPattern::InOwner(_) => (4, no_segment.clone(), None),
-            RepositoryUsesPattern::Any => (5, no_segment, None),
+            RepositoryUsesPattern::ExactRepo { repo, .. } => (2, Some(repo.clone()), None),
+            RepositoryUsesPattern::InRepo { repo, .. } => (3, Some(repo.clone()), None),
+            RepositoryUsesPattern::InOwner(_) => (4, None, None),
+            RepositoryUsesPattern::Any => (5, None, None),
         }
     }
 }
@@ -320,11 +318,9 @@ impl FromStr for RepositoryUsesPattern {
         let subpath_str = caps.get(3).map(|m| m.as_str());
         let git_ref = caps.get(4).map(|m| m.as_str());
 
-        // Parse repo segment (handles validation of multiple wildcards)
         let repo_parsed = ParsedSegment::parse(repo_str)
             .ok_or_else(|| anyhow::anyhow!("invalid pattern: {s}"))?;
 
-        // Parse subpath segment if present
         let subpath_parsed = subpath_str
             .map(|sp| {
                 ParsedSegment::parse(sp).ok_or_else(|| anyhow::anyhow!("invalid pattern: {s}"))
@@ -333,9 +329,7 @@ impl FromStr for RepositoryUsesPattern {
 
         // Build the appropriate pattern variant
         match (&repo_parsed, &subpath_parsed, git_ref) {
-            // ================================================================
-            // owner/* - matches any repo under owner
-            // ================================================================
+            // owner/* matches any repo under owner
             (ParsedSegment::Star, None, None) => Ok(RepositoryUsesPattern::InOwner(owner.into())),
 
             // owner/*@ref is invalid (can't have ref with repo star)
@@ -344,9 +338,6 @@ impl FromStr for RepositoryUsesPattern {
             // owner/*/... is invalid (can't have subpath after repo star)
             (ParsedSegment::Star, Some(_), _) => Err(anyhow::anyhow!("invalid pattern: {s}")),
 
-            // ================================================================
-            // Patterns without subpath
-            // ================================================================
             // owner/repo or owner/repo-* (no subpath, no ref)
             (ParsedSegment::Segment(repo), None, None) => Ok(RepositoryUsesPattern::ExactRepo {
                 owner: owner.into(),
@@ -363,9 +354,6 @@ impl FromStr for RepositoryUsesPattern {
                 })
             }
 
-            // ================================================================
-            // Patterns with subpath = *
-            // ================================================================
             // owner/repo/* or owner/repo-*/* (any subpath)
             (ParsedSegment::Segment(repo), Some(ParsedSegment::Star), None) => {
                 Ok(RepositoryUsesPattern::InRepo {
@@ -377,9 +365,6 @@ impl FromStr for RepositoryUsesPattern {
             // owner/repo/*@ref is invalid (can't combine subpath star with ref)
             (_, Some(ParsedSegment::Star), Some(_)) => Err(anyhow::anyhow!("invalid pattern: {s}")),
 
-            // ================================================================
-            // Patterns with exact or glob subpath
-            // ================================================================
             // owner/repo/subpath or owner/repo-*/subpath or owner/repo/subpath-*
             (ParsedSegment::Segment(repo), Some(ParsedSegment::Segment(subpath)), None) => {
                 Ok(RepositoryUsesPattern::ExactPath {
@@ -748,9 +733,6 @@ mod tests {
             ("owner/repo/subpath@*", None),
             ("owner/*@*", None),
             ("*@*", None),
-            // ================================================================
-            // NEW: Glob patterns
-            // ================================================================
             // Repo globs
             (
                 "owner/foo-*",
@@ -1014,9 +996,6 @@ mod tests {
             ("actions/checkout@v3", "actions/checkout@v3", true),
             ("actions/checkout/foo@v3", "actions/checkout/foo@v3", true),
             ("actions/checkout/foo@v1", "actions/checkout/foo@v3", false),
-            // ================================================================
-            // NEW: Glob pattern matching
-            // ================================================================
             // Repo globs - basic matching
             ("org/foo-bar@v1", "org/foo-*", true),
             ("org/foo-baz@v1", "org/foo-*", true),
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

File: crates/zizmor/src/models/uses.rs [hunk 1]
```diff
@@ -14,56 +14,208 @@ static REPOSITORY_USES_PATTERN: LazyLock<Regex> = LazyLock::new(|| {
...
+/// - `Star`: the full wildcard `*` (used for `owner/*` or `owner/repo/*`)
+/// - `Segment`: an exact match or glob pattern
+/// - Parse failure (multiple wildcards)
+#[derive(Debug, Clone, PartialEq)]
+pub(crate) enum ParsedSegment {
+    /// Just `*` - matches anything in this position
+    Star,
+    /// A concrete segment (exact or glob)
+    Segment(Segment),
+}
```
woodruffw: Q: can we unify these types? i.e. just one `enum Segment` with `{Star, Glob, Exact}`? I think that would be slightly easier to follow here.

crates/zizmor/src/models/uses.rs
```diff
@@ -96,27 +251,62 @@ impl RepositoryUsesPattern {
...
             RepositoryUsesPattern::InOwner(owner) => uses.owner().eq_ignore_ascii_case(owner),
             RepositoryUsesPattern::Any => true,
         }
     }
+
+    /// Returns a tuple used for ordering patterns by specificity.
+    /// Lower values are more specific and should come first when sorted.
+    fn specificity(&self) -> (u8, Segment, Option<Segment>) {
+        // Create a dummy segment for comparison purposes
+        let no_segment = Segment::Exact(String::new());
```
woodruffw: Making sure I understand: is the idea here that not all patterns have a relevant segment, so we're putting a dummy one in to make `Ord` and `PartialOrd` work below?

If so, I think we could maybe make this clearer by having this return `(u8, Option<Segment>, Option<Segment>)` and handling the optional segment twice.

crates/zizmor/src/models/uses.rs
```diff
@@ -126,49 +316,95 @@ impl FromStr for RepositoryUsesPattern {
...
-                repo: repo.into(),
-            }),
-            (owner, repo, Some(subpath), None) => Ok(RepositoryUsesPattern::ExactPath {
-                owner: owner.into(),
-                repo: repo.into(),
-                subpath: subpath.into(),
-            }),
-            (_, _, Some("*"), Some(_)) => Err(anyhow::anyhow!("invalid pattern: {s}")),
-            (owner, repo, subpath, Some(git_ref)) => Ok(RepositoryUsesPattern::ExactWithRef {
+        // Parse repo segment (handles validation of multiple wildcards)
```
woodruffw:
```suggestion
```

crates/zizmor/src/models/uses.rs
```diff
@@ -126,49 +316,95 @@ impl FromStr for RepositoryUsesPattern {
...
+        let repo_parsed = ParsedSegment::parse(repo_str)
+            .ok_or_else(|| anyhow::anyhow!("invalid pattern: {s}"))?;
+
+        // Parse subpath segment if present
```
woodruffw:
```suggestion
```

crates/zizmor/src/models/uses.rs
```diff
@@ -126,49 +316,95 @@ impl FromStr for RepositoryUsesPattern {
...
+            .map(|sp| {
+                ParsedSegment::parse(sp).ok_or_else(|| anyhow::anyhow!("invalid pattern: {s}"))
+            })
+            .transpose()?;
+
+        // Build the appropriate pattern variant
+        match (&repo_parsed, &subpath_parsed, git_ref) {
+            // ================================================================
+            // owner/* - matches any repo under owner
+            // ================================================================
```
woodruffw:
```suggestion
            // owner/* matches any repo under owner
```

crates/zizmor/src/models/uses.rs
```diff
@@ -126,49 +316,95 @@ impl FromStr for RepositoryUsesPattern {
...
+
+            // owner/*@ref is invalid (can't have ref with repo star)
+            (ParsedSegment::Star, None, Some(_)) => Err(anyhow::anyhow!("invalid pattern: {s}")),
+
+            // owner/*/... is invalid (can't have subpath after repo star)
+            (ParsedSegment::Star, Some(_), _) => Err(anyhow::anyhow!("invalid pattern: {s}")),
+
+            // ================================================================
+            // Patterns without subpath
+            // ================================================================
```
woodruffw:
```suggestion
```

crates/zizmor/src/models/uses.rs
```diff
@@ -126,49 +316,95 @@ impl FromStr for RepositoryUsesPattern {
...
+                    owner: owner.into(),
+                    repo: repo.clone(),
+                    subpath: None,
+                    git_ref: r.into(),
+                })
+            }
+
+            // ================================================================
+            // Patterns with subpath = *
+            // ================================================================
```
woodruffw:
```suggestion
```

crates/zizmor/src/models/uses.rs
```diff
@@ -126,49 +316,95 @@ impl FromStr for RepositoryUsesPattern {
...
+                    repo: repo.clone(),
+                })
+            }
+
+            // owner/repo/*@ref is invalid (can't combine subpath star with ref)
+            (_, Some(ParsedSegment::Star), Some(_)) => Err(anyhow::anyhow!("invalid pattern: {s}")),
+
+            // ================================================================
+            // Patterns with exact or glob subpath
+            // ================================================================
```
woodruffw:
```suggestion
```

crates/zizmor/src/models/uses.rs
```diff
@@ -382,26 +730,162 @@ mod tests {
...
                     git_ref: "release/v1".into(),
                 }),
             ),
-            // Invalid: no wildcards allowed when refs are present.
+            // Invalid: no wildcards allowed when refs are present (subpath star only).
             ("owner/repo/*@v1", None),
-            ("owner/repo/*/subpath@v1", None),
-            ("owner/*/subpath@v1", None),
-            ("*/*/subpath@v1", None),
+            // Note: owner/repo/*/subpath@v1 is now VALID - it's a subpath glob matching "*" + "/subpath"
```
woodruffw: I think this and other LLM comments can generally be removed -- they're not adding a ton of detail or context to the tests here.

```suggestion
```

crates/zizmor/src/models/uses.rs
```diff
@@ -382,26 +730,162 @@ mod tests {
...
             // Ref also cannot be a wildcard.
             ("owner/repo@*", None),
             ("owner/repo@**", None),
             ("owner/repo@***", None),
             ("owner/repo/subpath@*", None),
             ("owner/*@*", None),
             ("*@*", None),
+            // ================================================================
+            // NEW: Glob patterns
+            // ================================================================
```
woodruffw:
```suggestion
```

crates/zizmor/src/models/uses.rs
```diff
@@ -484,6 +1014,56 @@ mod tests {
             ("actions/checkout@v3", "actions/checkout@v3", true),
             ("actions/checkout/foo@v3", "actions/checkout/foo@v3", true),
             ("actions/checkout/foo@v1", "actions/checkout/foo@v3", false),
+            // ================================================================
+            // NEW: Glob pattern matching
+            // ================================================================
```
woodruffw:
```suggestion
```

docs/configuration.md
```diff
@@ -237,3 +237,81 @@ The following patterns are supported, in order of specificity:
 
         `*` matches `#!yaml uses: actions/checkout` and
         `#!yaml uses: pypa/gh-action-pypi-publish@release/v1`.
+
+#### Glob patterns
```
woodruffw: Two nits here:

1. I think we should _probably_ call these something other than "glob patterns" -- that term implies (IMO) full compatibility with `fnmatch(2)` syntax, when really what we're offering here is a very constrained wildcard match. Maybe we don't call them anything specific at all, and just describe them inline with the rest of the patterns? I'm not sure.
2. I think it'd be nice to unify this section more fully with the "Repository patterns" section right above it -- right now a user who reads this is going to go through the "Repository patterns" section, and then have to correct/adjust their understanding of how patterns work when they reach this lower section.

crates/zizmor/src/models/uses.rs
```diff
@@ -14,56 +14,208 @@ static REPOSITORY_USES_PATTERN: LazyLock<Regex> = LazyLock::new(|| {
...
+                    .all(|(v, s)| v.eq_ignore_ascii_case(s))
+            }
+        }
+    }
+
+    /// Returns the "specificity" of this segment for ordering purposes.
+    /// Lower values are more specific.
+    /// Exact matches are more specific than globs.
+    /// For globs, longer prefix+suffix means more specific.
+    fn specificity(&self) -> (u8, usize) {
```
woodruffw: Q: can we put this inline in the `Ord` implementation? Unless I'm missing something, I don't think having it in its own function does a ton for us.

(Now that I think about it I guess there's a benefit to being able to declare the ordering "key" distinct from `Ord`/without referencing two instances, so if that's the rationale that seems fine to me 🙂)

crates/zizmor/src/models/uses.rs
```diff
@@ -14,56 +14,208 @@ static REPOSITORY_USES_PATTERN: LazyLock<Regex> = LazyLock::new(|| {
...
+        match self {
+            // Exact is most specific (0), with no length consideration
+            Segment::Exact(_) => (0, 0),
+            // Glob is less specific (1), but longer literals are more specific (inverted)
+            Segment::Glob { prefix, suffix } => (1, usize::MAX - (prefix.len() + suffix.len())),
```
woodruffw: Can't this be `prefix.len() + suffix.len()`? Why do we have to begin at `size:::MAX` and subtract from there?

woodruffw (commented):
Thanks @ryandelano! This looks pretty good to me overall, I've left mostly questions and nitpicks in this round of review 🙂

crates/zizmor/src/models/uses.rs
ryandelano: We could, my reasoning was that `Star` acts purely a signal for which variant we're constructing given that it's never actually stored in a `RepositoryUsesPattern`. Unifying them would mean handling `Star` in the matching phase, which we could pretty easily do with an assertion that `Segment::Star` never appears after the construction phase. I didn't put it in the newest batch of changes, but does that approach sound better?

crates/zizmor/src/models/uses.rs
ryandelano: Yes you're right. Added!

crates/zizmor/src/models/uses.rs
ryandelano: Apologies, slipped through after using it for review!

docs/configuration.md
ryandelano: Makes sense. They're inline with the rest of the patterns now, so you can see how you like the flow. 

I also went ahead and renamed everything related to `Glob` in the code/comments to `Wildcard` to reduce confusion and make sure things match between docs and code.

crates/zizmor/src/models/uses.rs
ryandelano: That benefit is exactly what I was going for haha, just debugging/separation of concerns. Can inline as well if you'd prefer

crates/zizmor/src/models/uses.rs
ryandelano: The goal was to make sure that longer prefix/suffix literals should be considered "more specific" (and therefore be sorted earlier), but looking back over this, using `Reverse` for that makes that goal much clearer, so I refactored it.

ryandelano:
Responded to everything and resolved the comments changes, keep me posted on anything else :) 

edit: thought I re-requested the review when I commented this and apparently didn't, sorry!!

woodruffw:
> Responded to everything and resolved the comments changes, keep me posted on anything else :)
> 
> edit: thought I re-requested the review when I commented this and apparently didn't, sorry!!

Thanks, and no worries! I'm going to try and set aside some time to review this over the weekend.

woodruffw:
Sorry for the delay here, I'm doing another review tonight.

ryandelano:
Just checking in on this @woodruffw, let me know if there's any other info needed as I don't want to be a bother but would love to start using this in our workflows :)

woodruffw:
No worries. Sorry for the delayed review, I'm going to have a bit less time in the coming weeks for larger PRs. But I'll try and give this another pass soon.

ryandelano:
Totally makes sense, very much appreciate it!!

PR #1562, opened 2026-01-23
Fix anchor handling in flow sequences and mappings

### LINKED ISSUE #1558: [BUG]: Anchor prefix not stripped from values in flow sequences
### Pre-submission checks

### zizmor version

1.22 (main)

### Expected behavior

With the following structure:
```yaml
foo: [&x v, *x]
```

The query [foo, 0] returns `v`

### Actual behavior

The query returns `&x v`.

### Reproduction steps

You can add the following test to check the behaviour:

```rust
    #[test]
    fn test_inline_anchor_alias_pattern() {
        let yaml = r#"
foo: [&x v, *x]
        "#;

        let doc = Document::new(yaml).unwrap();

        let feature = doc.query_exact(&route!("foo", 0)).unwrap().unwrap();
        assert_eq!(doc.extract(&feature), "v");

        let feature = doc.query_exact(&route!("foo", 1)).unwrap().unwrap();
        assert_eq!(doc.extract(&feature), "v");
    }
```

### Logs

```plain text
---- tests::test_inline_anchor_alias_pattern stdout ----

thread 'tests::test_inline_anchor_alias_pattern' (400764) panicked at crates/yamlpath/src/lib.rs:1364:9:
assertion `left == right` failed
  left: "&x v"
 right: "v"
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

```

### Additional context

We are missing a test in `anchors-list` for this behavior.

DarkaMaul:
## Pre-submission checks

Please check these boxes:

      one first).: #1558 

## Summary

Anchor prefixes in YAML flow sequences and flow mappings are not being stripped during path queries, causing incorrect results or query failures.

### Bug 1: Flow Sequences

When querying a flow sequence with anchored values like `foo: [&x v, *x]`, the anchor prefix is incorrectly included in the extracted value:

```
YAML: foo: [&x v, *x]
  left: "&x v"
 right: "v"
```

### Bug 2: Flow Mappings

When querying a flow mapping with anchored keys like `top: { &a foo: &b bar }`, the key lookup fails entirely because it compares against the anchor-prefixed key:

```
called `Result::unwrap()` on an `Err` value: ExhaustedMapping("foo")
```

## Test Plan

I've added tests that I hope cover enough of the edge cases.

Fixes #1558.


crates/yamlpath/src/lib.rs
```diff
@@ -911,25 +920,35 @@ impl Document {
...
-                        .utf8_text(self.source().as_bytes())
-                        .expect("impossible: value for key should be UTF-8 by construction");
-
-                    match scalar.kind() {
-                        "single_quote_scalar" | "double_quote_scalar" => {
-                            let mut chars = key_value.chars();
-                            chars.next();
-                            chars.next_back();
-                            chars.as_str()
+            //
```
woodruffw:
```suggestion
```

docs/release-notes.md
```diff
@@ -25,6 +25,9 @@ of `zizmor`.
 
     Many thanks to @DarkaMaul for implementing this fix!
 
+* Fixed a bug in `yamlpath` where anchor prefixes in sequences and mapping
+  where not stripped during path queries (#1562)
```
woodruffw, line 29:
```suggestion
  were not stripped during path queries (#1562)
```

PR #1576, opened 2026-01-26
feat: remove ids fields and only use string comparisons

DarkaMaul:
## Pre-submission checks

Please check these boxes:

      one first).

- [X] I hereby disclose the use of an LLM or other AI coding assistant in the

## Summary

As discussed, a tentative change to remove the tree-sitter ids from yamlpath.
I have run local benchmarks, and the performance impact looks negligible.

## Test Plan

Since this is an internal only change, no new tests were added.
The test suite is expected to pass

### DIFF
--- a/crates/yamlpatch/src/lib.rs
+++ b/crates/yamlpatch/src/lib.rs
@@ -866,7 +866,7 @@ fn handle_block_mapping_addition(
     // Check if we need to add a newline before the entry
     // If the content at insertion point already ends with a newline, don't add another
     let needs_leading_newline = if insertion_point > 0 {
-        doc.source().chars().nth(insertion_point - 1) != Some('\n')
+        doc.source().as_bytes().get(insertion_point - 1) != Some(&b'\n')
     } else {
         true
     };
--- a/crates/yamlpatch/tests/unit_tests.rs
+++ b/crates/yamlpatch/tests/unit_tests.rs
@@ -1817,6 +1817,44 @@ fn test_merge_into_prevents_duplicate_keys() {
 }
 
 #[test]
+fn test_merge_into_with_unicode() {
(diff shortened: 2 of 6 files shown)


crates/yamlpath/src/lib.rs
```diff
@@ -542,7 +508,7 @@ impl Document {
     pub fn query_exact(&self, route: &Route) -> Result<Option<Feature<'_>>, QueryError> {
         let node = self.query_node(route, QueryMode::Exact)?;
 
-        if node.kind_id() == self.block_mapping_pair_id || node.kind_id() == self.flow_pair_id {
+        if matches!(node.kind(), "block_mapping_pair" | "flow_pair") {
```
woodruffw: As part of this refactor, I think it'd be cool if we made this more fluent. 

For example, we could have a `NodeExt` trait that has `is_pair(&self) -> bool`, `is_anchor(&self) -> bool`, etc., and then do `impl NodeExt for Node` to give ourselves a `Node` level-API. That would also cut down on number of open-coded comparisons we need to do 🙂

DarkaMaul: I had a go with it but I'm not super convinced by the API yet.

woodruffw: Hmm yeah, it's still kind of verbose. I'm not sure what else we could do to make this terser though; I'd be fine with merging as-is, unless you have other ideas here 🙂

DarkaMaul: I would go on with merging this one until we have a better option.

It's an internal, contained change, so easy to rollback, or iterate.

woodruffw:
That benchmark regression is almost certainly statistical noise/GitHub randomly assigning slower runners occasionally.

PR #1581, opened 2026-01-29
fix(yamlpatch): handle Unicode in MergeInto newline detection

### LINKED ISSUE #1580: [BUG]: auto-fix produces invalid YAML when `run:` blocks contain Unicode characters
### Pre-submission checks

### zizmor version

1.22.0

### Expected behavior

zizmor should apply autofixes correctly regardless of whether a `run:` block contains Unicode characters. When rewriting expressions inside a `run:` block, any additional YAML mappings introduced by the autofix (such as `env:`) must be inserted as proper YAML keys on a new line.

The presence of multi-byte UTF-8 characters (e.g. `✓`) in the block scalar should not break the auto-fix.

### Actual behavior

When running zizmor with `fix=all` on this file:

```yaml
name: Repro

on:
  issue_comment:
    types: [created]

jobs:
  repro:
    runs-on: ubuntu-latest
    steps:
      - shell: bash
        run: |
          echo "✓ ${{ github.event.comment.body }}"

      - shell: bash
        run: echo ok
```

it produces the following broken yaml file:

```yaml
name: Repro

on:
  issue_comment:
    types: [created]

jobs:
  repro:
    runs-on: ubuntu-latest
    steps:
      - shell: bash
        run: |
          echo "✓ ${GITHUB_EVENT_COMMENT_BODY}"        env:
          GITHUB_EVENT_COMMENT_BODY: ${{ github.event.comment.body }}

      - shell: bash
        run: echo ok
```

### Reproduction steps

Run `zizmor --fix=all` on the example from above.

### Logs

```plain text

```

### Additional context

_No response_

ManuelLerchnerQC:
## Pre-submission checks

Please check these boxes:

      one first).

## Summary

Fixes a bug in `yamlpatch` where `MergeInto` operations would insert content inline instead of on a new line when Unicode characters were present.

The root cause is a mismatch between byte offsets and character indices. When multi-byte UTF-8 characters (like `✓`) appear before the insertion point, the check looks at the wrong byte position, causing the newline detection to fail.

Closes #1580

## Test Plan

Added a unit test that reproduces the issue with a Unicode character in a run command. The test verifies that the `env:` block is correctly inserted on a new line rather than inline.

### DIFF
--- a/docs/release-notes.md
+++ b/docs/release-notes.md
@@ -33,6 +33,11 @@ of `zizmor`.
 * Fixed a bug in `yamlpath` where anchor prefixes in sequences and mapping
   were not stripped during path queries (#1562)
 
+* Fixed a bug where "merge into" autofixes would produce incorrect patches
+  in the presence of multi-byte Unicode characters (#1581)
+
+    Many thanks to @ManuelLerchnerQC for implementing this fix!
+
 ## 1.22.0
 
 ### Changes ⚠️


woodruffw (commented):
Thanks @ManuelLerchnerQC, this looks good to me! For good measure, could you add your repro from the original issue as a testcase for the template-injection audit? It should fit in well with the existing unit tests.

woodruffw:
Thanks again @ManuelLerchnerQC!

PR #1597, opened 2026-02-07
Fix for static evaluation fixes

### LINKED ISSUE #1578: [BUG]: `obfuscation` auto-fix failure
### Pre-submission checks

### zizmor version

zizmor 1.22.0

### Expected behavior

I have a github workflow that runs release-please - a piece of infrastructure to automate away toil from releases. I have a monorepo, so there are packages at arbitrary levels of directory depth.

This necessitates defining job-level outputs like below, with single-quoted interpolations:

```yaml
    outputs:
      iac/terraform/attribution.tfm--release_created: ${{ 'steps.release.outputs.iac/terraform/attribution.tfm--release_created' }}
      iac/terraform/attribution.tfm--tag_name: ${{ 'steps.release.outputs.iac/terraform/attribution.tfm--tag_name' }}
```

zizmor should not flag these as `obfuscations` as without the single-quotes, the workflow fails to parse by GHA.

### Actual behavior

zizmor flags these as obfuscations. Regular console output:

```console
help[obfuscation]: obfuscated usage of GitHub Actions features
    --> .github/workflows/release_please.yaml:58:59
     |
  58 |       iac/terraform/attribution.tfm--release_created: ${{ 'steps.release.outputs.iac/terraform/attribution.tfm--release_created' }}
     |                                                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ can be replaced by its static evaluation
     |

...
  46 findings (5 ignored, 27 suppressed, 14 fixable): 0 informational, 14 low, 0 medium, 0 high

  Fix Summary
  Failed to apply 14 fixes:

...
    obfuscation: .github/workflows/release_please.yaml (conflict after applying previous fixes: fix failed: Invalid operation: no match for 'Subfeature { after: 2342, fragment: Raw("'steps.release.outputs.iac/terraform/attribution.tfm--release_created'") }' in feature)
```

### Reproduction steps

A github workflow like:

```yaml
jobs:
  release-please:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd  # v6.0.2
        with:
          fetch-depth: 0  # ... because release-please scans historical commits to build releases, so we need all the history.
          persist-credentials: false
      - id: release
        uses: ./vendor/github.com/googleapis/release-please-action
        with:
          config-file: "tools/releasing/config.release-please.json"
          manifest-file: "tools/releasing/manifest.release-please.json"
          target-branch: "${{ inputs.rp_target_branch }}"
      - name: Print release outputs for debugging
        continue-on-error: true
        run: |-  # zizmor: ignore[template-injection] This is only going to JSONify the contents, not do anything with it.
          echo "Release outputs:"
          echo '${{ toJson(steps.release.outputs) }}'
    outputs:
      iac/terraform/attribution.tfm--release_created: ${{ 'steps.release.outputs.iac/terraform/attribution.tfm--release_created' }}
      iac/terraform/attribution.tfm--tag_name: ${{ 'steps.release.outputs.iac/terraform/attribution.tfm--tag_name' }}
```

Then `zizmor --fix=all ".github/workflows/release_please.yaml"`

### Logs

```plain text

```

### Additional context

(Perhaps there's a different way I should define such outputs, but I haven't found one)

I worked around this by the following zizmor configuration:

```yaml
rules:
  obfuscation:
    ignore:
      - release_please.yaml
```

woodruffw:
This actually fixes two bugs:

* We were previously incorrectly fixing `${{ 'abc' }}` as `${{ abc }}`, rather than an *unfenced* `abc`.
* We would previously fail to apply a fix using `RewriteFragment` if the patch's route was empty (i.e. indicated the entire document) *and* the document had leading whitespace/newlines.

~~Needs more tests.~~

Fixes #1578.

### DIFF
--- a/docs/release-notes.md
+++ b/docs/release-notes.md
@@ -44,6 +44,12 @@ of `zizmor`.
 * Fixed a bug where the [template-injection] audit would produce
   duplicated pedantic-only findings (#1589)
 
+* Fixed a bug where the [obfuscation] audit would produce incorrect
+  autofixes for a subset of constant-reducible expressions (#1597)
+
+* Fixed a bug where the [obfuscation] audit would fail to apply fixes
+  to a subset of inputs with leading whitespace (#1597)
+
 ## 1.22.0
 
 ### Changes ⚠️


crates/zizmor/tests/integration/test-data/obfuscation/issue-1578.yml
woodruffw, line 1: Noting for myself: the weird spacing here is intentional, it causes the autofix to fail.

crates/zizmor/src/audit/obfuscation.rs
```diff
@@ -367,6 +373,58 @@ mod tests {
         fixed_document.source().to_string()
     }
 
+    #[tokio::test]
+    async fn test_obfuscation_fix_static_evaluation() {
+        let workflow_content = r#"name: Test Workflow
```
woodruffw: Noting for myself: the lack of a leading newline here is load-bearing; the autofix fails if there's a newline at the top of the file. This is clearly wrong.

woodruffw: I'm pretty sure the problem here is with empty routes, i.e. those that point to the document root -- logically this should correspond to a span of the entire input source, but it probably _actually_ corresponds to a span of the top feature, which really begins at whatever offset is after the leading whitespace.

woodruffw: Yeah, that was it. Turns out this is the same bug as above, which is nice.

crates/zizmor/src/audit/obfuscation.rs
```diff
@@ -274,13 +271,19 @@ impl Audit for Obfuscation {
...
                 if parsed.constant_reducible() {
+                    // If the entire expression is constant reducible we need to replace the whole thing,
+                    // including its fencing, to avoid leaving behind a semantically different fenced expression.
+                    // For example, `${{ 'foo' }}` is equivalent to `foo`, but if we only replaced the inner
+                    // expression we'd end up with `${{ foo }}`, which is not.
+
                     // Get the main expression's origin from the first annotation
-                    if let Some((_, main_origin, _)) = obfuscated_annotations.first()
+                    // TODO: Remove this check?
+                    if let Some((_, _, _)) = obfuscated_annotations.first()
```
woodruffw: Need to think about this harder, but I think this is fine to remove now.

PR #1598, opened 2026-02-07
Bump to latest http-cache alpha

woodruffw:
See https://github.com/06chaynes/http-cache/issues/141.

This is not a full solution yet, but gets closer to not failing on older versions of zizmor's caches.


woodruffw:
#1600

PR #1599, opened 2026-02-09
Add secrets-outside-env

### LINKED ISSUE #184: Audit idea: Secrets outside of environments
### Discussed in https://github.com/woodruffw/zizmor/discussions/181

<sup>Originally posted by **AndrewMohawk** November 19, 2024</sup>
Loving the tool so far! I'd love to have zizmor also determine secrets referenced outside of environments within GH actions. I see it fairly often and it means that if anyone has the ability to create a PR with an updated workflow they can exfil secrets that arent tied to an environment (and that environment having branch protections). The caveat here is that an attacker would have had to submit a PR before the malicious one (usually you would see them as doing typo fixes first). </div>

woodruffw:
This adds a `secrets-outside-env` audit that finds `secrets` context accesses in jobs that aren't guarded by a dedicated environment. The comments in the audit describe some sources of false negatives with this approach.

Needs tests and docs.

Closes #184. Closes #637.

### DIFF
--- a/docs/release-notes.md
+++ b/docs/release-notes.md
@@ -9,6 +9,11 @@ of `zizmor`.
 
 ## Next (UNRELEASED)
 
+### New Features 🌈
+
+* **New audit**: [secrets-outside-env] detects usage of the `secrets` context
+  in jobs that don't have a corresponding `environment` (#1599)
+
 ### Enhancements 🌱
 
 * `zizmor`'s LSP mode is now configuration-aware, and will load
@@ -1488,5 +1493,6 @@ This is one of `zizmor`'s bigger recent releases! Key enhancements include:
 [archived-uses]: ./audits.md#archived-uses
 [impostor-commit]: ./audits.md#impostor-commit
 [misfeature]: ./audits.md#misfeature
+[secrets-outside-env]: ./audits.md#secrets-outside-env
 
 [exit code]: ./usage.md#exit-codes


docs/audits.md
```diff
@@ -1304,6 +1304,65 @@ that explicitly forwards each secret actually needed by the reusable workflow.
...
+
+### Remediation
+
+In general, secrets should be configured at the environment level, and only
+the job or jobs that need a secret should use the corresponding environment.
+
+!!! important
+
+    You **must** move your secrets into the environment's secrets (and remove
+    them from the repo/org-wide secrets) in order for this to be effective.
```
lazka, line 1337: I was a bit confused about how zizmor knows the source after fixing things, something like this would have helped me:

> Zizmor will assume that secrets are environment secrets if the job references an
environment. It cannot actually see where secrets are stored in GitHub's settings.

but ymmv

woodruffw: Thanks! Yeah, zizmor doesn't know the source (it could, but not without expanding its token requirements significantly). I'll tweak the docs.

(More broadly, I'm kind of low-confidence on the value of this audit in the regular persona. If you're finding it annoying, that would be very valuable to know so I can evaluate lowering it to pedantic!)

lazka: I don't mind it personally, but I don't have to deal with that many secrets (also since trusted publishing, thanks ;) )

woodruffw: Good to know! #1740 will hopefully clarify this a bit 🙂

PR #1609, opened 2026-02-12
Detect and report GitHub API rate limiting

### LINKED ISSUE #1252: CLI: improve error message on rate limiting from GitHub
From @p-linnane:

```
Run zizmor --format sarif . > results.sarif
+ zizmor --format sarif .
🌈 zizmor v1.15.2
 WARN collect_inputs: zizmor::registry::input: failed to validate input as dependabot config: input does not match expected validation schema
fatal: no audit was performed
impostor-commit failed on file://./.github/workflows/actionlint.yml

Caused by:
    0: couldn't list tags for actions/checkout
    1: request error while accessing GitHub API
    2: HTTP status client error (403 Forbidden) for url (https://api.github.com/repos/actions/checkout/tags?page=0&per_page=100)
```

This is a pretty confusing error message for us to present -- the top-level error makes it sound like the problem was that the input was malformed, when the actual error is a request error (an HTTP 403).

dguido:
## Pre-submission checks

## Summary

- Add `RateLimitMiddleware` (implements `reqwest_middleware::Middleware`) that detects HTTP 403 + `x-ratelimit-remaining: 0` and converts to a dedicated `ClientError::RateLimited` error
- Show a clear error message with actionable help: wait and retry, check GH_TOKEN, or use --offline
- Narrow retry classifier to only retry 5xx/429/408 (previously retried all 4xx including 403)
- Uses middleware approach as suggested by @woodruffw; checks `x-ratelimit-remaining: 0` to distinguish rate-limit 403s from other 403s (auth failures, etc.)

Closes #1252

## Test Plan

- [x] `cargo build` — clean, zero warnings
- [x] `cargo test -p zizmor` — 194 passed, 0 failed
- [x] `cargo fmt --check` — clean
- [x] `cargo clippy -p zizmor` — clean (pre-existing test-only clippy issues unchanged)

### DIFF
--- a/crates/zizmor/src/audit/mod.rs
+++ b/crates/zizmor/src/audit/mod.rs
@@ -388,3 +388,23 @@ pub(crate) trait Audit: AuditCore {
         Ok(results)
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use crate::audit::AuditError;
+    use crate::github::ClientError;
+
+    #[test]
+    fn test_downcast_ref_recovers_rate_limit_error() {
+        let client_error = ClientError::RateLimited {
+            reset: "~5 minutes".into(),
+        };
+        let audit_error = AuditError::new("test-audit", client_error);
+        let downcasted = audit_error.downcast_ref::<ClientError>();
+        assert!(downcasted.is_some());
+        assert_eq!(
+            downcasted.unwrap().rate_limit_reset(),
+            Some("~5 minutes"),
+        );
+    }
+}


crates/zizmor/src/github.rs
```diff
@@ -228,6 +265,69 @@ impl<T: CacheManager> reqwest_middleware::Middleware for ChainedCache<T> {
...
+
+        if res.status() != StatusCode::FORBIDDEN {
+            return Ok(res);
+        }
+
+        let is_rate_limited = res
+            .headers()
+            .get("x-ratelimit-remaining")
+            .and_then(|v| v.to_str().ok())
+            .is_some_and(|v| v == "0");
```
woodruffw: This will panic if the server sends a malformed header, e.g. `x-ratelimit-remaining: <invalid-utf8>`. It's probably fine to just compare for the literal byte string `b"0"` here instead.

(Or we could percolate an error on an invalid header, but that seems overkill.)

crates/zizmor/src/github.rs
```diff
@@ -228,6 +265,69 @@ impl<T: CacheManager> reqwest_middleware::Middleware for ChainedCache<T> {
...
+
+        if !is_rate_limited {
+            return Ok(res);
+        }
+
+        let reset = res
+            .headers()
+            .get("x-ratelimit-reset")
+            .and_then(|v| v.to_str().ok())
+            .and_then(|v| v.parse::<u64>().ok())
```
woodruffw, line 320: Same thing re: panics. I think an appropriate failure mode here would be fall back to `"unknown"` here too.

crates/zizmor/src/github.rs
```diff
@@ -228,6 +265,69 @@ impl<T: CacheManager> reqwest_middleware::Middleware for ChainedCache<T> {
...
+        req: reqwest::Request,
+        extensions: &mut http::Extensions,
+        next: reqwest_middleware::Next<'_>,
+    ) -> reqwest_middleware::Result<Response> {
+        let res = next.run(req, extensions).await?;
+
+        if res.status() != StatusCode::FORBIDDEN {
+            return Ok(res);
+        }
+
```
woodruffw, line 306: I think we should also probably check the origin here, and avoid trying to apply any rate-limit header checks if the host doesn't match the GitHub API host. That'll make it harder for us to panic on differently-shaped headers if we start to reuse our client stack for other requests.

(This matters because some servers like to put relative durations in `x-ratelimit-reset` and similar, which we'd handle incorrectly below.)

crates/zizmor/src/github.rs
```diff
@@ -973,4 +1088,137 @@ mod tests {
...
+    #[test]
+    fn test_rate_limit_reset_through_middleware() {
+        let inner = ClientError::RateLimited {
+            reset: "~5 minutes".into(),
+        };
+        let err = ClientError::Middleware(reqwest_middleware::Error::Middleware(inner.into()));
+        assert_eq!(err.rate_limit_reset(), Some("~5 minutes"));
+    }
+
+    #[test]
```
woodruffw: The "through" tests below seem pretty duplicative; I think they should just be removed.

crates/zizmor/src/github.rs
```diff
@@ -973,4 +1088,137 @@ mod tests {
...
+        assert_eq!(err.rate_limit_reset(), None);
+    }
+
+    #[test]
+    fn test_rate_limit_reset_none_for_non_client_middleware_error() {
+        let err = ClientError::Middleware(reqwest_middleware::Error::Middleware(anyhow::anyhow!(
+            "some other middleware error"
+        )));
+        assert_eq!(err.rate_limit_reset(), None);
+    }
```
woodruffw, line 1137: Safe to remove too.

crates/zizmor/src/github.rs
```diff
@@ -973,4 +1088,137 @@ mod tests {
...
+
+    #[test]
+    fn test_rate_limit_reset_through_nested_inner() {
+        let rate_limited = ClientError::RateLimited {
+            reset: "~2 minutes".into(),
+        };
+        let inner = ClientError::Inner(Arc::new(rate_limited));
+        let outer = ClientError::Inner(Arc::new(inner));
+        assert_eq!(outer.rate_limit_reset(), Some("~2 minutes"));
+    }
```
woodruffw, line 1195: Safe to remove IMO.

crates/zizmor/src/main.rs
```diff
@@ -959,11 +959,52 @@ async fn main() -> ExitCode {
...
+                let group =
+                    Group::with_title(Level::ERROR.primary_title("GitHub API rate limit exceeded"))
+                        .element(
+                            Level::HELP.message(format!(
+                                "rate limit resets in {reset}; wait and try again"
+                            )),
+                        )
+                        .element(Level::HELP.message(
+                            "ensure GH_TOKEN is set to a valid token for higher rate limits",
+                        ))
```
woodruffw: This is probably misleading, since zizmor doesn't access GitHub's API unless it has a valid token. What's actually happened in this case is that someone has exceeded their *authenticated* rate limit, so we should probably just limit the message to telling them to wait.

crates/zizmor/src/main.rs
```diff
@@ -959,11 +959,52 @@ async fn main() -> ExitCode {
                 fatal = "fatal".red().bold()
             );
 
-            let report = match &err {
+            // Check all error paths that can carry a rate-limit error.
+            // NOTE: Error::Audit wraps errors through anyhow, making
```
woodruffw: I think this limitation makes this PR unmergeable, at least for now -- most API operations happen during audit steps, so a ratelimit is overwhelmingly likely to be hit during the audit phase. A larger refactor isn't out of the question, but I think it'd need to be human-driven.

woodruffw (changes requested):
Mostly minor feedback, but there's one thing here that hard-blocks this approach (not being able to actually pull the ratelimit error out in the most common case).

crates/zizmor/src/github.rs
dguido: Done — switched to `v.as_bytes() == b"0"` which avoids the `to_str()` entirely.

crates/zizmor/src/github.rs
dguido: Done — the `x-ratelimit-reset` header now uses `.and_then(|v| v.to_str().ok())` followed by `.and_then(|v| v.parse::<u64>().ok())`, falling back to `"unknown"` via `.unwrap_or_else(|| "unknown".into())` if either step fails.

crates/zizmor/src/github.rs
dguido: Done — `RateLimitMiddleware` now takes an `api_host: String` field (set from the client's configured API URL at construction) and short-circuits with `next.run()` if `req.url().host_str()` doesn't match.

crates/zizmor/src/github.rs
dguido: Removed `test_rate_limit_reset_through_list_tags` and `test_rate_limit_reset_through_list_branches`. Kept `test_rate_limit_reset_through_middleware` since it covers the one wrapping layer the middleware actually produces.

crates/zizmor/src/github.rs
dguido: Removed.

crates/zizmor/src/github.rs
dguido: Removed.

crates/zizmor/src/main.rs
dguido: Good call — removed the "ensure GH_TOKEN is set" help line. The message now just says the reset time and to wait and try again.

crates/zizmor/src/main.rs
dguido: Addressed — `AuditError` now has a `downcast_ref` method that reaches through the `anyhow::Error` wrapper to recover the original `ClientError`. The `Error::Audit` arm in `main.rs` uses it: `source.downcast_ref::<ClientError>().and_then(|ce| ce.rate_limit_reset())`.

This turned out to be simpler than expected because `anyhow::Error` already provides `downcast_ref<T>()` — it preserves the concrete type behind its type-erased storage. So rather than needing to restructure the error hierarchy or change how audit errors propagate, we just needed a one-line delegation method on `AuditError` that forwards to `self.source.downcast_ref::<T>()`. No changes to the `Audit` trait, no changes to how individual audits construct errors, and the existing `AuditCore::err()` constructor already wraps via `Into<anyhow::Error>` which preserves the concrete type.

Added a test in `audit/mod.rs` that round-trips a `ClientError::RateLimited` through `AuditError::new` → `downcast_ref` → `rate_limit_reset()` to verify the chain works.

PR #1611, opened 2026-02-12
feat: support `-` for reading input from stdin

### LINKED ISSUE #1174: CLI: support `-` for reading from stdin
### Pre-submission checks

### zizmor version

zizmor 1.13.0

### Expected behavior

When a file (device or not) is passed to zizmor, zizmor should read the file and process it

### Actual behavior

zizmor treats it as a repo slug (just like `zizmor zizmorcore/zizmor` would)

### Reproduction steps

1. Run `zizmor /path/to/character/file` (like `/dev/sda`, if run in a terminal stdin should be `/dev/pts/X` or `/dev/ttyX` so `zizmor /proc/self/fd/0` should be fine too)
2. Observe that zizmor, instead of reading the file and trying to process it, treats the path as a repo slug

### Logs

```plain text
❯ zizmor --verbose /dev/sda
fatal: no audit was performed
error: can't retrieve repository: /dev/sda
  |
  = help: try removing --offline or passing --gh-token <TOKEN>
❯ zizmor --verbose /proc/self/fd/0 # it's a symlink to /dev/pts/3
fatal: no audit was performed
error: can't retrieve repository: /proc/self/fd/0
  |
  = help: try removing --offline or passing --gh-token <TOKEN>
```

### Additional context

While I wouldn't expect someone to store a workflow in a raw disk, it can be useful to make zizmor accept a workflow from stdin using the aforementioned /dev/pts (since `-` is not supported).

The code that incorrectly categorizes a path as a slug is: 
https://github.com/zizmorcore/zizmor/blob/896470ed028618e1c01c76fef53c3e85dc18d6c9/crates/zizmor/src/registry/input.rs#L492-L498
I would alter the behavior to first check if the path is a directory (invoking collect_from_file if that's the case), checking if it exists (instead of if it's a file, and invoking collect_from_file) and then assume that if a string is not a valid path then it must be a slug.

dguido:
## Pre-submission checks

## Summary

- Add support for `zizmor -` to read a single workflow, action, or Dependabot config from stdin
- Input type is automatically inferred by trying each parser in order (workflow → action → dependabot)
- Validates that `-` is the only input and rejects `--fix` with stdin
- Uses existing `InputKey::Local` with synthetic path `<stdin>` to minimize churn; no local config discovery for stdin (uses global config or defaults since there's no filesystem path)

Closes #1174

## Test Plan

- [x] `echo 'on: push ...' | zizmor -` parses as workflow
- [x] `echo 'name: ... runs: ...' | zizmor -` parses as action
- [x] `echo 'version: 2 ...' | zizmor -` parses as dependabot
- [x] `zizmor - ./other-input` fails with clear error
- [x] `zizmor --fix -` fails with clear error
- [x] Invalid YAML on stdin produces helpful errors (strict and non-strict)
- [x] All 201 existing tests continue to pass

### DIFF
--- a/docs/release-notes.md
+++ b/docs/release-notes.md
@@ -9,6 +9,10 @@ of `zizmor`.
 
 ## Next (UNRELEASED)
 
+### New Features 🌈
+
+* `zizmor` now allows users to audit from stdin, by passing `zizmor -` (#1611)
+
 ## 1.23.1
 
 ### Bug Fixes 🐛


crates/zizmor/src/registry/input.rs
```diff
@@ -549,11 +550,75 @@ impl InputGroup {
...
     }
 
+    async fn collect_from_stdin(options: &CollectionOptions) -> Result<Self, CollectionError> {
+        let mut contents = String::new();
+        std::io::stdin()
+            .read_to_string(&mut contents)
+            .map_err(CollectionError::Io)?;
+
+        // No local config discovery for stdin; use global config or defaults.
+        let config = Config::discover(options, || async { Ok(None) }).await?;
```
woodruffw: This should probably just be `Config::default()` since that's what `::discover` falls back to when you give it a no-op discovery function.

crates/zizmor/src/registry/input.rs
```diff
@@ -549,11 +550,75 @@ impl InputGroup {
...
+                }
+            }
+        }
+
+        // All types failed. In strict mode, report the first error
+        // (Workflow, the most common stdin use case) rather than the
+        // last, so users get relevant diagnostics.
+        if options.strict {
+            return Err(first_err.expect("at least one kind was tried"));
+        }
```
woodruffw: Instead of a loop here, I think this could be done with error combinators. That would be more idiomatic and would allow us to remove the infallible `expect`.

crates/zizmor/src/registry/input.rs
```diff
@@ -549,11 +550,75 @@ impl InputGroup {
...
+        let mut contents = String::new();
+        std::io::stdin()
+            .read_to_string(&mut contents)
+            .map_err(CollectionError::Io)?;
+
+        // No local config discovery for stdin; use global config or defaults.
+        let config = Config::discover(options, || async { Ok(None) }).await?;
+
+        let mut group = Self::new(config);
+        let key = InputKey::local(Group::from("-"), "<stdin>", None);
```
woodruffw: I actually think we should have a new `InputKey` variant here, since `InputKey::Local` assumes that it's given a real path and not `<stdin>`. I don't think that assumption is stressed anywhere at the moment but it'd be ideal to preserve it.

crates/zizmor/tests/integration/cli.rs
```diff
@@ -38,3 +38,207 @@ fn test_gh_token_github_token_conflict() -> anyhow::Result<()> {
...
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+";
+    // NOTE: We use .args(["-"]) instead of .input("-") because the
+    // test harness replaces all occurrences of the input string in the
+    // output, and `-` would corrupt arrows, flags, etc.
+    insta::assert_snapshot!(zizmor().stdin(workflow).no_config(true).args(["-"]).run()?);
```
woodruffw: This and all the other snapshots should be inline, rather than detached.

woodruffw (changes requested):
Some things that need addressing, but this looks pretty reasonable overall. It also needs updates to the usage docs to note that `-` will become a valid input.

crates/zizmor/src/registry/input.rs
dguido: Replaced with `Config::default()`.

crates/zizmor/src/registry/input.rs
dguido: Replaced the loop with sequential `match`/`if let` tries (workflow → action → dependabot). The workflow error is now bound directly as `workflow_err`, removing the `first_err` Option and the infallible `expect`.

crates/zizmor/src/registry/input.rs
dguido: Added an `InputKey::Stdin(StdinKey)` variant. Updated `Display`, `sarif_path`, `presentation_path`, `filename`, and `group`, plus all pattern-match sites in the model files, `fix.rs`, and `registry.rs`.

crates/zizmor/tests/integration/cli.rs
dguido: Converted all 8 detached snapshots to inline and deleted the `.snap` files. Also updated usage docs to document `-` as a valid input.

crates/zizmor/src/registry/input.rs
```diff
@@ -549,11 +570,71 @@ impl InputGroup {
...
+        }
+
+        if let Ok(()) = group.register(InputKind::Dependabot, contents, key, true) {
+            return Ok(group);
+        }
+
+        // All types failed. In strict mode, report the workflow error
+        // (the most common stdin use case) so users get relevant diagnostics.
+        if options.strict {
+            return Err(workflow_err);
```
woodruffw: I think this is going to result in user confusion: what we should do is collect each of the errors and show all of them in the final error, with appropriate context.

(Otherwise a user who tries to audit a slightly-wrong action definition will get a completely unrelated error for an invalid workflow, which will mislead them.)

crates/zizmor/src/registry/input.rs
```diff
@@ -549,11 +570,71 @@ impl InputGroup {
...
+        std::io::stdin()
+            .read_to_string(&mut contents)
+            .map_err(CollectionError::Io)?;
+
+        let mut group = Self::new(Config::default());
+        let key = InputKey::stdin();
+
+        // Infer the input type by trying each parser in order.
+        // Workflow is tried first since it's the most common stdin use case.
+        // We always use strict=true so schema mismatches surface as errors,
```
woodruffw: This says we're always using `strict: true`, but the code doesn't actually do that. But also I think it's fine for us to not always use `strict: true`, since in stdin mode we'll always collect exactly 0 or 1 input anyways, and in the 0 case zizmor will fail because no inputs were collected.

crates/zizmor/src/main.rs
```diff
@@ -780,6 +782,28 @@ async fn run(app: &mut App) -> Result<ExitCode, Error> {
 
     eprintln!("🌈 zizmor v{version}", version = env!("CARGO_PKG_VERSION"));
 
+    // Validate stdin input constraints: `-` must be the only input,
+    // and cannot be combined with `--fix`.
+    if app.inputs.iter().any(|i| i == "-") {
```
woodruffw, line 799: Nitpick, more idiomatic:

```suggestion
    if app.inputs.contains("-") {
```

crates/zizmor/tests/integration/cli.rs
```diff
@@ -38,3 +38,333 @@ fn test_gh_token_github_token_conflict() -> anyhow::Result<()> {
...
+";
+    let output = zizmor()
+        .stdin(workflow)
+        .no_config(true)
+        .args(["--format=sarif", "-"])
+        .run()?;
+
+    // Verify the output is valid JSON and contains <stdin> as the artifact.
+    assert!(output.contains("\"artifactLocation\""));
+    assert!(output.contains("<stdin>"));
```
woodruffw: This should be an inline snapshot like the other integration tests.

docs/usage.md
```diff
@@ -40,6 +41,31 @@ There are three input sources that `zizmor` knows about:
...
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+EOF
+```
+
+When reading from stdin, `zizmor` automatically infers the input type
+(workflow, action, or Dependabot config) by trying each parser in order.
```
woodruffw:
```suggestion
(workflow, action, or Dependabot config).
```

woodruffw (approved):

Issue #827, opened 2025-05-19
New audit: old(er) runs-on


woodruffw:
This is another pedantic/auditor-only audit most likely: we should flag when a job's `runs-on` references a pretty old image, e.g. `uses: ubuntu-20.04`.

Ideally we'd do this via a service like <https://endoflife.date>, if they have a dataset for this.

andrewpollack:
Hey @woodruffw , I'd love to take a shot at this!

I see https://github.com/endoflife-date/endoflife.date/issues/7519 has someone looking into it. I'll take the time until that is resolved to get more familiar developing on zizmor 🙏

woodruffw:
Awesome, thanks a ton @andrewpollack! Don't hesitate to ask if you have any questions about the codebase!

kypherk:
Hi @woodruffw,
I can begin working on this if it's still relevant.
endoflife.date provides data on this now: https://endoflife.date/github-actions-runner-images

kypherk:
@woodruffw Is this a good issue for me to begin working on?

woodruffw:
@andrewpollack originally volunteered to work on this, so I'd like him to confirm he's no longer planning for it (no worries either way, but that keeps the review a bit easier for me).

woodruffw:
(Separately, regardless of who works on this, I'd love to hear your design for how you're planning on approaching this: are you going to make this an online audit, or will we cache the EOL data as part of zizmor's binary, etc.?)

andrewpollack:
Apologies, life has lifed, so I cannot work on this. Another person can feel free to pick this up!

woodruffw:
No worries, thanks @andrewpollack! @kypherk can you share what you're envisioning here?

kypherk:
I don't think an online audit is necessary. Runner deprecation announcements are infrequent and zizmor releases frequently enough to stay in sync. 
I'm thinking something similar to the archived-uses audit where the EOL data is stored in crates/zizmor/data, and this new audit can check all of the `uses:` statements against it.

With this approach I think it'd be ideal to automatically keep the JSON in sync with endoflife.date, perhaps using github actions?

PR #1613, opened 2026-02-12
Add stale-runner audit

### LINKED ISSUE #1612: New audit: stale runners
I hit this pretty often; it's be nice to have an audit that flags `ubuntu22.04`, Windows server 2022, etc.

Many self-hosted runners (e.g. Depot) use similar patterns as well, so we could probably match those.

dguido:
## Pre-submission checks

## Summary

- Adds a new `stale-runner` audit that flags workflows using deprecated or removed GitHub Actions runner images
- Classifies runners as **removed** (High severity: `ubuntu-18.04`, `ubuntu-20.04`, `macos-11`, `macos-12`, `macos-13`, `windows-2019`) or **stale** (Medium severity: `ubuntu-22.04`, `macos-14`, `windows-2022`)
- Handles literal `runs-on` labels, expressions with matrix expansions, and bare expressions in label position

Closes #1612

## Test Plan

- [x] Tests for removed runners (`ubuntu-20.04`, `macos-12`, `windows-2019`) — 3 High severity findings
- [x] Tests for stale runners (`ubuntu-22.04`, `windows-2022`) — 2 Medium severity findings
- [x] Tests for current runners (`ubuntu-latest`, `ubuntu-24.04`, `macos-latest`, `windows-latest`) — no findings
- [x] Tests for matrix expansion containing stale runner — detects `ubuntu-22.04` in matrix
- [x] Updated existing `template_injection::test_issue_22_repro` snapshot (now also flags `macos-14` in matrix)
- [x] `cargo clippy -p zizmor -- -D warnings` passes
- [x] `cargo fmt -- --check` passes
- [x] Full test suite passes (207 integration + 109 unit)

### DIFF
--- a/crates/zizmor/src/audit/stale_runner.rs
+++ b/crates/zizmor/src/audit/stale_runner.rs
@@ -91,26 +91,44 @@ impl Audit for StaleRunner {
 
             match &job.runs_on {
                 LoE::Literal(RunsOn::Target(labels)) => {
-                    let Some(label) = labels.first() else {
-                        continue;
-                    };
-
-                    if let Some(staleness) = runner_staleness(label) {
-                        results.push(
-                            Self::finding()
-                                .confidence(Confidence::High)
-                                .severity(severity_for(&staleness))
-                                .add_location(
-                                    job.location()
-                                        .primary()
-                                        .with_keys(["runs-on".into()])
-                                        .annotated(annotation_for(&staleness, label)),
-                                )
-                                .build(workflow)?,
-                        );
-                    } else if ExplicitExpr::from_curly(label).is_some() {
-                        // Expression in the first label position — we can't
-                        // resolve it statically, so flag with low confidence.
+                    for label in labels {
+                        if let Some(staleness) = runner_staleness(label) {
+                            results.push(
+                                Self::finding()
+                                    .confidence(Confidence::High)
+                                    .severity(severity_for(&staleness))
+                                    .add_location(
+                                        job.location()
+                                            .primary()
+                                            .with_keys(["runs-on".into()])
+                                            .annotated(annotation_for(&staleness, label)),
+                                    )
+                                    .build(workflow)?,
+                            );
+                        } else if ExplicitExpr::from_curly(label).is_some() {
+                            results.push(
+                                Self::finding()
+                                    .confidence(Confidence::Low)
+                                    .severity(Severity::Medium)
+                                    .add_location(
+                                        job.location()
+                                            .primary()
+                                            .with_keys(["runs-on".into()])
+                                            .annotated(
+                                                "expression may expand to a \
+                                                 stale runner",
+                                            ),
+                                    )
+                                    .build(workflow)?,
+                            );
+                        }
+                    }
+                }
+                LoE::Literal(RunsOn::Group { .. }) => {
+                    // Runner groups are user-managed; nothing to flag.
+                }
+                LoE::Expr(exp) => {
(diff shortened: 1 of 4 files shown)


crates/zizmor/src/audit/stale_runner.rs
```diff
@@ -0,0 +1,184 @@
...
+/// Check whether a runner label is stale or removed.
+fn runner_staleness(label: &str) -> Option<Staleness> {
+    match label {
+        // Removed runners.
+        "ubuntu-18.04" | "ubuntu-20.04" | "macos-11" | "macos-12" | "macos-13" | "windows-2019" => {
+            Some(Staleness::Removed)
+        }
+
+        // Stale (deprecated, approaching end-of-life) runners.
+        "ubuntu-22.04" | "macos-14" | "windows-2022" => Some(Staleness::Stale),
```
woodruffw, line 47: I think this hard-coded approach won't be maintainable over time -- ideally what we'd do here is keep this information periodically updated (checked into the repo) via something like endoflife.date.

Ref: https://github.com/endoflife-date/endoflife.date/issues/7519

(This also made me realize that #1612 is a dupe of #827.)

crates/zizmor/src/audit/stale_runner.rs
```diff
@@ -0,0 +1,184 @@
...
+            let Job::NormalJob(job) = job else {
+                continue;
+            };
+
+            match &job.runs_on {
+                LoE::Literal(RunsOn::Target(labels)) => {
+                    for label in labels {
+                        if let Some(staleness) = runner_staleness(label) {
+                            results.push(
+                                Self::finding()
```
woodruffw, line 97: A future version of this audit will probably need to emit pedantic findings for these, since older runners are still supported by GitHub, just older. So they're not security or reliability risks per se, just something to be aware of.

crates/zizmor/src/audit/stale_runner.rs
```diff
@@ -0,0 +1,184 @@
...
+    StaleRunner,
+    "stale-runner",
+    "uses a stale or removed GitHub Actions runner"
+);
+
+/// Whether a runner label is stale (deprecated) or removed.
+enum Staleness {
+    /// The runner has been removed and is no longer available.
+    Removed,
+    /// The runner is deprecated and approaching end-of-life.
```
woodruffw, line 34: We *probably* shouldn't call these deprecated, since they aren't actually deprecated. Maybe just `stale`?

crates/zizmor/src/audit/stale_runner.rs
woodruffw: Alternatively, a more "timeless" approach here would be try and parse the various runner formats, but I think that'll probably be brittle.

crates/zizmor/src/audit/stale_runner.rs
```diff
@@ -0,0 +1,184 @@
...
+    match label {
+        // Removed runners.
+        "ubuntu-18.04" | "ubuntu-20.04" | "macos-11" | "macos-12" | "macos-13" | "windows-2019" => {
+            Some(Staleness::Removed)
+        }
+
+        // Stale (deprecated, approaching end-of-life) runners.
+        "ubuntu-22.04" | "macos-14" | "windows-2022" => Some(Staleness::Stale),
+
+        _ => None,
```
woodruffw, line 49: Note: a full audit here would also need to extract some common self-hosted runner formats, e.g. Depot's naming scheme.

woodruffw (changes requested):

Issue #935, opened 2025-06-11
Audit idea: TOCTOU PR/branch checks


woodruffw:
GitHub Actions is susceptible to TOCTOU vulnerabilities in a handful of places: `workflow_dispatch` and other "maintainer dispatch" events (like `issue_comment`) where a maintainer initiates an action based on the perceived state of a PR branch.

The basic vulnerability here is:

1. The attacker creates a PR/branch/other symbolic object with a certain (innocent-looking) state;
2. The maintainer performs their dispatched event, e.g. adding a label, leaving a comment, triggering a `workflow_dispatch`
3. The dispatched workflow uses the symbolic object's reference (PR number, branch name)
4. The attacker _races_ the dispatched workflow so that the symbolic object points to something malicious instead (e.g. a new commit on the PR/branch that wasn't there at review time).

Adnan Khan has a clean example of this within Dependabot's CI/CD [on his blog](https://adnanthekhan.com/posts/dependabot-core-toctou-writeup/):

![Image](https://github.com/user-attachments/assets/4e167596-98d0-4899-b6db-c8a5196a82dc)

Here, the vulnerable workflow is triggered via `workflow_dispatch` and takes the PR number as an input, so the attacker can race to push new state onto the PR and end up controlling context/potentially executing code within a privileged trigger.

We should attempt to detect these within `zizmor` as well. Needs some further design thought as well.

woodruffw:
The fix PR also shows one possible mitigation for this: https://github.com/dependabot/dependabot-core/pull/11223

TL;DR: The potentially vulnerable workflow should check not just the PR's approval state, but also the exact commit that was approved. If that commit diverges at time-of-use, the operation shouldn't proceed.

woodruffw:
Thought a bit more about this. I think a subset of this is really easy within zizmor:

* The workflow has `issue_comment` or `workflow_dispatch` or `pull_request_target: labeled`
* The workflow uses a TOCTOU-prone context, e.g. `github.event.issue.number`, `github.ref`, `github.ref_name`, etc.

The latter might be noisy if the workflow only logs the mutable reference instead of actually using it to concretize a SHA, so we'll also need some context sensitivity.

PR #1614, opened 2026-02-12
Flag issue_comment as a dangerous trigger

### LINKED ISSUE #1606: Enhance `dangerous-triggers`: flag `issue_comment` as a dangerous trigger
`dangerous-triggers` currently flags `pull_request_target` and `workflow_run` because they run in the context of the base repository. The `issue_comment` trigger shares this property but is not flagged.

When a workflow triggers on `issue_comment`:

- It runs in the context of the **base repository**, not the fork
- It has access to repository **secrets** (including those not exposed to `pull_request` workflows)
- `GITHUB_TOKEN` has the repository's default permissions (often including `contents: write`)
- On public repos, **any user** who can comment on issues/PRs can trigger the workflow — even users with zero repository permissions

This makes `issue_comment` fundamentally similar to `pull_request_target` in terms of security posture. All three recent AI agent attack disclosures rely on it:

- [Claude Code Action RCE](https://johnstawinski.com/2026/02/05/trusting-claude-with-a-knife-unauthorized-prompt-injection-to-rce-in-anthropics-claude-code-action/) — `issue_comment` trigger for Claude Code
- [PromptPwnd](https://www.aikido.dev/blog/promptpwnd-github-actions-ai-agents) — `issues` and `issue_comment` triggers for multiple AI agents
- A real-world bounty report where `issue_comment` triggered a review agent that checked out PR code and executed an attacker-controlled `npx` binary

Beyond AI agents, `issue_comment` is dangerous for any workflow that:
- Checks out PR code referenced from the comment context
- Runs commands based on comment content
- Has write permissions to the repository

## Nuance on severity

`issue_comment` is more commonly used legitimately than `pull_request_target` — chatops, bot commands, triage automation. This means we need to think about noise.

Some options:

1. Flag at **Medium severity, Medium confidence** (lower than `pull_request_target`'s High/Medium), Regular persona
2. Flag at **High severity, Medium confidence**, Pedantic persona (only shown to users who opt in)
3. Flag only when combined with other risk signals (e.g., checkout steps, `run:` blocks, broad permissions) — but this adds complexity

I'd lean toward option 1: Medium/Medium, Regular persona. It's still an important signal but acknowledges that `issue_comment` has more legitimate uses.

## Proposed change

Add a check to `DangerousTriggers::audit_workflow`. This likely requires adding a `has_issue_comment()` method to the Workflow model (similar to the existing `has_pull_request_target()` and `has_workflow_run()`).

The finding message could be something like: "`issue_comment` trigger runs in the base repository context with access to secrets"

## Remediation guidance

- Restrict who can trigger the workflow using `if: github.event.comment.author_association == 'MEMBER'` or similar conditions
- Use minimal permissions (`permissions: {}` at workflow level, explicit per-job)
- Consider whether `pull_request` (which runs in the fork context) would suffice
- Never check out and execute PR code in an `issue_comment`-triggered workflow
- Be aware that `GITHUB_TOKEN` in `issue_comment` workflows may have broader access than in `pull_request` workflows

## Related

- #935 — TOCTOU; `issue_comment` workflows that reference PR branches are susceptible to race conditions
- #1605 — The proposed `agentic-actions` audit would specifically flag AI agents on `issue_comment` triggers at higher severity

dguido:
## Pre-submission checks

## Summary

- Adds `issue_comment` to the `dangerous-triggers` audit at **Medium severity / Medium confidence** (Regular persona), acknowledging it has more legitimate uses than `pull_request_target` or `workflow_run` while still flagging the security risk
- Adds `has_issue_comment()` method to the Workflow model following existing `has_pull_request_target()` / `has_workflow_run()` pattern
- Enables and populates the previously-stubbed `dangerous_triggers` integration test module with tests for all three trigger types

Closes #1606

## Test Plan

- [x] `issue_comment` flagged as bare event (`on: issue_comment`)
- [x] `issue_comment` flagged with event body (`on: issue_comment: types: [created]`)
- [x] `pull_request_target` still flagged at High severity
- [x] `workflow_run` still flagged at High severity
- [x] Multiple dangerous triggers in one workflow all reported
- [x] Safe triggers (`push`) produce no findings
- [x] Existing `template_injection/issue-418-repro` snapshot updated (uses `issue_comment`)
- [x] Full integration suite passes (204/204)

### DIFF
--- a/crates/zizmor/src/audit/dangerous_triggers.rs
+++ b/crates/zizmor/src/audit/dangerous_triggers.rs
@@ -55,6 +55,24 @@ impl Audit for DangerousTriggers {
                     .build(workflow)?,
             );
         }
+        if workflow.has_issue_comment() {
+            findings.push(
+                Self::finding()
+                    .confidence(Confidence::Medium)
+                    .severity(Severity::Medium)
+                    .add_location(
+                        workflow
+                            .location()
+                            .primary()
+                            .with_keys(["on".into()])
+                            .annotated(
+                                "issue_comment runs in the base repository context \
+                                 with access to secrets",
+                            ),
+                    )
+                    .build(workflow)?,
+            );
+        }
 
         Ok(findings)
     }
--- a/crates/zizmor/src/audit/dangerous_triggers.rs
+++ b/crates/zizmor/src/audit/dangerous_triggers.rs
@@ -1,13 +1,16 @@
+use std::ops::Deref;
 use std::sync::LazyLock;
 
-use github_actions_expressions::{Expr, context::ContextPattern};
+use github_actions_expressions::{
+    Expr, SpannedExpr, call::Call, context::ContextPattern, literal::Literal,
+};
 use github_actions_models::common::If;
 
 use super::{Audit, AuditLoadError, audit_meta};
 use crate::audit::AuditError;
 use crate::config::Config;
 use crate::finding::{Confidence, Finding, Severity, location::Locatable as _};
-use crate::models::workflow::{Job, JobCommon, NormalJob, Workflow};
+use crate::models::workflow::{Job, NormalJob, Workflow};
 use crate::state::AuditState;
 use crate::utils::ExtractedExpr;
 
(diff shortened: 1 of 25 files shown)


woodruffw (changes requested):
I'm vaguely worried this is going to be too noisy/disruptive, given how common `issue_comment` is. I'm not inherently opposed to it, but I think I'd like to add more "crater" tests to understand the impact before merging this.

(An alternative would be to push ahead with an approach like #935.)

dguido:
Thanks for the review. Pushed a new iteration that addresses code quality issues and should also help with the noise concern:

- **Trait contract fix**: Removed the `audit_normal_job` override and inlined the `issue_comment` job-level check into `audit_workflow`. The previous implementation overrode both methods with `audit_workflow` manually re-dispatching to `audit_normal_job`, which goes against the trait docs.

- **Trusted-role validation**: `has_author_association_guard` now validates that `author_association` is compared against a trusted role (`OWNER`, `MEMBER`, `COLLABORATOR`), not just that the context is referenced anywhere in the expression. Previously, a guard like `if: github.event.comment.author_association == 'CONTRIBUTOR'` would suppress the finding even though `CONTRIBUTOR` doesn't restrict execution to trusted actors. This should reduce false negatives without adding noise—it only changes behavior for workflows with weak guards.

- **Duplicate fixture removed**: `issue-comment-no-guard.yml` was byte-identical to `issue-comment-bare.yml` (same code path), so it's been replaced with `issue-comment-weak-guard.yml` which tests the new trusted-role validation (`CONTRIBUTOR` guard correctly produces a finding).

Agreed that crater testing before merge makes sense—happy to wait on that. The trusted-role validation in this iteration should make the audit more precise, which may help with the noise concern.

Issue #1270, opened 2025-10-22
Feature: exact matches in `forbidden-uses`


raboof:
### Pre-submission checks

- [x] I am **not** reporting a bug (crash, false positive/negative, etc). These must be filed via the bug report template.
- [x] I have looked through both the [open](https://github.com/zizmorcore/zizmor/issues?q=is%3Aissue+is%3Aopen+) and [closed](https://github.com/zizmorcore/zizmor/issues?q=is%3Aissue%20state%3Aclosed) issues for a duplicate request.

### What's the problem this feature will solve?

Projects using old versions of actions is a latent risk

### Describe the solution you'd like

I'd like to be able to allowlist only the latest versions of actions

woodruffw:
Hi @raboof, thanks for opening an issue. 

Could you take a look at https://github.com/zizmorcore/zizmor/issues/260 and see if that covers what you're thinking here? To my read these two are duplicates, but there's plenty of design room for constitutes an "old" version of an action.

raboof:
Indeed defining 'old action' is challenging. At the ASF we've put together some tooling to maintain a list of exact trusted tags/commits (https://github.com/apache/infrastructure-actions/blob/main/actions.yml). That also includes some (albeit limited) review of new tags, and defines 'old' as 'X months after the next version became available'. TBH it's a bit too early to tell whether this setup will work out for us, but if it does then we'd prefer 'exact matches' over 'N months'. I appreciate we're a bit of a special case, though - not sure how likely other 'auditor personas' would be to maintain such a list ;).

We're currently feeding that list into the GitHub Actions allowlist configuration, but that's a bit of a blunt tool - if zizmor could #1271 then that'd seem really attractive.

dguido:
PR: https://github.com/zizmorcore/zizmor/pull/1615

The `forbidden-uses` audit already supports exact ref matching via `ExactWithRef` patterns (e.g. `actions/checkout@v4`). The documentation incorrectly stated this wasn't possible. This PR:

- Removes the outdated limitation bullet in the `forbidden-uses` docs that claimed exact matches aren't supported
- Adds a configuration example showing version-pinned allowlists, which is the use case described in this issue

With the current implementation, an allowlist like this will only permit the specified versions and flag everything else:

```yaml
rules:
  forbidden-uses:
    config:
      allow:
        - actions/checkout@v4
        - actions/setup-python@v5
```

PR #1615, opened 2026-02-12
docs: document exact ref matching in forbidden-uses

dguido:
## Pre-submission checks

## Summary

- Remove outdated limitation in `forbidden-uses` docs that claimed exact ref matches aren't supported (they are, via `ExactWithRef` patterns like `actions/checkout@v4`)
- Add a configuration example showing version-pinned allowlists

Refs #1270

## Test Plan

Docs-only change — verified replacement URLs return valid content.

### DIFF
--- a/docs/audits.md
+++ b/docs/audits.md
@@ -764,22 +764,6 @@ for details.
             - github/codeql-action/*
     ```
 
-!!! example
-
-    The following configuration would allow only specific versions of actions,
-    rejecting any other versions. This is useful for maintaining a vetted
-    allowlist of exact action versions:
-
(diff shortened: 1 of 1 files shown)


docs/audits.md
```diff
@@ -768,6 +764,22 @@ for details.
...
+    ```yaml title="zizmor.yml"
+    rules:
+      forbidden-uses:
+        config:
+          allow:
+            - actions/checkout@v4
+            - actions/setup-python@v5
+            - github/codeql-action/init@v3
+    ```
+
```
woodruffw: I don't think this example adds much now that the misleading caveat is removed.

```suggestion
```

Issue #1626, opened 2026-02-17
[BUG]: concurrency-limits audit creates duplicate findings for each job


DarkaMaul:
### Pre-submission checks

- [x] I am **not** filing a feature request. These should be filed via the feature request form instead.
- [x] I have checked the [Troubleshooting Guide](https://docs.zizmor.sh/troubleshooting) for my problem.
- [x] I have looked through both the [open](https://github.com/zizmorcore/zizmor/issues?q=is%3Aissue+is%3Aopen+) and [closed](https://github.com/zizmorcore/zizmor/issues?q=is%3Aissue%20state%3Aclosed) issues for a duplicate report.

### zizmor version

main

### Expected behavior

When a workflow is missing concurrency settings entirely (no workflow-level concurrency), the concurrency-limits audit should produce a single finding that:

- Points to the workflow as the primary location (indicating the root cause)
- Shows all affected jobs as related locations
- Makes it clear that adding workflow-level concurrency would fix all jobs at once

### Actual behavior

The audit produces N separate findings (one per job) that all report "missing concurrency setting". This creates noisy output and doesn't make it clear that a single workflow-level fix would resolve all issues. 

### Reproduction steps

Running `zizmor` on `nuget-push.yml` yields to the following results:

```console
cargo run -- --persona=pedantic crates/zizmor/tests/integration/test-data/use-trusted-publishing/nuget-push.yml

[trimmed for clarity]

help[concurrency-limits]: insufficient job-level concurrency limits
  --> crates/zizmor/tests/integration/test-data/use-trusted-publishing/nuget-push.yml:1:1
   |
 1 | / on: [push]
 2 | |
 3 | | name: use-trusted-publishing
...  |
37 | |         env:
38 | |           NUGET_API_KEY: ${{ steps.login.outputs.NUGET_API_KEY }}
   | |__________________________________________________________________^ missing concurrency setting
   |
   = note: audit confidence → High

help[concurrency-limits]: insufficient job-level concurrency limits
  --> crates/zizmor/tests/integration/test-data/use-trusted-publishing/nuget-push.yml:1:1
   |
 1 | / on: [push]
 2 | |
 3 | | name: use-trusted-publishing
...  |
37 | |         env:
38 | |           NUGET_API_KEY: ${{ steps.login.outputs.NUGET_API_KEY }}
   | |__________________________________________________________________^ missing concurrency setting
   |
   = note: audit confidence → High

[...]

PR #1627, opened 2026-02-17
fix: concurrency-limits reports only 1 finding per worfklow

DarkaMaul:
## Pre-submission checks

Please check these boxes:

      one first). (#1626 )

## Summary

This PR fixes the concurrency-limits audit to report one finding per workflow instead of generating separate findings for each affected job.

### Before
Capture d’écran 2026-02-17 à 11 46 56

### After

Capture d’écran 2026-02-17 à 11 43 11

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -37,7 +37,7 @@ jobs:
         with:
           persist-credentials: false
 
-      - uses: Swatinem/rust-cache@779680da715d629ac1d338a641029a2f4372abb5 # v2.8.2
+      - uses: Swatinem/rust-cache@c19371144df3bb44fab255c43d04cbc2ab54d1c4 # v2.9.1
         with:
           cache-all-crates: true
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,7 +29,7 @@ jobs:
       - name: Format
         run: cargo fmt --check
 
-      - uses: Swatinem/rust-cache@779680da715d629ac1d338a641029a2f4372abb5 # v2.8.2
+      - uses: Swatinem/rust-cache@c19371144df3bb44fab255c43d04cbc2ab54d1c4 # v2.9.1
 
       - name: Lint
         run: cargo clippy -- --deny warnings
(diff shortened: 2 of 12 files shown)


crates/zizmor/src/audit/concurrency_limits.rs
```diff
@@ -102,6 +80,55 @@ impl Audit for ConcurrencyLimits {
...
+                if !jobs_missing_cancel_in_progress.is_empty() {
+                    let mut finding_builder = Self::finding()
+                        .confidence(Confidence::High)
+                        .severity(Severity::Low)
+                        .persona(Persona::Pedantic)
+                        .add_location(
+                            workflow
+                                .location()
+                                .primary()
+                                .with_keys(["on".into()])
```
DarkaMaul, line 94: FLAG: I've iterated with various solutions and this one is the cleanest I've found.

- Flag at the workflow level (but this created interleaved output with job reports)
- Flag at the name key (but this is not a required key)
- Don't flag at the workflow level - only at each job (but I thought the message was not clear enough)

I'm happy for feedback here if you have a better solution.

woodruffw: Yeah, it's kind of unfortunate how hard these kinds of "layered" findings are to represent. 

I think this approach is fine -- adding each affected job to the workflow-level finding is a nice touch.

crates/zizmor/src/audit/concurrency_limits.rs
```diff
@@ -102,6 +80,55 @@ impl Audit for ConcurrencyLimits {
...
+                                .annotated("workflow is missing concurrency setting"),
+                        );
+
+                    for job in &jobs_missing_concurrency {
+                        finding_builder = finding_builder.add_location(
+                            job.location()
```
woodruffw: I _think_ you can use `location_with_grip` here to produce a nicer "grip" location for a job.

(The "grip" location will use the job's name if present, and will fall back to the job's ID if the name isn't present.)

Ref: 

https://github.com/zizmorcore/zizmor/blob/04d270c39f4f00989c8b02d827d090127c867123/crates/zizmor/src/models/workflow.rs#L398-L409

woodruffw (approved):

woodruffw:
Looks like the crater snapshots need updating too, but otherwise LGTM.

woodruffw:
Thanks @DarkaMaul!

woodruffw:
Oh, this still needs the crater test snapshot updates.

Issue #1625, opened 2026-02-16
[BUG]: Incorrect evaluation of number in expressions


DarkaMaul:
### Pre-submission checks

- [x] I am **not** filing a feature request. These should be filed via the feature request form instead.
- [x] I have checked the [Troubleshooting Guide](https://docs.zizmor.sh/troubleshooting) for my problem.
- [x] I have looked through both the [open](https://github.com/zizmorcore/zizmor/issues?q=is%3Aissue+is%3Aopen+) and [closed](https://github.com/zizmorcore/zizmor/issues?q=is%3Aissue%20state%3Aclosed) issues for a duplicate report.

### zizmor version

main

### Expected behavior

Consider the following workflow

```yaml
name: Test
on: push
jobs:
  test:
    runs-on: ubuntu-latest
    steps:
      - if: ${{ '   1   ' == 1 }}
        run: echo test
```

Weirdly enough, the GitHub action expression parser trims the value and convert it (so '   1   ' => '1' => 1)

Proof: https://github.com/trail-of-forks/zizmor/actions/runs/22066419083/workflow

Running `zizmor --fix` on the workflow should yield to:

```yaml
name: Test
on: push
jobs:
  test:
    runs-on: ubuntu-latest
    steps:
      - if: ${{ true }}
        run: echo test
```

### Actual behavior

Current result:

```yaml
name: Test
on: push
jobs:
  test:
    runs-on: ubuntu-latest
    steps:
      - if: ${{ false }}
        run: echo test
```

### Reproduction steps

The error is probably from:

https://github.com/zizmorcore/zizmor/blob/f9f7f9acc07c6e145104a9df855449fbc8df3fbb/crates/github-actions-expressions/src/lib.rs#L743

### Logs

```plain text

```

woodruffw:
Thanks @DarkaMaul, that's definitely a bug. If GHA trims before coercing numbers, we should do the same.

PR #1628, opened 2026-02-17
fix: Fix zizmor handling of numbers in expressions

DarkaMaul:
## Pre-submission checks

Please check these boxes:

      one first) ( #1625)

## Summary

The PR fixes `zizmor` handling of numbers in expression to align it with GitHub Action expression parser.

## Test Plan

This change is self contained with multiple tests. The only problem is that we have no easy way of tracking GitHub changes to their expression parser.

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -39,7 +39,7 @@ jobs:
         with:
           cache-all-crates: true
 
-      - uses: astral-sh/setup-uv@61cb8a9741eeb8a550a1b8544337180c0fc8476b # v7.2.0
+      - uses: astral-sh/setup-uv@eac588ad8def6316056a12d4907a9d4d84ff7a3b # v7.3.0
 
       - name: Build zizmor (release)
         run: cargo build --release
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -44,7 +44,7 @@ jobs:
 
       - uses: Swatinem/rust-cache@779680da715d629ac1d338a641029a2f4372abb5 # v2.8.2
 
-      - uses: astral-sh/setup-uv@61cb8a9741eeb8a550a1b8544337180c0fc8476b # v7.2.0
+      - uses: astral-sh/setup-uv@eac588ad8def6316056a12d4907a9d4d84ff7a3b # v7.3.0
 
       - name: Test dependencies
         run: |
@@ -71,7 +71,7 @@ jobs:
         with:
           persist-credentials: false
 
-      - uses: astral-sh/setup-uv@61cb8a9741eeb8a550a1b8544337180c0fc8476b # v7.2.0
+      - uses: astral-sh/setup-uv@eac588ad8def6316056a12d4907a9d4d84ff7a3b # v7.3.0
 
       - name: Test site
         run: make site
--- a/.github/workflows/codegen.yml
+++ b/.github/workflows/codegen.yml
@@ -63,7 +63,7 @@ jobs:
         with:
           persist-credentials: false
 
-      - uses: astral-sh/setup-uv@61cb8a9741eeb8a550a1b8544337180c0fc8476b # v7.2.0
+      - uses: astral-sh/setup-uv@eac588ad8def6316056a12d4907a9d4d84ff7a3b # v7.3.0
 
       - name: try to refresh context capabilities
         run: |
@@ -101,7 +101,7 @@ jobs:
         with:
           persist-credentials: false
 
-      - uses: astral-sh/setup-uv@61cb8a9741eeb8a550a1b8544337180c0fc8476b # v7.2.0
+      - uses: astral-sh/setup-uv@eac588ad8def6316056a12d4907a9d4d84ff7a3b # v7.3.0
 
       - name: try to refresh CodeQL injection sinks
         run: |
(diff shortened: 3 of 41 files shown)


crates/github-actions-expressions/src/lib.rs
```diff
@@ -763,6 +757,37 @@ impl Evaluation {
...
 }
 
+/// Parse a string into a number following GitHub Actions coercion rules.
+///
+/// The string is trimmed and then parsed as:
+/// - Empty/whitespace-only → 0.0
+/// - `0x` prefix → hexadecimal integer
+/// - `0o` prefix → octal integer
+/// - Otherwise → standard float parsing (decimal, scientific notation,
+///   Infinity, -Infinity, NaN)
```
woodruffw: Hmm, are octals and hex numbers actually supported by GHA? The docs are naturally unclear:

https://docs.github.com/en/enterprise-cloud@latest/actions/reference/workflows-and-actions/expressions#literals

(The docs for `number` say "any JSON number," which would need to be base 10. But then the example immediately below shows a hex number, naturally.)

woodruffw: (If so, there's a possibility we need to update the Pest grammar as well.)

woodruffw: Based on this, I think GitHub actually just uses JS's `Number` ctor:

https://github.com/actions/languageservices/blob/fb5c6e4f27bb1ddf512609a6a341aadd17ce86f3/expressions/src/data/string.ts#L14-L16

...so naturally we have all of the JS coercion rules:

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number#number_coercion

DarkaMaul: I’m on my phone right now but the workflow here shows that it does support octal and hexadecimal : https://github.com/trail-of-forks/zizmor/actions/runs/22103104435/job/63877999605

woodruffw: Gotcha. In that case I think we do (unfortunately) need to ensure our grammar handles those, plus implement the full JS coercion/number literal rules.

woodruffw: I've filed https://github.com/github/docs/issues/43008 for the doc bug on GH's side.

DarkaMaul: > Gotcha. In that case I think we do (unfortunately) need to ensure our grammar handles those, plus implement the full JS coercion/number literal rules.

I think we should _not_ do this (sadly).

If we look at this [run](https://github.com/trail-of-forks/zizmor/actions/runs/22187841727/job/64166647261), they are several differences between `Number` and what GitHub Actions are doing internally.


- binary notation (supported in Number)
```
=== Binary notation support ===
0b1010==10: FAIL
0B1010==10: FAIL
0b0==0: FAIL
0b1==1: FAIL
0b11==3: FAIL
' 0b11 '==3: FAIL
```

- leading zeroes:
```
=== Leading zeros behavior ===
0123 == 123 (decimal): PASS
0123 == 83 (octal): FAIL
00 == 0: PASS
007 == 7: PASS
010 == 10 (decimal): PASS
010 == 8 (octal): FAIL
```

```js
> Number(010)
8
> Number(0123)
83
```

- signed hexa
```
-0xff == -255: FAIL (NaN)
+0xff == 255: FAIL (NaN)
-0o10 == -8: FAIL (NaN)
+0o10 == 8: FAIL (NaN)
-0b11 == -3: FAIL (NaN)
```
```js
> -0xff == -255
true
> +0xff == 255
true
> -0o10 == -8
true
> -0b11 == -3
true
```


- and let's not talk about `NaN` semantic:
```
=== NaN semantics ===
NaN != NaN: false (WRONG)
NaN == NaN: true (WRONG)
```

```js
> NaN != NaN
true
> NaN == NaN
false
```

woodruffw: Ugh, what the hell. I don't really understand _where_ GitHub Actions number semantics come from, then. I think we need to fully understand how their own parser/evaluator works here before we can address this.

Would you mind going through that source tree and trying to understand what they're actually doing?

DarkaMaul: Found it here: https://github.com/actions/runner/blob/main/src/Sdk/Expressions/Sdk/ExpressionUtility.cs#L223

I had an AI extract this function, create a binary, and reproduce the workflow result!

I'm working on aligning our `parse_number_from_str` on their function.

DarkaMaul: I think this new implementation is aligned with the GH Action Runner one, and it matches the behavior of my test run [here](https://github.com/trail-of-forks/zizmor/actions/runs/22220338438)

woodruffw: Thanks! I'll try and take a look today.

woodruffw: Looks good to me, thanks @DarkaMaul! I think the pest grammar itself still needs to be updated too, right now it doesn't allow hex, octal, or exponent forms in the grammar itself.

woodruffw (approved):
Thanks @DarkaMaul, nice work!

Please add a release-notes entry and this'll be good to go 🙂 

(Normally I'd push up on your branch but this is on trail-of-forks, so there's the GitHub org 3P PR bug blocking it.)

Issue #1634, opened 2026-02-19
[BUG]: known-vulnerable-actions can give a confusing error message


rmuir:
### Pre-submission checks

- [x] I am **not** filing a feature request. These should be filed via the feature request form instead.
- [x] I have checked the [Troubleshooting Guide](https://docs.zizmor.sh/troubleshooting) for my problem.
- [x] I have looked through both the [open](https://github.com/zizmorcore/zizmor/issues?q=is%3Aissue+is%3Aopen+) and [closed](https://github.com/zizmorcore/zizmor/issues?q=is%3Aissue%20state%3Aclosed) issues for a duplicate report.

### zizmor version

zizmor 1.22.0

### Expected behavior

The online known-vulnerable-actions began failing for me, in a correct way, but with a confusing error message

I have:
```yaml
 - name: Run Trivy vulnerability scanner in repo mode
   uses: aquasecurity/trivy-action@b6643a29fecd7f34b3597bc6acb0a98b03d33ff8 # 0.33.1
```

This version is apparently vulnerable: https://github.com/aquasecurity/trivy-action/security/advisories/GHSA-9p44-j4g5-cfx5

Error message:
```
fatal: no audit was performed
'known-vulnerable-actions' audit failed on file://./.github/workflows/scanning.yml

Caused by:
    0: error in 'known-vulnerable-actions' audit
    1: Cannot resolve version 0.34.0 to commit hash for aquasecurity/trivy-action
```

### Actual behavior

the audit fails with a bit of a confusing error ("cannot resolve version to commit hash"), it took me some time to track down that the real problem is, my workflow was using a vulnerable version (0.33.1)

### Reproduction steps

Here's the full job:

```yaml
jobs:
  trivy:
    name: Scan for secrets and misconfig
    runs-on: ubuntu-24.04
    timeout-minutes: 15
    permissions:
      contents: read # private repository
      actions: read # private repository
    steps:
      - name: Checkout repository
        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
        with:
          persist-credentials: false

      - name: Run Trivy vulnerability scanner in repo mode
        uses: aquasecurity/trivy-action@b6643a29fecd7f34b3597bc6acb0a98b03d33ff8 # 0.33.1
        with:
          scan-type: "fs"
          scanners: "misconfig,secret"
          format: "github"
          exit-code: 1
          version: "v${{ env.TRIVY_VERSION }}"
[...]

PR #1640, opened 2026-02-20
fix(known-vulnerable-actions): report fix instead of early exit on fail

reubenwong97:
## Pre-submission checks

Please check these boxes:

- [x] **Mandatory**: This PR corresponds to an issue: #1634 .

## Summary

There was a misleading error message about a fatal error when a vulnerable action was used, and a fix was not successfully generated. I believe the issue lies here: https://github.com/zizmorcore/zizmor/blob/main/crates/zizmor/src/audit/known_vulnerable_actions.rs#L264. When the creating the upgrade fix fails, the error is immediately propagated without properly reporting the finding. This change will check the result of the attempted fix resolution, and only apply it if it was successful.

After this change, the error is reported properly:
```text
warning[known-vulnerable-actions]: action has a known vulnerability
  --> /tmp/trivy.yml:19:9
   |
19 |         uses: aquasecurity/trivy-action@b6643a29fecd7f34b3597bc6acb0a98b03d33ff8 # 0.33.1
   |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ this step
   |
```

### DIFF
--- a/crates/zizmor/src/main.rs
+++ b/crates/zizmor/src/main.rs
@@ -94,13 +94,17 @@ struct App {
     #[arg(short, long, env = "ZIZMOR_OFFLINE")]
     offline: bool,
 
-    /// The GitHub API token to use [env: GH_TOKEN or GITHUB_TOKEN]
-    #[arg(long, env, hide_env = true, value_parser = GitHubToken::new)]
+    /// The GitHub API token to use [env: GH_TOKEN or GITHUB_TOKEN or ZIZMOR_GITHUB_TOKEN]
+    #[arg(long, env, hide_env = true, value_parser = GitHubToken::new, conflicts_with_all = ["github_token", "zizmor_github_token"])]
     gh_token: Option<GitHubToken>,
 
     /// This is an alias for `--gh-token` / `GH_TOKEN`.
-    #[arg(long, env, hide = true, value_parser = GitHubToken::new, conflicts_with = "gh_token")]
+    #[arg(long, env, hide = true, value_parser = GitHubToken::new, conflicts_with_all = ["gh_token", "zizmor_github_token"])]
     github_token: Option<GitHubToken>,
+
+    /// This is an alias for `--gh-token` / `GH_TOKEN` / `--github-token` / `GITHUB_TOKEN`
+    #[arg(long, env, hide = true, value_parser = GitHubToken::new, conflicts_with_all = ["gh_token", "github_token"])]
+    zizmor_github_token: Option<GitHubToken>,
 
     /// The GitHub Server Hostname. Defaults to github.com
     #[arg(long, env = "GH_HOST", default_value_t)]
@@ -737,10 +741,12 @@ async fn run(app: &mut App) -> Result<ExitCode, Error> {
         app.persona = Persona::Pedantic;
     }
 
-    // Merge `--github-token` into `--gh-token`, if present.
-    if let Some(token) = app.github_token.take() {
-        app.gh_token = Some(token);
-    }
+    // Merge `--github-token` or `--zizmor-github-token` into `--gh-token`, if present.
+    app.gh_token = app
+        .gh_token
+        .take()
+        .or(app.github_token.take())
+        .or(app.zizmor_github_token.take());
 
     // Unset the GitHub token if we're in offline mode.
     // We do this manually instead of with clap's `conflicts_with` because
--- a/crates/zizmor/tests/integration/cli.rs
+++ b/crates/zizmor/tests/integration/cli.rs
@@ -1,8 +1,8 @@
 use crate::common::zizmor;
 
-/// Test that `--gh-token` and `--github-token` conflict with each other.
+/// Test that `--gh-token`, `--github-token` and `--zizmor-github-token` conflict with each other.
 #[test]
-fn test_gh_token_github_token_conflict() -> anyhow::Result<()> {
+fn test_gh_token_github_token_zizmor_github_token_conflict() -> anyhow::Result<()> {
     // As CLI flags.
     insta::assert_snapshot!(
         zizmor()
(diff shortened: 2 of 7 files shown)


reubenwong97:
Hi @woodruffw ,

Could you help to take a look at this change? Thank you.

Best,
Reuben

woodruffw:
Thanks @reubenwong97! I'll give this a look later today.

crates/zizmor/src/audit/known_vulnerable_actions.rs
```diff
@@ -262,10 +262,12 @@ impl KnownVulnerableActions {
             // to do, since we'd have to decide which finding to attach that
             // fix to.
             if let Some(first_patched_version) = first_patched_version {
-                let fix = self
+                if let Ok(fix) = self
```
woodruffw: This change _looks_ right to me, but it'd be interesting to better understand why the fix creation failed here. 

The original error was this:

```
fatal: no audit was performed
'known-vulnerable-actions' audit failed on file://./.github/workflows/scanning.yml

Caused by:
    0: error in 'known-vulnerable-actions' audit
    1: Cannot resolve version 0.34.0 to commit hash for aquasecurity/trivy-action
```

But there is a v0.34.0 tag:

https://github.com/aquasecurity/trivy-action/releases/tag/0.34.0


So maybe this is an instance where we tried both `0.34.0` and `v0.34.0` as tag refs, and failed early when the first didn't exist? That would make sense to me, but it'd be good to know that for certain 🙂

woodruffw:
@reubenwong97 It'd be great to also get a test case added for this, I think there are similar test cases already in the file's unit tests (or in the audit-specific integration tests).

crates/zizmor/src/audit/known_vulnerable_actions.rs
reubenwong97: Hi @woodruffw ,

It would seem that is indeed the case. Its actually `v0.34.0` that is tried first, and that fails. Because we can't find the commit, we return `None`, which ends up propagating as an error.

```rust
   for tag in tags {
        tracing::warn!("Tag name: {}", tag.name);
        if tag.name == git_ref {
            return Ok(Some(tag.commit.sha));
        }
    }
```
In the above area (the warn is just so its easier for me to read):
```text
WARN audit:audit{input=Workflow(file:///tmp/trivy.yml)}:commit_for_ref{owner="aquasecurity" repo="trivy-action" git_ref="v0.34.0"}: zizmor::github: Tag name: 0.34.0
```
We were searching for `v0.34.0`, but `0.34.0` was the one that existed and was in use.

What would the desired behavior for zizmor be here?

reubenwong97:
> @reubenwong97 It'd be great to also get a test case added for this, I think there are similar test cases already in the file's unit tests (or in the audit-specific integration tests).

Hi @woodruffw ,

Added a test. I believe this needs to be an online test?.. Referenced the other tests that required `GH_TOKEN` and asserted that findings could complete successfully.

crates/zizmor/src/audit/known_vulnerable_actions.rs
woodruffw: Ah gotcha, I got mixed up with `v0.34.0` (the release name) and `0.34.0` the actual tag ref. Makes sense!

I think the desired behavior here is that we _should_ be correctly generating a fix here, rather than failing fast on the first lookup. I'm actually kind of confused as to why we don't successfully fall back to looking up `0.34.0`.

In other words:

1. We should generate a fix in this case.
2. In a case where both lookups fail, we should emit a warning (but not fail the entire audit)

reubenwong97: Cool! I'll work on this.

woodruffw:
> Added a test. I believe this needs to be an online test?.. Referenced the other tests that required `GH_TOKEN` and asserted that findings could complete successfully.

Thanks! Yep, it does need to be online.

crates/zizmor/src/audit/known_vulnerable_actions.rs
reubenwong97: Hi @woodruffw ,

I've updated the behavior, as well as the test, which is now able to automatically suggest a fix. The original behavior did in fact attempt to also search using the bare version. However, this was only in the event of an error from `commit_for_ref`. If the search had no error, but did not yield any results, it did not try the bare version. The new behavior now does that. The presence of an auto-fix is now reported:

```text
warning[known-vulnerable-actions]: action has a known vulnerability
  --> /tmp/trivy.yml:18:9
   |
18 |         uses: aquasecurity/trivy-action@b6643a29fecd7f34b3597bc6acb0a98b03d33ff8 # 0.33.1
   |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ this step
   |
```

Could you help to review again? Thank you!

woodruffw: Nice, good catch! I'll try and give this a review in a bit.

crates/zizmor/src/audit/known_vulnerable_actions.rs
```diff
@@ -165,7 +165,21 @@ impl KnownVulnerableActions {
                     .commit_for_ref(uses.owner(), uses.repo(), &prefixed_version)
                     .await
                 {
-                    Ok(commit) => commit.map(|commit| (&prefixed_version, commit)),
+                    Ok(commit) => {
```
woodruffw: I think we can deduplicate a bit here by having this be `Ok(Some(commit))` and then having the other match be `Ok(None) | Err(_)'. That way we only need one copy of the fallback check 🙂

reubenwong97: Hi @woodruffw ,

Thanks for the suggestion! It looks a lot nicer now 😄

woodruffw (approved):
Thanks @reubenwong97, nice work!

PR #1646, opened 2026-02-23
Tweak release workflow triggers

woodruffw:
The idea here is to do more single-sourcing on release state, i.e. all releases begin when I create a draft release (triggering release binary builds). Then, the rest of the release process continues when the release is actually published (triggering PyPI wheel uploads).

This should also allow me to enable immutable releases on GitHub.


woodruffw:
Will do a prerelease after this.

PR #1647, opened 2026-02-23
yamlpath 0.34.0

woodruffw:


woodruffw:
@kusari-inspector feedback this looks like a false positive -- presumably in-tree crates shouldn't be flagged like this 🙂

PR #1655, opened 2026-02-23
feat: new audit: agentic-actions (#1605)

### LINKED ISSUE #1605: New audit: `agentic-actions` — detect insecure AI coding agent configurations
AI coding agents (Claude Code Action, Gemini CLI, OpenAI Codex, GitHub Copilot) are increasingly used in GitHub Actions for PR review, issue triage, and code generation. These introduce a new vulnerability class: **prompt injection → code execution**.

Attacker-controlled content (issue titles, PR bodies, comments) flows into an LLM prompt, and the LLM has access to privileged tools (shell commands, `gh` CLI, file writes) that it can be instructed to use by injected content. This differs from traditional template injection (`${{ }}`):

- Traditional: attacker text is interpolated into code and executed directly
- AI agent: attacker text is interpreted *semantically* by an LLM, which then uses its tools to carry out the injected instructions
- The standard `template-injection` mitigation (use environment variables) doesn't help because the LLM processes the *content*, not the code structure

Recent public disclosures:

- [Trusting Claude with a Knife](https://johnstawinski.com/2026/02/05/trusting-claude-with-a-knife-unauthorized-prompt-injection-to-rce-in-anthropics-claude-code-action/) — PR title modified after `issue_comment` trigger → prompt injection → binary overwrite → RCE with full secret access
- [PromptPwnd](https://www.aikido.dev/blog/promptpwnd-github-actions-ai-agents) — Issue body prompt injection → `gh issue edit` exfiltrating `GITHUB_TOKEN` and API keys. Demonstrated across Claude Code, Gemini CLI, OpenAI Codex, and GitHub AI Inference. At least five Fortune 500 companies affected.

## Proposed detection

Maintain a registry of known AI agent actions and their security-relevant configuration parameters (similar to how `template-injection` tracks `ACTION_INJECTION_SINKS`):

| Action | Dangerous config parameters |
|--------|---------------------------|
| `anthropics/claude-code-action` | `allowed_non_write_users`, `prompt`, `direct_prompt` |
| `google-gemini/gemini-cli-action` | prompt-related inputs |
| `openai/codex-action` | `allow-users`, `safety-strategy` |
| `github/ai-inference` | `enable-github-mcp` |
| Copilot-related actions (TBD) | prompt inputs |

When an AI agent action is detected, analyze the surrounding context for risk signals:

1. **Trigger type** — Is the workflow triggered by events where untrusted users can inject content?
   - `issues`, `issue_comment`: anyone who can file issues on public repos
   - `pull_request_target`: external forks
   - `pull_request`: external contributors
   - `discussion`, `discussion_comment`

2. **Permissive user configuration** — Does the action explicitly allow untrusted users to invoke it?
   - `allowed_non_write_users: "*"` (Claude Code Action — lets any GitHub user trigger the agent)
   - `allow-users: "*"` (OpenAI Codex — same)
   - These settings are the difference between "only maintainers can invoke" and "any random user can invoke"

3. **Attacker-controlled data in prompts** — Are expressions like `${{ github.event.issue.title }}`, `${{ github.event.issue.body }}`, `${{ github.event.comment.body }}`, `${{ github.event.pull_request.title }}` used in the action's `with:` inputs or `env:` block? Even when passed through environment variables, these flow into the LLM prompt semantically.

4. **Permission scope** — Does the workflow/job have write permissions? `contents: write` combined with an AI agent means the agent (or an attacker through it) can push code, merge PRs, and establish persistence. The Stawinski article specifically demonstrates using the exfiltrated PAT with `contents:write` to add workflows to the default branch.

## Severity

| Scenario | Severity | Confidence | Persona |
|----------|----------|------------|---------|
| AI agent + attacker-controllable trigger + permissive users (`*`) | High | High | Regular |
| AI agent + attacker-controllable trigger + attacker data in prompt + write perms | High | High | Regular |
| AI agent + attacker-controllable trigger + attacker data in prompt | Medium | High | Regular |
| AI agent + `issue_comment` trigger (base repo context with secrets) | Medium | Medium | Regular |
| AI agent + any attacker-controllable trigger | Low | Medium | Pedantic |

## Implementation notes

- The audit should implement `audit_workflow` to get trigger context, then walk steps looking for known AI agent actions
- The agent action registry should be extensible as new agents appear (this space is evolving rapidly)
- Remediation advice should include: restricting trigger types, limiting user access (`allowed_non_write_users` to specific trusted users), minimizing permissions, avoiding passing attacker-controlled context directly into prompts

## Related

- #935 — TOCTOU vulnerabilities; the Stawinski article specifically exploits a TOCTOU where the PR title is changed between trigger and prompt construction
- Distinct from `template-injection` because the threat model and mitigations are fundamentally different — this is about LLM semantic interpretation, not syntactic code injection

BuffaloWill:
Title: feat: new audit: agentic-actions

---

## Pre-submission checks

- [x] **Mandatory**: This PR corresponds to an issue: #1605 .

## Summary

Adds an `agentic-actions` audit that detects insecure configurations of
AI coding agent actions (Claude Code Action, Codex, Gemini CLI, etc.).

Checks performed:

- **Permissive users** (`Permissive user config`): detects `allowed_non_write_users`, `allowed_bots` (Claude)
  and `allow-users`, `allow-bots` (Codex) set to wildcard `*`.
- **Missing Gemini tool restrictions** (`Unrestricted tools`): flags Gemini actions missing `settings`
  with `coreTools` or `excludeTools` constraints. This allows for any tool to be run.
- **Dangerous sandbox/safety overrides** (`Excessive permissions`): catches `sandbox: danger-full-access` (Codex),
  `safety-strategy: unsafe` (Codex), `Bash(*)` in `claude_args`, and Gemini
  `sandbox: false`.
- **Attacker-controllable triggers**: flags workflows triggered by `issue_comment`,
  `pull_request_target`, or similar events that let untrusted users invoke an agent.
  I added a `tip` to list detected gates (if conditions, actor checks) so maintainers
  can assess whether existing mitigations are sufficient. Without it, this check seems
  much more concerning than it might be, given the gates.
- **Attacker-controlled expressions**: detects GitHub expressions like
  `${{ github.event.issue.body }}` flowing into agent prompt fields.
- **Archived actions**: flags `google-gemini/gemini-cli-action`, which should be
  migrated to `google-github-actions/run-gemini-cli`.

Closes #1605.

## Test Plan

18 YAML test fixtures covering each detection category. Integration tests
assert expected finding counts per fixture.

### DIFF
--- a/crates/zizmor/src/audit/agentic_actions.rs
+++ b/crates/zizmor/src/audit/agentic_actions.rs
@@ -200,7 +200,7 @@ impl AgenticActions {
                 BareEvent::PullRequestReviewComment => "pull_request_review_comment",
                 BareEvent::DiscussionComment => "discussion_comment",
                 BareEvent::Discussion => "discussion",
-                _ => "unknown",
+                _ => unreachable!(),
             };
 
             let gates = if let Trigger::Events(events) = &workflow.on {


crates/zizmor/src/config/schema.rs
```diff
@@ -77,6 +77,7 @@ macro_rules! define_audit_rules {
 }
 
 define_audit_rules! {
+    agentic_actions,
```
woodruffw: New audits go at the bottom of this list

crates/zizmor/src/registry.rs
```diff
@@ -44,6 +44,7 @@ impl AuditRegistry {
             }};
         }
 
+        register_audit!(audit::agentic_actions::AgenticActions);
```
woodruffw: Same as config, new audits go at the bottom

crates/zizmor/src/config/schema.rs
BuffaloWill: 👍 thank you. fixed.

crates/zizmor/src/registry.rs
BuffaloWill: fixed.

crates/zizmor/src/audit/agentic_actions.rs
```diff
@@ -0,0 +1,686 @@
...
+    BareEvent::Issues,
+    BareEvent::PullRequestTarget,
+    BareEvent::PullRequest,
+    BareEvent::PullRequestReview,
+    BareEvent::PullRequestReviewComment,
+    BareEvent::DiscussionComment,
+    BareEvent::Discussion,
+];
+
+const ATTACKER_CONTROLLED_PATTERNS: &[&str] = &[
```
woodruffw, line 97: This should use the expression and `ContextPattern` APIs -- without those, this will have false negatives on alternative context syntaxes (like `github['event'][...]`) and won't correctly handle non-trivial expressions.

For reference, I think `template-injection` shows how to parse an expression and use `dataflow_contexts()` to access the "tainting" contexts in each expression.

(I strongly recommend doing this part at least partially manually, my experience with Claude and other LLMs is that they get "lazy" at the layers of abstraction here!)

crates/zizmor/src/audit/agentic_actions.rs
```diff
@@ -0,0 +1,686 @@
...
+            EnvValue::String(s) => serde_json::from_str(s)
+                .inspect_err(|e| {
+                    tracing::warn!("agentic-actions: malformed JSON in settings field: {e}");
+                })
+                .ok(),
+            _ => None,
+        });
+
+        let has_restriction = settings.as_ref().is_some_and(|s: &serde_json::Value| {
+            s.get("coreTools").is_some() || s.get("excludeTools").is_some()
```
woodruffw: Do you have a docs link for these Gemini CLI config settings? I tried finding `coreTools` and `excludeTools` in their docs but couldn't.

There are similar `tools.core`, `tools.allowed`, and `tools.exclude` documented here: https://geminicli.com/docs/reference/configuration/#available-settings-in-settingsjson

(But I don't know Gemini well enough to understand if those are the same.)

crates/zizmor/src/audit/agentic_actions.rs
```diff
@@ -0,0 +1,686 @@
...
+    "${{ github.event.discussion.title }}",
+    "${{ github.event.discussion.body }}",
+    "${{ github.event.head_commit.message }}",
+    "${{ github.event.head_commit.author.email }}",
+    "${{ github.event.head_commit.author.name }}",
+    "${{ github.head_ref }}",
+];
+
+const EXPANDABLE_COMMANDS: &[&str] =
+    &["echo", "cat", "printf", "tee", "head", "tail", "wc", "sort"];
```
woodruffw, line 115: What makes these tools dangerous? They look relatively innocuous to me; is it that they can introduce arbitrary data and thus poison the prompt?

crates/zizmor/src/audit/agentic_actions.rs
```diff
@@ -0,0 +1,686 @@
...
+
+    fn summarize_list(items: &[String], max: usize) -> String {
+        if items.len() <= max {
+            items.join(", ")
+        } else {
+            format!("{}, ... ({} total)", items[..max].join(", "), items.len())
+        }
+    }
+
+    fn check_archived_action<'doc>(
```
woodruffw: What is this check doing? This looks redundant with `archived-uses`, but maybe I'm missing something.

Ref: https://docs.zizmor.sh/audits/#archived-uses

crates/zizmor/src/audit/agentic_actions.rs
```diff
@@ -0,0 +1,686 @@
...
+        };
+        Ok(vec![
+            Self::finding()
+                .severity(Severity::Medium)
+                .confidence(Confidence::High)
+                .add_location(
+                    step.location()
+                        .primary()
+                        .with_keys(["uses".into()])
+                        .annotated(format!("archived; migrate to {replacement}")),
```
woodruffw: If this is indeed distinct from `archived-uses`, this should be a generated fix instead of an inline annotation. But I think this might just be a dupe 🙂

crates/zizmor/src/audit/agentic_actions.rs
```diff
@@ -0,0 +1,686 @@
...
+            let name = match event {
+                BareEvent::IssueComment => "issue_comment",
+                BareEvent::Issues => "issues",
+                BareEvent::PullRequestTarget => "pull_request_target",
+                BareEvent::PullRequest => "pull_request",
+                BareEvent::PullRequestReview => "pull_request_review",
+                BareEvent::PullRequestReviewComment => "pull_request_review_comment",
+                BareEvent::DiscussionComment => "discussion_comment",
+                BareEvent::Discussion => "discussion",
+                _ => "unknown",
```
woodruffw: Is this unreachable?

crates/zizmor/src/audit/agentic_actions.rs
```diff
@@ -0,0 +1,686 @@
...
+                );
+
+            builder = if all_gates.is_empty() {
+                builder.tip("no gates detected — any user can trigger this agent")
+            } else {
+                builder.tip(format!(
+                    "detected gates that may limit exposure:\n{}",
+                    all_gates
+                        .iter()
+                        .map(|g| format!("  • {g}"))
```
woodruffw, line 395: Pretty-printing into the tip body like this might be too clever, and is likely to break the annotation library we use to annotate spans.

Instead, I think there are two options here:

1. Emit one tip per "gate" here;
2. Turn these "gates" into distinct locations in the finding instead, since that's what they really are.

I think (2) is probably best, since it'd be consistent with all of the other audits.

crates/zizmor/src/audit/agentic_actions.rs
woodruffw: Related, I think this should use the same `CONTEXT_CAPABILITIES_FST` as `template-injection`, rather than hard-coding specific bad contexts here.

crates/zizmor/src/audit/agentic_actions.rs
BuffaloWill: Thanks, good call. It was deprecated (https://github.com/google-gemini/gemini-cli/blob/0e85e021dcfdbe7abb2594f38f516ba3bb7ef33e/docs/get-started/configuration-v1.md) in v1, but I have seen it in workflows using older versions. I swapped out that code to check for `tools.core` and `tools.exclude`. I left out `tools.allowed` because afaik it only controls confirmation bypass, not tool availability.

crates/zizmor/src/audit/agentic_actions.rs
BuffaloWill: 👍 Yes, they can be used to introduce data into the prompt. They can also exfiltrate secrets;  e.g., echo $GITHUB_TOKEN piped into a workflow output or issue comment. I might remove `wc` and `sort` though; I have validated stealing secrets with echo, cat, and printf but not wc or sort.

crates/zizmor/src/audit/agentic_actions.rs
BuffaloWill: You're right, I think this is redundant with `archived-uses`. Thank you. I removed it.

crates/zizmor/src/audit/agentic_actions.rs
BuffaloWill: Removed 👍

crates/zizmor/src/audit/agentic_actions.rs
woodruffw: Thanks! If both are still supported, maybe we should detect both?

crates/zizmor/src/audit/agentic_actions.rs
BuffaloWill: fixed 🙏

crates/zizmor/src/audit/agentic_actions.rs
BuffaloWill: I am open to it; I just might need some guidance on the detection logic. I think we would need to detect the version first; i.e.`uses: google-gemini/gemini-cli-action@v1`, `uses: google-gemini/gemini-cli-action@v1.2.3`, commit sha aligning with v1 version.

crates/zizmor/src/audit/agentic_actions.rs
woodruffw: Sorry, I meant two things here 😅

1. It was a question, I wanted to make sure it's _actually_ unreachable. 
2. If it is actually unreachable, is there a way we can remove the explicit runtime check? IME these tend to be forgotten and eventually violated over time; it'd be ideal to have the compiler enforce it for us.

crates/zizmor/src/audit/agentic_actions.rs
BuffaloWill: Thanks @woodruffw. Should I add `pub(crate)` to the existing `from_context` method in `template_injection.rs`? I think this would allow `agentic_actions.rs` to call `Capability::from_context()` directly, then replace the hardcoded `ATTACKER_CONTROLLED_PATTERNS` list with the same FST lookup that template_injection uses. At a high level, parse the expressions first, then classify each context via from_context()? afaict I wouldn't have to copy or move code to a shared location this way.

crates/zizmor/src/audit/agentic_actions.rs
BuffaloWill: Sorry 😄.

1. Yes, I believe it is unreachable. 
2. Yes, I can do that.

Issue #1560, opened 2026-01-23
Switch to TOML for zizmor's config


woodruffw:
I want this for two reasons:

1. Using YAML for the config was mostly a mistake -- I picked it because serge-yaml was already a dependency. 
2. Having the config be `zizmor.toml` instead would mean no more JSON Schema conflicts between `zizmor.yml` (the config) and `zizmor.yml` (the workflow).

The downside to doing this is that it'll be disruptive, so I can't _force_ it until a 2.x release. But I can probably start with it now and offer a migration path (`zizmor --migrate-config`?)

reubenwong97:
Hi @woodruffw ,

Do you think this is something I could look into? This appears to be a fairly self-contained task dealing with the config module.

Best,
Reuben

woodruffw:
Hey @reubenwong97, yeah, go for it! This is going to be an interesting one since it has compatibility implications, but the task itself is definitely self-contained.

reubenwong97:
Hi @woodruffw ,

I'm taking a look at this now, and am thinking about the approach here. Should I try to make the toml config compatible with the yaml config (if possible), and have `zizmor.toml` as one of the `CONFIG_CANDIDATES`? Or should I go down the path assuming that yaml configs would be phased out when we release this as part of a 2.0 release?

Thank you!

woodruffw:
> I'm taking a look at this now, and am thinking about the approach here. Should I try to make the toml config compatible with the yaml config (if possible), and have `zizmor.toml` as one of the `CONFIG_CANDIDATES`? Or should I go down the path assuming that yaml configs would be phased out when we release this as part of a 2.0 release?

Yeah, I think it should be 1-1 compatible with the current YAML config. What I was thinking was that it should be added to the `CONFIG_CANDIDATES` like you said, and given priority over YAML if both are present (i.e. we should always try and load TOML first, then fall back to YAML).

In terms of migration strategy, what I was thinking was:

1. We land TOML support and make it the default
2. If the user loads from a YAML config, we emit a loud warning telling them that they should switch to the TOML format
    - As part of this, we should maybe support some kind of `zizmor --migrate-config` option that will help people auto-migrate from YAML to TOML
3. With 2.x, we'll remove the YAML support entirely and simply ignore any legacy YAML configs that are present

My original thought was to retain the exact same shape/data layout for the YAML and TOML configs, but I could be convinced that they should be different if you can think of a good reason to do that. I'm open to ideas there 🙂

[...]

PR #1666, opened 2026-02-26
feat(config): use toml for configuring zizmor

reubenwong97:
## Pre-submission checks

Please check these boxes:

- [x] **Mandatory**: This PR corresponds to an issue: #1560 .

## Summary

This PR is currently a **DRAFT** for the purposes of obtaining feedback.

The changes are broadly:
* Introduction of a `ConfigFormat` enum so that the appropriate deserializer is called.
* Updates of signatures of `load` to accept `ConfigFormat`
* Use of `serde_json::Value` as the intermediate representation. This was chosen since its variants could be produced by yaml or toml configs in zizmor (as opposed to `serde_yaml::Value` which has `Tagged` variants for example).
* Removal of `ForbiddenUsesConfigInner`, the change to `serde_json` removed keys being deserialized as yaml tags from being an issue.
* Additional error variants for `ConfigErrorInner::Syntax` since it could be due to a toml or serde_yaml deserialization error.

### TODO:
[ ] Migration from yaml config to toml

## Test Plan

I've so far tested dummy configs (yaml and toml) against the coreutils workflows and managed to obtain the same results. When the approach is more set in stone, I will add some tests.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1216,7 +1216,7 @@ dependencies = [
  "serde",
  "serde_derive",
  "sysinfo",
- "toml",
+ "toml 0.9.10+spec-1.1.0",
  "uuid",
 ]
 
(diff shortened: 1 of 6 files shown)


reubenwong97:
Hi @woodruffw ,

I have a first draft of the change to use TOML configs here. Could you help to take a quick look when you have time? I'm not the most confident in these changes, primarily due to introducing an extra `ConfigFormat` that needs to be passed into `load` call-sites, as well as the error variants and use of `serde_json::Value` as the intermediate type that yaml and toml can deserialize into.

Thanks a lot!

Best,
Reuben

woodruffw:
Thanks @reubenwong97, I'll try and give this a review in the coming days! My time for the next few days is somewhat limited, so it might take me a bit.

PR #1677, opened 2026-03-01
feat(unpinned-images): detect docker/podman commands in run: steps

### LINKED ISSUE #738: Feature: `unpinned-images` could discover `docker pull ...` patterns in `run:` clauses
Has this idea while reviewing #733.

Right now we check for `container:` and `services:` for image references, but lots of users probably also do `docker pull ...` in `run:` clauses and forget to tag/hash-pin there.

We should consider detecting those as well, in a manner similar to how we handle shell parsing in `github-env`.

(This will require a bit of a refactor, most likely.)

jfagoagas:
## Pre-submission checks

Please check these boxes:

      one first).

### Motivation

GitHub's [agentic workflows](https://github.github.com/gh-aw/) are increasingly generating `run:` step scripts that include `docker pull`/`docker run` commands. Unlike human-authored workflows where a developer might remember to pin images, agent-generated scripts almost never pin to a digest. This turns every unpinned `docker pull` into a supply chain injection point — an attacker who hijacks a tag on a public registry gets code execution inside the workflow.

The existing `unpinned-images` audit only covered declarative `container:` and `services:` blocks, missing the imperative `docker` commands that are becoming more common as CI pipelines get more dynamic.

This also opens the door for future audits targeting other agentic workflow patterns: unvalidated tool outputs piped into shell commands, dynamic action references constructed from agent suggestions, or secrets passed to agent-spawned containers.

## Summary

Closes #738.

Extends the `unpinned-images` audit to detect unpinned image references in `docker pull`, `docker run`, `docker create`, and their `podman` equivalents within `run:` steps, in addition to the existing `container:` and `services:` checks.

- Uses tree-sitter bash parsing (following the `github-env` audit pattern) to identify docker/podman command invocations, avoiding false positives from strings and comments (e.g. `echo "docker pull ubuntu"`)
- Handles flags, combined short flags (`-dit`), and `--flag=value` syntax to correctly extract the image argument
- Skips shell variable expansions (`$IMAGE`) that can't be statically resolved
- Adds `audit_step` for workflow steps and `audit_composite_step` for action steps

## Test Plan

- [ ] `cargo build` compiles
- [ ] `cargo test -p zizmor -- unpinned_images` — existing snapshot tests unchanged, new tests pass
- [ ] `cargo insta review` — accept new snapshots for `test_run_docker_pedantic` and `test_run_docker_regular`
- [ ] `cargo clippy` — no new warnings
- [ ] Unit tests cover: argument extraction, flag skipping, combined flags, subcommand filtering, echo false-positive, variable skipping, podman, multiline scripts

### DIFF
--- a/crates/zizmor/src/audit/unpinned_images.rs
+++ b/crates/zizmor/src/audit/unpinned_images.rs
@@ -128,7 +128,12 @@ impl UnpinnedImages {
                 let text = node.utf8_text(source).ok()?;
                 text.strip_prefix('\'')?.strip_suffix('\'')
             }
-            _ => node.utf8_text(source).ok(),
+            // Bare words contain plain text with no quoting or expansion.
+            "word" => node.utf8_text(source).ok(),
+            // Unknown node types (e.g. `simple_expansion`, `expansion`,
+            // `command_substitution`, `concatenation`) contain dynamic
+            // content — reject them.
+            _ => None,
         }
     }
 
@@ -885,8 +890,18 @@ mod tests {
             .unwrap();
         assert!(images.is_empty());
 
-        // Variable expansion should be skipped
+        // Variable expansion should be skipped (simple_expansion node)
         let images = sut.bash_docker_images("docker pull $IMAGE_NAME").unwrap();
+        assert!(images.is_empty());
+
+        // Brace variable expansion should be skipped (expansion node)
+        let images = sut.bash_docker_images("docker pull ${IMAGE_NAME}").unwrap();
+        assert!(images.is_empty());
+
+        // Concatenation of word + expansion should be skipped
+        let images = sut
+            .bash_docker_images("docker pull myregistry/${IMAGE_NAME}:latest")
+            .unwrap();
         assert!(images.is_empty());
 
         // Podman support


crates/zizmor/src/audit/unpinned_images.rs
```diff
@@ -1,38 +1,238 @@
...
+                .filter_map(|cap| cap.node.utf8_text(script.as_bytes()).ok())
+                .collect();
+
+            if let Some(image_ref) = Self::extract_docker_image(&args) {
+                // Skip arguments containing shell variable expansions or
+                // GitHub Actions expressions — we can't statically resolve those.
+                if !image_ref.contains('$') {
+                    images.push(image_ref.to_string());
+                }
+            }
```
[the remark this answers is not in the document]
jfagoagas: Implemented in [`97e1662` (this PR)](https://github.com/zizmorcore/zizmor/pull/1677/commits/97e1662438db05780ed931a8cc9dfeb27018257a)

crates/zizmor/src/audit/unpinned_images.rs
```diff
@@ -1,38 +1,238 @@
...
 use super::{Audit, AuditLoadError, audit_meta};
 
-pub(crate) struct UnpinnedImages;
+/// Tree-sitter query matching `docker` and `podman` command invocations in bash.
+const BASH_DOCKER_CMD_QUERY: &str = r#"
+(command
+  name: (command_name) @cmd
+  argument: (_)* @args
+  (#match? @cmd "^(docker|podman)$")
+) @span
```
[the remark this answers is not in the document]
jfagoagas: Implemented in [`97e1662` (this PR)](https://github.com/zizmorcore/zizmor/pull/1677/commits/97e1662438db05780ed931a8cc9dfeb27018257a)

crates/zizmor/src/audit/unpinned_images.rs
```diff
@@ -1,38 +1,238 @@
...
+    fn extract_docker_image<'a>(args: &[&'a str]) -> Option<&'a str> {
+        let subcommand = args.first()?;
+        match *subcommand {
+            "pull" | "run" | "create" => {
+                let rest = &args[1..];
+                Self::first_positional_arg(rest)
+            }
+            _ => None,
+        }
+    }
```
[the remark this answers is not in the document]
jfagoagas: Implemented in [`97e1662` (this PR)](https://github.com/zizmorcore/zizmor/pull/1677/commits/97e1662438db05780ed931a8cc9dfeb27018257a)

crates/zizmor/src/audit/unpinned_images.rs
```diff
@@ -1,38 +1,238 @@
...
+const BASH_DOCKER_CMD_QUERY: &str = r#"
+(command
+  name: (command_name) @cmd
+  argument: (_)* @args
+  (#match? @cmd "^(docker|podman)$")
+) @span
+"#;
+
+/// Boolean short flags for docker pull/run/create that do NOT consume the next argument.
+const BOOLEAN_SHORT_FLAGS: &[&str] = &["-d", "-i", "-t", "-P", "-q", "-a"];
```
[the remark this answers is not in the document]
jfagoagas: Implemented in [`97e1662` (this PR)](https://github.com/zizmorcore/zizmor/pull/1677/commits/97e1662438db05780ed931a8cc9dfeb27018257a)

crates/zizmor/src/audit/unpinned_images.rs
```diff
@@ -1,38 +1,319 @@
...
+    "--env-host",
+    "--http-proxy",
+    "--no-hostname",
+    "--no-hosts",
+    "--passwd",
+    "--read-only-tmpfs",
+    "--replace",
+    "--rmi",
+    "--rootfs",
+];
```
woodruffw, line 77: I don't think this approach is sustainable long-term, unfortunately -- we either need a lighter-weight way to detect docker images/invocations in `run:` blocks, or something that's more formal and testable (e.g. explicitly modeling these tools using `clap`).

(As is, this is going to be brittle on any CLI changes to either `docker` or `podman`, which in practice is going to mean a lot of false positive reports.)

crates/zizmor/src/audit/unpinned_images.rs
```diff
@@ -1,38 +1,319 @@
...
+                    // Multiple children means interpolation — skip these.
+                    None
+                }
+            }
+            "raw_string" => {
+                // Single-quoted: strip the surrounding quotes.
+                let text = node.utf8_text(source).ok()?;
+                text.strip_prefix('\'')?.strip_suffix('\'')
+            }
+            _ => node.utf8_text(source).ok(),
```
woodruffw: This fallback seems risky -- I think you want to match `"word"` explicitly here and then have unknown node types here be `_ => None`?

crates/zizmor/src/audit/unpinned_images.rs
```diff
@@ -1,38 +1,319 @@
...
+    /// In tree-sitter-bash:
+    /// - `word` nodes contain bare text (no quotes)
+    /// - `string` nodes are double-quoted — `utf8_text()` includes the `"`s
+    /// - `raw_string` nodes are single-quoted — `utf8_text()` includes the `'`s
+    fn arg_text<'a>(node: &tree_sitter::Node, source: &'a [u8]) -> Option<&'a str> {
+        match node.kind() {
+            "string" => {
+                // For double-quoted strings, extract the string_content child
+                // to avoid the surrounding quotes.
+                if node.named_child_count() == 1 {
```
woodruffw: I think this should explicitly match the `string_content` node, rather than assuming that any child of `string` is `string_content` -- the latter is probably not guaranteed to be true.

For example, `"$(abc)"` is `string > command_substitution` instead of `string > string_content`.

crates/zizmor/src/audit/unpinned_images.rs
```diff
@@ -1,38 +1,319 @@
...
+                .captures
+                .iter()
+                .filter(|cap| cap.index == args_idx)
+                .filter_map(|cap| Self::arg_text(&cap.node, script.as_bytes()))
+                .collect();
+
+            if let Some(image_ref) = Self::extract_docker_image(&args) {
+                // Skip arguments containing shell variable expansions or
+                // GitHub Actions expressions — we can't statically resolve those.
+                if !image_ref.contains('$') {
```
woodruffw, line 175: Noting: this might not be solvable in a clean way (because of actions expressions), but ideally we'd do this filtering via the CST instead (since `$` could be a valid part of an image identifier, in principle).

crates/zizmor/src/audit/unpinned_images.rs
```diff
@@ -1,38 +1,319 @@
...
+        });
+
+        Ok(images)
+    }
+
+    /// Extract the image reference from docker/podman command arguments.
+    ///
+    /// `args` contains all arguments after the command name (docker/podman).
+    /// Handles global flags before the subcommand (e.g. `docker --context foo run alpine`).
+    fn extract_docker_image<'a>(args: &[&'a str]) -> Option<&'a str> {
```
woodruffw, line 188: Per the comment above, I think this is unfortunately going to be too brittle. If the goal is to parse Docker CLI arguments, we should use `clap` or another fully-equipped parser that can track positionals/consumption states for us.

woodruffw: (This is going to be nontrivial, however, since Docker and Podman have large CLIs.)

crates/zizmor/src/audit/unpinned_images.rs
```diff
@@ -1,38 +1,319 @@
...
+use crate::utils;
 
 use super::{Audit, AuditLoadError, audit_meta};
 
-pub(crate) struct UnpinnedImages;
+/// Tree-sitter query matching `docker` and `podman` command invocations in bash.
+///
+/// NOTE: `@span` is required by [`utils::SpannedQuery`] and can be used in the
+/// future to produce more precise finding locations within multiline scripts.
+const BASH_DOCKER_CMD_QUERY: &str = r#"
```
woodruffw, line 25: What about pwsh, etc.? We generally try to support those too since they're common enough (unfortunately) in GitHub Actions.

(Those don't need to be part of an MVP here, but curious if you have thoughts on them.)

crates/zizmor/src/audit/unpinned_images.rs
```diff
@@ -44,13 +325,55 @@ audit_meta!(
...
+                "unpinned-images: couldn't determine shell type for {workflow}:{job} step {stepno}; assuming bash",
+                workflow = step.workflow().key.presentation_path(),
+                job = step.parent.id(),
+                stepno = step.index
+            );
+            "bash"
+        });
+
+        for image_str in self.docker_images_in_run(run, shell)? {
+            let location = step.location().primary().with_keys(["run".into()]);
```
woodruffw, line 373: This _probably_ needs to use a subfeature to get the exact command/argument span, since `run:` blocks can be very large.

jfagoagas:
> Thanks for opening a PR @jfagoagas.

 Thank you for maintaining such a great tool 🙌 

> I'm interested in this feature, but I'm concerned about this approach's stability/brittleness; I've left some comments to that effect.

I'll take a look and reply inline.

> Separately, this PR is _probably_ too big to review -- anything above roughly ~500 lines takes me much longer to read and understand, so it'd be ideal to break this up into several smaller PRs (each with tests, where possible). If you're interested in taking that on, we can discuss what an MVP/tractable design would look like here.

I thought this would happen when I was creating it. I saw this change requires a refactor in the audit, so please let me know what's your plan to split this up! We can create a chain of PRs where each one points another and can be merged independently.

crates/zizmor/src/audit/unpinned_images.rs
jfagoagas: I thought about this but I didn't want to add more to this PR. We can add this as part of the plan.

crates/zizmor/src/audit/unpinned_images.rs
jfagoagas: I wasn't aware of `clap` but definitely something to take a look to track this. We can make this simpler covering some scenarios.

crates/zizmor/src/audit/unpinned_images.rs
jfagoagas: I think both CLIs are stable enough but we can explore other options to make this simpler.

crates/zizmor/src/audit/unpinned_images.rs
jfagoagas: Fixed in [`4824027` (this PR)](https://github.com/zizmorcore/zizmor/pull/1677/commits/4824027c586c92065e5e13c78c890d38a4a0b22d)

crates/zizmor/src/audit/unpinned_images.rs
jfagoagas: Fixed in [`2cf625f` (this PR)](https://github.com/zizmorcore/zizmor/pull/1677/commits/2cf625f4dc3e1e8143bb3811019f01260696f16b)

crates/zizmor/src/audit/unpinned_images.rs
jfagoagas: I need to look into this in depth. Rust is not one of my preferred/known languages.

woodruffw:
> Thank you for maintaining such a great tool 🙌

Thank you for the kind words!

Yeah, my preference here would be to come up with a sequence of PRs that add this, each no more than a few hundred lines. Ideally we'd start with just Docker, and do:

1. A PR containing accurate models of the Docker CLI, so we can parse it and extract image references without ad-hoc CLI parsing. These could be based on `clap` or something else; it's a somewhat open design space (but we already use `clap` for zizmor's own CLI). This would also ideally contain tests so we can pre-validate it before adding it to the audit.
2. Integrations into this audit, basically extracting `command (name) (argument) (argument) ...` trees from the bash CST and parsing them with (1) + reaching determinations.
3. Repeating the same for Podman's CLI.
4. Repeating the same for non-bash shells, primarily pwsh. This is optional though; I guess it's probably not very common to invoke Docker or other OCI containers on Windows runners now that I think about it.

I think a sequence roughly like that would make this tractable for me to review.

jfagoagas:
> Thank you for the kind words!
> 
> 
> 
> Yeah, my preference here would be to come up with a sequence of PRs that add this, each no more than a few hundred lines. Ideally we'd start with just Docker, and do:
> 
> 
> 
> 1. A PR containing accurate models of the Docker CLI, so we can parse it and extract image references without ad-hoc CLI parsing. These could be based on `clap` or something else; it's a somewhat open design space (but we already use `clap` for zizmor's own CLI). This would also ideally contain tests so we can pre-validate it before adding it to the audit.
> 
> 2. Integrations into this audit, basically extracting `command (name) (argument) (argument) ...` trees from the bash CST and parsing them with (1) + reaching determinations.
> 
> 3. Repeating the same for Podman's CLI.
> 
> 4. Repeating the same for non-bash shells, primarily pwsh. This is optional though; I guess it's probably not very common to invoke Docker or other OCI containers on Windows runners now that I think about it.
> 
> 
> 
> I think a sequence roughly like that would make this tractable for me to review. 

Sounds like a plan. I'll start with the first one once I get the chance to, most likely during the weekend. I'll keep this PR as a reference if you don't mind until the feature is implemented.

woodruffw:
Yeah, no worries. Thanks for contributing!

jfagoagas:
@woodruffw before jumping into writing code I'd like to validate the following plan with you:

Plan: Split unpinned-images-run-block into reviewable PR chain

## PR chain

### PR 1: Docker CLI models with clap (~350-400 lines, from main)
- New file: `crates/zizmor/src/models/docker.rs`
- Clap-derive structs modeling Docker CLI for pull, run, create subcommands
- Single public API: `DockerCli::parse_image(args: &[&str]) -> Option<String>`
- Clap handles combined `-dit` flags, `--flag=value`, `-eFOO=bar` automatically
- `try_parse_from` failures return None (safe false negatives)
- Unit tests only, no audit dependency
- Docker-only flags (no Podman)

### PR 2: Audit integration (~300-400 lines, net negative from deletions)
- Delete hand-rolled flag parsing (~140 lines of const arrays + helper functions)
- Replace `Self::extract_docker_image(&args)` with `DockerCli::parse_image(&args_refs)`
- Keep all tree-sitter logic, `check_image_ref`, Audit trait impls unchanged
- Add test fixture (`run-docker.yml`) and integration tests

### PR 3: Podman CLI model (~200-250 lines)
- Add `PodmanCli` struct with Podman-specific flags (`--remote`, `--syslog`, `--env-host`, `--replace`, etc.)
- `-a` semantics differ between Docker and Podman, separate structs keep this clean
- Dispatch on `@cmd` tree-sitter capture: "docker" → DockerCli, "podman" → PodmanCli

### PR 4: Documentation (final)
- `docs/audits.md` updates for `run:` block detection

## Open questions: PR review workflow

Two options for how to structure the chain:

Option A — Sequential merge into main: Each PR targets main. PR #2 waits for PR #1 to merge, PR #3 waits for PR #2. You review and merge one at a time.

Option B — Stacked PRs: PR 1 targets main, PR 2 targets PR 1's branch, PR 3 targets PR 2's branch. You can review the full chain at once, then merge bottom-up.

Which do you prefer?

woodruffw:
Thanks @jfagoagas, that plan sounds reasonable to me. I prefer to do sequential merges for this, as long as that works for you!

(I have some small nitpicks on each, but they're minor and can be addressed on-the-fly.)

crates/zizmor/src/audit/unpinned_images.rs
```diff
@@ -1,38 +1,329 @@
...
-pub(crate) struct UnpinnedImages;
+/// Tree-sitter query matching `docker` and `podman` command invocations in bash.
+///
+/// NOTE: `@span` is required by [`utils::SpannedQuery`] and can be used in the
+/// future to produce more precise finding locations within multiline scripts.
+const BASH_DOCKER_CMD_QUERY: &str = r#"
+(command
+  name: (command_name) @cmd
+  argument: (_)* @args
+  (#match? @cmd "^(docker|podman)$")
```
rmuir, line 29:
```suggestion
  (#any-of? @cmd "docker" "podman")
```

Issue #1670, opened 2026-02-27
[BUG]: `if: 0` in step or job condition causes deserialization failure


DarkaMaul:
### Pre-submission checks

- [x] I am **not** filing a feature request. These should be filed via the feature request form instead.
- [x] I have checked the [Troubleshooting Guide](https://docs.zizmor.sh/troubleshooting) for my problem.
- [x] I have looked through both the [open](https://github.com/zizmorcore/zizmor/issues?q=is%3Aissue+is%3Aopen+) and [closed](https://github.com/zizmorcore/zizmor/issues?q=is%3Aissue%20state%3Aclosed) issues for a duplicate report.

### zizmor version

main

### Expected behavior

zizmor should parse workflows containing `if: 0` (or any bare integer in an `if:` condition) without error. GitHub Actions accepts bare integers in `if:` conditions and treats `0` as falsy and non-zero as truthy.

### Actual behavior

zizmor fails to load the workflow with:

```
jobs: data did not match any variant of untagged enum Job at line 4 column 3
```

No audit is performed.

### Reproduction steps

1. Create a workflow file with a bare integer in an `if:` condition:

```yaml
name: test
on: push
jobs:
  test:
    runs-on: ubuntu-latest
    steps:
      - name: step1
        if: 0
        run: echo hello
```

2. Run `zizmor` on it:

```
zizmor workflow.yml
```

```console
🌈 zizmor v1.23.0-rc5
fatal: no audit was performed
error: failed to load file:///tmp/test-if-0.yml as workflow
  |
  = help: this typically indicates a bug in zizmor; please report it
  = help: https://github.com/zizmorcore/zizmor/issues/new?template=bug-report.yml

Caused by:
    0: failed to load file:///tmp/test-if-0.yml as workflow
    1: couldn't turn input into a an appropriate model
    2: jobs: data did not match any variant of untagged enum Job at line 4 column 3
```

### Logs

```plain text

```

### Additional context

Found in the wild in [codeql_windows_msys2.yml](https://github.com/86Box/86Box/blob/9d9d356a154eb7f038702db1cb4c7a3f611f1b83/.github/workflows/codeql_windows_msys2.yml#L171)

woodruffw:
Thanks, seems like it should be a straightforward fix. I think we can probably add another deserialization helper (similar to `scalar_or_vector`) that just coerces numbers to boolean states.

(I suspect `if: 1` also works.)

PR #1683, opened 2026-03-03
Fix 1670

DarkaMaul:
## Pre-submission checks

Please check these boxes:

      one first) #1670 

- [X] I hereby disclose the use of an LLM or other AI coding assistant in the

## Summary

GitHub Actions also accepts bare numeric values in `if:` conditions, this PR handles it.

## Test Plan

Run the tests from zizmor

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -45,7 +45,7 @@ jobs:
         run: cargo build --release
 
       - name: Run the benchmarks
-        uses: CodSpeedHQ/action@dbda7111f8ac363564b0c51b992d4ce76bb89f2f # v4.5.2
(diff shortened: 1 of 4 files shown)


crates/github-actions-models/src/common.rs
```diff
@@ -165,15 +165,44 @@ impl From<BoS> for String {
...
+/// Internal helper for deserializing `If` conditions.
+/// Coerces YAML numeric values to booleans.
+#[derive(Deserialize)]
+#[serde(untagged)]
+enum RawIf {
+    Bool(bool),
+    Int(i64),
+    Float(f64),
+    Expr(String),
+}
```
woodruffw: Nit: you can move this into the `fn deserialize` body below to make the encapsulation/deser-only use more obvious 🙂 

(There's no performance/code size consequence to doing so.)

PR #1687, opened 2026-03-03
Case insensitive comparisons

### LINKED ISSUE #1629: [BUG]: Incorrect evaluation of case-insensitive string comparison in expressions
### Pre-submission checks

### zizmor version

main

### Expected behavior

Consider the following workflow:

```yaml
name: Test Case Sensitivity
on:
  pull_request:
    types: [opened]
jobs:
  test:
    runs-on: ubuntu-latest
    steps:
      - if: ${{ 'hello' == 'HELLO' }}
        run: echo "case insensitive match"
```

GitHub Actions string comparisons are **case-insensitive** ([docs](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/evaluate-expressions-in-workflows-and-actions#operators)). So `'hello' == 'HELLO'` evaluates to `true`.

Running `zizmor --fix` on the workflow should yield:

```yaml
name: Test Case Sensitivity
on:
  pull_request:
    types: [opened]
jobs:
  test:
    runs-on: ubuntu-latest
    steps:
      - if: true
        run: echo "case insensitive match"
```

### Actual behavior

`zizmor` incorrectly evaluates the comparison as case-sensitive, producing:

```yaml
name: Test Case Sensitivity
on:
  pull_request:
    types: [opened]
jobs:
  test:
    runs-on: ubuntu-latest
    steps:
      - if: false
        run: echo "case insensitive match"
```

### Reproduction steps

1. Save the workflow above as `test-case.yml`
2. Run `zizmor --fix test-case.yml`
3. Observe that the `if:` condition is replaced with `false` instead of `true`

### Logs

```plain text

```

### Additional context

This affects at least 49 differential test cases against the reference [languageservices](https://github.com/actions/languageservices) implementation: 41 from operator comparisons (Latin, Cyrillic, Greek characters), 4 from `endsWith`, and 4 from `startsWith`.

DarkaMaul:
## Pre-submission checks

Please check these boxes:

      one first). (#1629)

## Summary

This PR fixes a bug in  `github-actions-expressions` where the comparison between string was case sensitive. 

The [documentation](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/evaluate-expressions-in-workflows-and-actions#operators) says:

> GitHub ignores case when comparing strings.
 
This is confirmed by this [run](https://github.com/trail-of-forks/zizmor/actions/runs/22631929072/job/65584151304)

## Test Plan
I've added tests to ensure the behavior is consistent.

Fixes #1629

### DIFF
--- a/docs/release-notes.md
+++ b/docs/release-notes.md
@@ -163,7 +163,7 @@ of `zizmor`.
 
     Users who with to retain the old (permissive policy) for first-party
     actions may configure it explicitly in their `zizmor.yml`:
-
+  
     ```yaml title="zizmor.yml"
     rules:
       unpinned-uses:
@@ -205,21 +205,21 @@ of `zizmor`.
 * The [use-trusted-publishing] audit now detects additional publishing command
   patterns, including common "wrapped" patterns like `bundle exec gem publish`
   (#1394)
-
+  
 * zizmor now produces better error messages on a handful of error cases involving
   invalid input files. Specifically, a subset of syntax and schema errors now
   produce more detailed and actionable error messages (#1396)
-
+  
 * The [use-trusted-publishing] audit now detects additional publishing command
   patterns, including `uv run ...`, `uvx ...`, and `poetry publish`
   (#1402)
-
+  
 * zizmor now produces more useful and less ambiguous spans for many findings,
   particularly those from the [anonymous-definition] audit (#1416)
-
+  
 * zizmor now discovers configuration files named `zizmor.yaml`, in addition
   to `zizmor.yml` (#1431)
-
+  
 * zizmor now produces a more useful error message when input collection
   yields no inputs (#1439)
 
@@ -227,7 +227,7 @@ of `zizmor`.
   link rendering behavior. This is particularly useful in environments that
   advertise themselves as terminals but fail to correctly render or ignore
   OSC 8 links (#1454)
-
+ 
 ### Performance Improvements 🚄
 
 * The [impostor-commit] audit is now significantly faster on true positives,
@@ -238,9 +238,9 @@ of `zizmor`.
 ### Bug Fixes 🐛
 
 * Fixed a bug where the [obfuscation] audit would crash if it encountered
-  a CMD shell that was defined outside of the current step block (i.e.
+  a CMD shell that was defined outside of the current step block (i.e. 
   as a job or workflow default) (#1418)
-
+  
 * Fixed a bug where the `opentofu` ecosystem was not recognized in
   Dependabot configuration files (#1452)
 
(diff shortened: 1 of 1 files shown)


crates/github-actions-expressions/src/call.rs
```diff
@@ -169,8 +169,8 @@ impl<'src> Call<'src> {
             | Evaluation::Number(_)
             | Evaluation::Boolean(_)
             | Evaluation::Null => {
-                let search_str = search.sema().to_string().to_lowercase();
-                let item_str = item.sema().to_string().to_lowercase();
+                let search_str = search.sema().to_string().to_uppercase();
+                let item_str = item.sema().to_string().to_uppercase();
```
woodruffw, line 173: Am I missing something here? This should already be case insensitive; this just changes the comparison from lowercase normalized to uppercase.

crates/github-actions-expressions/src/call.rs
```diff
@@ -210,8 +210,8 @@ impl<'src> Call<'src> {
                 | Evaluation::Null,
             ) => {
                 // Case-insensitive comparison
-                let string_str = search_string.sema().to_string().to_lowercase();
```
woodruffw, line 213: Let's keep all case comparisons are lowercase-normalized; IME that's more common (and there's a potential micro-optimization for the runtime, since lowercase characters are more common than uppercase ones in most languages).

crates/github-actions-expressions/src/call.rs
```diff
@@ -654,7 +654,7 @@ mod tests {
             ),
             (
                 "contains(fromJSON('[\"a\", \"b\", \"c\"]'), 'B')",
-                Evaluation::Boolean(false), // Array search is exact match, not case-insensitive
+                Evaluation::Boolean(true), // Array search uses == semantics, which is case-insensitive for strings
```
woodruffw, line 657: TIL 🙂

crates/github-actions-expressions/src/call.rs
DarkaMaul: The idea here was to handle the [greek](https://github.com/zizmorcore/zizmor/pull/1687/files/6a17927883beb8b28c6cc706f9963f1b88c4595c#diff-5f7ae3b71414e1eb3db14d31c5f57101a69fd0945aa4f706cece0a11ef512c6e) sigma - where to lowercase would not properly handle the equality.

This is described [here](https://github.com/actions/languageservices/blob/main/expressions/testdata/operators_case_insensitive.json#L91)

docs/release-notes.md
```diff
@@ -160,7 +163,7 @@ of `zizmor`.
 
     Users who with to retain the old (permissive policy) for first-party
     actions may configure it explicitly in their `zizmor.yml`:
-  
+
```
DarkaMaul: FLAG: sorry for this change; I'll revert it, that's a spurious config in my editor I can't track.

crates/github-actions-expressions/src/call.rs
woodruffw: Oh right, I forgot Greek has that last-letter casing rule. Hmm, are we sure that there aren't alphabets/languages with similar upper-case specializations?

(TIL about this test suite in languageservices, I hadn't seen that before! It'd be really great to have that integrated into zizmor's expression tests.)

woodruffw: (I'm fine with changing to uppercase for now though, thanks for explaining!)

woodruffw:
Thanks!

PR #1720, opened 2026-03-08
feat: support `-` for reading input from stdin

### LINKED ISSUE #1174: CLI: support `-` for reading from stdin

koriyoshi2041:
## Pre-submission checks

## Summary

Add support for `zizmor -` to read a single workflow, action, or Dependabot config from stdin. The input type is automatically inferred by trying each parser in order (workflow -> action -> dependabot).

Key design decisions:

- `-` must be the only input; combining with other inputs is rejected at the CLI level
- `--fix` is incompatible with stdin since there's no file to write back to
- New `InputKey::Stdin(StdinKey)` variant avoids reusing `InputKey::Local` with a synthetic path, preserving the invariant that `Local` always refers to a real filesystem path
- All parse errors are collected and logged so users get context about every failed type attempt, not just a single misleading error (addresses [review feedback](https://github.com/zizmorcore/zizmor/pull/1611#discussion_r2011736281) from #1611)
- Invalid YAML short-circuits immediately (no point trying other parsers)
- No local config discovery for stdin; uses `Config::default()`
- Usage docs updated to document `-` as a fourth input source

Closes #1174

## Test Plan

- [x] `echo 'on: push ...' | zizmor -` parses as workflow (test_stdin_workflow)
- [x] `echo 'name: ... runs: ...' | zizmor -` parses as action (test_stdin_action)
- [x] `echo 'version: 2 ...' | zizmor -` parses as dependabot (test_stdin_dependabot)
- [x] `zizmor - ./other-input` fails with clear error (test_stdin_with_other_inputs)
- [x] `zizmor --fix -` fails with clear error (test_stdin_with_fix)
- [x] Empty stdin produces collection error with all parse failures logged (test_stdin_empty)
- [x] Invalid YAML on stdin warns and exits (test_stdin_invalid_yaml)
- [x] Invalid YAML with `--strict-collection` fails with syntax error (test_stdin_invalid_yaml_strict)
- [x] All 199 existing tests continue to pass (207 total with 8 new)
- [x] `cargo fmt --check` clean
- [x] `cargo clippy -p zizmor --all-targets` clean (no new warnings)

### DIFF
--- a/crates/zizmor/src/main.rs
+++ b/crates/zizmor/src/main.rs
@@ -231,6 +231,8 @@ struct App {
     /// (typically `action.yml`), entire directories, or a `user/repo` slug
     /// for a GitHub repository. In the latter case, a `@ref` can be appended
     /// to audit the repository at a particular git reference state.
+    ///
+    /// Use `-` to read a single input from stdin.
     #[arg(required = true)]
     inputs: Vec<String>,
 }
@@ -792,6 +794,28 @@ async fn run(app: &mut App) -> Result<ExitCode, Error> {
 
     eprintln!("🌈 zizmor v{version}", version = env!("CARGO_PKG_VERSION"));
 
+    // Validate stdin input constraints: `-` must be the only input,
+    // and cannot be combined with `--fix`.
+    if app.inputs.contains(&"-".to_string()) {
+        if app.inputs.len() > 1 {
+            let mut cmd = App::command();
+            cmd.error(
+                clap::error::ErrorKind::ArgumentConflict,
+                "`-` (stdin) cannot be combined with other inputs",
+            )
+            .exit();
+        }
+
+        if app.fix.is_some() {
+            let mut cmd = App::command();
+            cmd.error(
+                clap::error::ErrorKind::ArgumentConflict,
+                "`--fix` cannot be used with `-` (stdin)",
+            )
+            .exit();
+        }
+    }
+
     let collection_mode_set = CollectionModeSet::from(app.collect.as_slice());
 
     let min_severity = match app.min_severity {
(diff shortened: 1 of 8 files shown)


woodruffw:
Please don't submit AI-generated PRs on top of existing PRs, whether those are AI generated or not. This is not reviewable.

Issue #1577, opened 2026-01-26
Feature: How to report duplicate results due to anchors


DarkaMaul:
### Pre-submission checks

- [x] I am **not** reporting a bug (crash, false positive/negative, etc). These must be filed via the bug report template.
- [x] I have looked through both the [open](https://github.com/zizmorcore/zizmor/issues?q=is%3Aissue+is%3Aopen+) and [closed](https://github.com/zizmorcore/zizmor/issues?q=is%3Aissue%20state%3Aclosed) issues for a duplicate request.

### What's the problem this feature will solve?

Considering the following workflow:

```yaml
jobs:
  basic:
    name: basic
    runs-on: ubuntu-latest
    steps:
      - run: &run |
          echo "doing a thing: ${{ github.event.head_commit.message }}"

      - run: *run
```

The current zizmor output is like this:

<img width="829" height="344" alt="Image" src="https://github.com/user-attachments/assets/57f85927-c658-4efb-b140-70cbd71e2e04" />

The finding is duplicated, creating a suboptimal user experience.

With the additional efforts to support `anchors` as first class YAML features citizens in zizmor, we could probably do something to improve it.

### Describe the solution you'd like

I see some options:
- Report at the anchor usage (*run) with a backtrace to the anchor definition (the one I like the most)
- Report only at the anchor usage (if that makes sense in the context)
- Keep it like this (the problem is indeed only at the anchor definition), but remove the duplicate

woodruffw:
Thanks for enumerating these @DarkaMaul!

Option (a) is probably the nicest from a user/UX perspective, but it's also probably the most intrusive in terms of changes needed (IIUC, zizmor's core would need to become anchor-aware, which it currently isn't). So I like the idea, but I'm worried about the architectural implications 🙂 

OTOH Option (c) seems pretty reasonable to me in terms of complexity -- deduplicating any findings that have the exact same ID + source locations should be relatively straightforward, and shouldn't be disruptive to users.

PR #1732, opened 2026-03-09
Fix #1681

### LINKED ISSUE #1681: [BUG]: `template-injection` crashes on YAML alias used as a scalar `run` value
### Pre-submission checks

### zizmor version

main

### Expected behavior

`zizmor` should audit the workflow successfully not crash. The workflow is valid GitHub Actions YAML.

### Actual behavior

`zizmor` crashes with a fatal error:

```
fatal: no audit was performed
'template-injection' audit failed on file:///tmp/test-case.yml

Caused by:
    0: error in 'template-injection' audit
    1: syntax node `flow_node` is missing child field `key`
```

### Reproduction steps

Run `zizmor` on the following workflow

```yaml
name: Build and Test

on:
  push:

jobs:
  build:
    runs-on: ubuntu-latest
    steps:
      - name: Print build info
        run: &print-info echo "Building ref ${{ github.ref }} triggered by ${{ github.actor }}"

      - name: Print build info again
        run: *print-info
```

### Logs

```plain text
DEBUG audit: zizmor: scheduling superfluous-actions on file:///tmp/yaml-anchor-alias.yml
 WARN audit: zizmor: one or more inputs contains YAML anchors; you may encounter crashes or unpredictable behavior
 WARN audit: zizmor: for more information, see: https://docs.zizmor.sh/usage/#yaml-anchors
fatal: no audit was performed
'template-injection' audit failed on file:///tmp/yaml-anchor-alias.yml

Caused by:
    0: error in 'template-injection' audit
    1: syntax node `flow_node` is missing child field `key`
```

### Additional context

Found in the wild in [bitcoin/bitcoin](https://github.com/bitcoin/bitcoin/blob/6b0a980de94bc5b64703151be708991f954161ea/.github/workflows/ci.yml#L165) with a valid GitHub Actions [run](https://github.com/bitcoin/bitcoin/actions/runs/22578591853).

Reproduced the trigger from this issue in this [run](https://github.com/trail-of-forks/zizmor/actions/runs/22585306417/job/65428418569)

DarkaMaul:
## Pre-submission checks

Please check these boxes:

      one first).

## Summary

When a YAML alias (*foo) referenced an anchored block, `zizmor` would crash. The `KeyOnly` query mode was navigating parent nodes from the resolved anchor definition site instead of the alias reference site, landing on a node type that doesn't have a key field.  

## Test Plan

- Added a reproducer in the anchors tests

Fixes #1681

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -58,11 +58,12 @@ checksum = "683d7910e743518b0e34f1186f92494becacb047c7b6bf616c96772180fef923"
 
 [[package]]
 name = "annotate-snippets"
-version = "0.12.11"
+version = "0.12.13"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "16e4850548ff4a25a77ce3bda7241874e17fb702ea551f0cc62a2dbe052f1272"
+checksum = "74fc7650eedcb2fee505aad48491529e408f0e854c2d9f63eb86c1361b9b3f93"
 dependencies = [
  "anstyle",
+ "memchr",
  "unicode-width 0.2.2",
 ]
 
@@ -73,7 +74,22 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "43d5b281e737544384e969a5ccad3f1cdd24b48086a0fc1b2a5262a26b8f4f4a"
 dependencies = [
  "anstyle",
- "anstyle-parse",
+ "anstyle-parse 0.2.7",
+ "anstyle-query",
+ "anstyle-wincon",
+ "colorchoice",
+ "is_terminal_polyfill",
+ "utf8parse",
+]
+
+[[package]]
+name = "anstream"
+version = "1.0.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "824a212faf96e9acacdbd09febd34438f8f711fb84e09a8916013cd7815ca28d"
+dependencies = [
+ "anstyle",
+ "anstyle-parse 1.0.0",
  "anstyle-query",
  "anstyle-wincon",
  "colorchoice",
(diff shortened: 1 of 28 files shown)


crates/yamlpath/src/lib.rs
```diff
@@ -854,7 +854,11 @@ impl Document {
 
         // Our focus node might be an alias, in which case we need to
         // do one last leap to get our "real" final focus node.
-        // TODO(ww): What about nested aliases?
```
DarkaMaul, line 857: FLAG: I did not want to open a PR just for this, but I believe you can't have nested aliases - so I deleted the comment.

crates/yamlpath/src/lib.rs
```diff
@@ -874,6 +878,17 @@ impl Document {
             }
             _ => focus_node,
         };
+        let was_alias = pre_resolution_node.id() != focus_node.id();
+
+        // For Pretty and KeyOnly modes, we need the structural position
+        // in the tree. When an alias was resolved, use the pre-resolution
+        // node so we navigate the alias reference site, not the anchor
+        // definition site.
+        let structural_node = if was_alias {
```
DarkaMaul, line 887: FLAG: I don't like `structural_node`, but both my agent and I could not find a better name - WDYT?

(some ideas: navigation node, reference node ...)

crates/zizmor/tests/integration/e2e/anchors.rs
```diff
@@ -31,15 +31,37 @@ fn test_basic() -> Result<()> {
...
     13 |           "doing a thing: ${{ github.event.issue.title }}"
        |                               ^^^^^^^^^^^^^^^^^^^^^^^^ may expand into attacker-controllable code
+    14 |
+    15 |       - run: *run
+       |         --- this run block
+       |
+       = note: audit confidence → High
+       = note: this finding has an auto-fix
+
+    error[template-injection]: code injection via template expansion
```
DarkaMaul, line 43: FLAG: We will need #1577 to remove the duplicates from here.

crates/yamlpath/src/lib.rs
woodruffw: Hmm. `structural_node` seems fine to me. At least, I can't think of a better name either 🙂

Issue #1741, opened 2026-03-10
[BUG]:  zizmor recompiles itself without code changes causing unnecessary recompilation


DarkaMaul:
### Pre-submission checks

- [x] I am **not** filing a feature request. These should be filed via the feature request form instead.
- [x] I have checked the [Troubleshooting Guide](https://docs.zizmor.sh/troubleshooting) for my problem.
- [x] I have looked through both the [open](https://github.com/zizmorcore/zizmor/issues?q=is%3Aissue+is%3Aopen+) and [closed](https://github.com/zizmorcore/zizmor/issues?q=is%3Aissue%20state%3Aclosed) issues for a duplicate report.

### zizmor version

main

### Expected behavior

Running `cargo test` (or `cargo build`) twice in a row with no source changes should not recompile the `zizmor` crate on the second invocation.

### Actual behavior

The `zizmor` crate is recompiled on every `cargo test` / `cargo build` invocation, even when no files have changed.

### Reproduction steps

1. Clone the repository and run `cargo build`
2. Run `cargo build` again immediately, with no changes
3. Observe that `Compiling zizmor v1.23.1` appears in the output again

### Logs

```plain text
$ cargo test "e2e::anchors" -p zizmor --test integration 2>&1 | head -3
   Compiling zizmor v1.23.1 (/path/to/crates/zizmor)
    Finished `test` profile [unoptimized + debuginfo] target(s) in 21.68s
     Running tests/integration/main.rs

$ cargo test "e2e::anchors" -p zizmor --test integration 2>&1 | head -3
   Compiling zizmor v1.23.1 (/path/to/crates/zizmor)
    Finished `test` profile [unoptimized + debuginfo] target(s) in 22.59s
     Running tests/integration/main.rs
```

woodruffw:
I kind of suspect this is because of the build.rs file we use to insert generated data. I'm not sure if there's a clean way to get cargo to _not_ run that build script, since (IIUC) it needs to run the script to determine what (if any) invalidations have occurred.

woodruffw:
#1742

PR #1742, opened 2026-03-10
fix: Fix recompilation of zizmor

DarkaMaul:
## Pre-submission checks

Please check these boxes:

      one first). (#1741)

      use that violates the policy.

## Summary

This change seems to fix the recompilation even if source file do not change:

Measured with `cargo run -p zizmor -- --version` after a clean build:

| Scenario | Before fix | After fix |
|---|---|---|
| Cached run | **21.3s** | **0.5s** |

## Test Plan

Run `cargo run -- --version` twice in a row, before an after the fix.

## Note

The fix was found by Claude, and it works in practice _but_ I can't find any indications of it in the [manual](https://doc.rust-lang.org/cargo/reference/build-scripts.html) (except that all examples use `println` instead of `print`)

### DIFF
--- a/docs/release-notes.md
+++ b/docs/release-notes.md
@@ -18,11 +18,6 @@ of `zizmor`.
 
     Many thanks to @shaanmajid for proposing and implementing this improvement!
 
-### Bug Fixes 🐛
-
-* Fixed a bug in `build.rs` causing `zizmor` to recompile on every build even without
-  code changes (#1742)
-
 ## 1.23.1
 
 ### Bug Fixes 🐛


docs/release-notes.md
```diff
@@ -18,6 +18,11 @@ of `zizmor`.
 
     Many thanks to @shaanmajid for proposing and implementing this improvement!
 
+### Bug Fixes 🐛
+
```
woodruffw: This change doesn't need a note IMO 🙂

DarkaMaul: Reverted!

PR #1756, opened 2026-03-12
fix: allow empty images

miketheman:
## Pre-submission checks

Please check these boxes:

      one first).

      use that violates the policy.

## Summary

Two commits, one fixing the current false positive detection for a valid input, and the other addressing the need for more specificity in the string to handle conditionals.

## Test Plan

Three test inputs have been added, and only a single pedantic finding is surfaced, instead of the current behavior which would surface low confidence false positives.

### DIFF
--- a/docs/release-notes.md
+++ b/docs/release-notes.md
@@ -49,6 +49,11 @@ of `zizmor`.
 * The [ref-version-mismatch] audit now uses a more useful audit description
   for its findings (#1843)
 
+* The [unpinned-images] audit now produces more precise findings for
+  image references that are computed through expressions (#1756)
+
+    Many thanks to @miketheman for implementing this improvement!
+
 ### Bug Fixes 🐛
 
 * Fixed a bug where the [concurrency-limits] audit reported findings


crates/zizmor/src/audit/unpinned_images.rs
```diff
@@ -7,12 +7,41 @@ use crate::{
...
+            Some(values)
+        }
+        Expr::BinOp {
+            op: BinOp::And,
+            rhs,
+            ..
+        } => extract_image_values(rhs),
+        _ => None,
+    }
+}
```
woodruffw: I _think_ this is a good candidate for the `github-actions-expressions` crate APIs: it generalizes to something like `fn leaf_expressions(...) -> Vec<SpannedExpr<'src>>`, i.e. all possible leaf expressions of a given expression. We should support things other than string too, since we would ideally emit two findings for something like this:

```yaml
image: ${{ foo.abc || foo.def }}
```

(Returning `SpannedExpr` also allows us to provide more precise finding subfeatures.)

miketheman: Thanks for the feedback! I moved the core of the logic to `github-actions-expressions` and call it in unpinned.

crates/zizmor/tests/integration/audit/unpinned_images.rs
```diff
@@ -7,7 +7,7 @@ fn test_pedantic_persona() -> anyhow::Result<()> {
             .input(input_under_test("unpinned-images.yml"))
             .args(["--persona=pedantic"])
             .run()?,
-        @r"
+        @"
```
miketheman, line 10: This kept coming in via `cargo fmt` so I left it in.

PR #1759, opened 2026-03-13
feat(secrets-outside-env): allowlist secrets in config

### LINKED ISSUE #1750: Feature: allowlist exceptions to secrets-outside-env in config file
### Pre-submission checks

The `secrets-outside-env` audit can be useful, but in exceptional cases, secrets are unrelated to deployments/environments. The false positives make it tricky to work with.

The audit could be more user-friendly, if it could be configured for the exceptional cases, e.g. allowlisting certain secrets in some way (aside from the default GITHUB_TOKEN).

Ideally, I could configure zizmor.yml with a per-audit configuration, e.g. to allow-list a specific org secret that doesn't relate to deployments.

### Additional context

Examples include:
* third-party CI services (e.g. code coverage)
* actions that need to query github (e.g. zizmor action) from GHE-DR

For more details, see this [discussion](https://github.com/orgs/zizmorcore/discussions/1749)

rmuir:
## Pre-submission checks

Please check these boxes:

      one first).

      use that violates the policy.

No LLM but very green with rust: possibly worse :)

## Summary

The `secrets-outside-env` audit gains an `allow` list in the configuration file of secret names that should be ignored by the audit.

Closes #1750

## Test Plan

Tests attempt to mimic similar tests of other audits that have audit-specific configuration. I created three test configurations, allowing 0, 1, or 2 secrets. There's a new test document that attempts to access 2 secrets, plus the `GITHUB_TOKEN`, outside of an environment.

### DIFF
--- a/support/zizmor.schema.json
+++ b/support/zizmor.schema.json
@@ -246,8 +246,7 @@
           "default": [],
           "items": {
             "type": "string"
-          },
-          "uniqueItems": true
+          }
         }
       },
       "additionalProperties": false


crates/zizmor/src/audit/secrets_outside_env.rs
```diff
@@ -67,6 +69,14 @@ impl Audit for SecretsOutsideEnvironment {
                     continue;
                 }
 
+                // If the user has configured the audit, check secret's name against the allowlist.
+                if let Some(config) = config
+                    && let Some(secret_name) = context.single_tail()
+                    && config.allow.contains(secret_name)
```
woodruffw: I _think_ we need to do something a little more general here, since `contains` will be case-sensitive but contexts themselves aren't, e.g. `secrets.foo` and `secrets.FOO` should both match `FOO` in the allowlist.

I think the easiest way to do that would be to have `SecretsOutsideEnvConfig` store a `HashSet<ContextPattern>` and then use `ContextPattern::matches` to determine if there's a match between the config and a secrets reference 🙂

crates/zizmor/src/audit/secrets_outside_env.rs
```diff
@@ -67,6 +69,14 @@ impl Audit for SecretsOutsideEnvironment {
                     continue;
                 }
 
+                // If the user has configured the audit, check secret's name against the allowlist.
+                if let Some(config) = config
```
woodruffw: Thinking out loud: it's not necessary for this PR, but it'd be cool if we could collapse this state by having `impl Default for Config` and then having the default config just be `secrets.GITHUB_TOKEN`.

crates/zizmor/tests/integration/test-data/secrets-outside-env/multiple-secrets.yml
```diff
@@ -0,0 +1,18 @@
...
+name: multiple-secrets
+
+jobs:
+  test:
+    name: test
+    runs-on: ubuntu-latest
+    permissions: {}
+    steps:
+      - name: vulnerable-1
+        run: echo ${{ secrets.NOT_SO_SECRET }}
```
woodruffw: Nit: tests for `secrets.not_so_secret` and `secrets['not_so_secret']` would also be great here, just to ensure we're handling those kinds of edge cases!

(The `ContextPattern` APIs will do that for us.)

crates/zizmor/src/audit/secrets_outside_env.rs
rmuir: thanks for pointing this out: totally forgot about these weird quirks of actions!

I will look into the ContextPattern some more... at a glance it seemed a bit daunting to try to implement hashing support correctly for it. I could be missing something here.

There's also a related problem: the configuration here for this allowlist will currently allow "any string"... but ideally these strings would be validated as actions expressions identifier? Would be nice if JSON schema could reflect that too, catch any issues in the editor.

Since this is new configuration element... one alternative way to solve it, maybe config could just require `[A-Z_][A-Z0-9_-]*` (uppercase identifiers) for allowlist elements? If we allow lowercase, it is a bit confusing from JSON schema perspective (the `"uniqueItems": true` makes less sense).

woodruffw: > at a glance it seemed a bit daunting to try to implement hashing support correctly for it. I could be missing something here.

You can add a derive for `Hash` to the type to have Rust do it for you 🙂 -- it should be able to derive without any issues, since it's just a newtype over a string.

> but ideally these strings would be validated as actions expressions identifier? Would be nice if JSON schema could reflect that too, catch any issues in the editor.

Yeah, ideally we would encode this as part of the deserialization. Doing that with `ContextPattern` will be a bit clunky though, since that accepts a subset of valid context syntaxes. I'll need to think about that, but skipping that validation on the config seems fine for an MVP here.

rmuir: thank you for your patience... will take another stab tonight.

When testing, I found another inconsistency with json schema: in the other direction. jsonschema specifies `uniqueItems` (due to HashSet usage), and the editor will respect that and show the user an error if they have duplicates. But zizmor CLI will not raise an error on duplicates (I guess because schema support is an optional feature not used?)

woodruffw: No worries! I appreciate your effort on this.

> But zizmor CLI will not raise an error on duplicates (I guess because schema support is an optional feature not used?)

Yeah, this is because the JSON schema isn't actually taken into consideration when loading the config: the config gets loaded via serde, which has format/implementation-specific behavior for things like duplicate keys.

We could probably further restrict this with `serde_with` or similar (or with our own small helper).

rmuir: > You can add a derive for `Hash` to the type to have Rust do it for you 🙂 -- it should be able to derive without any issues, since it's just a newtype over a string.

I think that's my issue, it is just a newtype over a string and the map won't handle case-insensitivity if I derive these.

For now (at least to have correct behavior and tests), I pushed a commit to use simple lowercasing. I know this isn't what you want... I really tried to use the ContextPattern but in every attempt it did not look... good. I think it is just a limitation on my side.

woodruffw: No worries, I can take a look tonight.

woodruffw: I poked at this a bit and I agree that it's going to be nontrivial to use the `ContextPattern` APIs here. So I think it's fine to just do the simple lowercasing for now and move forwards with that!

PR #1766, opened 2026-03-16
fix: preserve env var case in template injection fixes

### LINKED ISSUE #1765: [BUG]: template-injection fix uppercases `env.*` variable names
### Pre-submission checks

### zizmor version

1.23.1

### Expected behavior

Fixing `${{ env.build_tag }}` should produce `${build_tag}`, preserving the original case since the variable already exists.

### Actual behavior

The fix uppercases it to `${BUILD_TAG}`, which breaks the workflow on Linux where env vars are case-sensitive.

### Reproduction steps

1. Create a workflow file `repro.yml`:

```yaml
name: Build
on:
  pull_request:

jobs:
  build:
    runs-on: ubuntu-latest
    steps:
      - name: Set build info
        run: |
          echo "build_tag=$(git describe --tags)" >> "$GITHUB_ENV"

      - name: Use build info
        run: |
          echo "Building ${{ env.build_tag }}"
```

2. Run `zizmor --fix=all repro.yml`

3. The result:

```yaml
      - name: Use build info
        run: |
          echo "Building ${BUILD_TAG}"
```

`build_tag` was renamed to `BUILD_TAG`.

### Logs

```plain text

```

### Additional context

The uppercasing in `context_to_env_var()` makes sense for non-`env` contexts where zizmor is creating a new variable, but for `env.*` contexts the variable already exists and its name should be preserved. GitHub also recommends treating env vars as case-sensitive regardless of OS ([docs](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-commands#setting-an-environment-variable)).

---

**Disclosure:** This issue was drafted with AI assistance (Claude Code).

I intend to submit a PR with a fix for this.

gcl-sekoia:
## Pre-submission checks

Please check these boxes:

      one first).

      use that violates the policy.

## Summary

Fixes #1765.

When fixing template injections, `context_to_env_var()` unconditionally uppercases all generated env var names. For `env.*` contexts this changes the variable name (e.g. `env.build_tag` → `BUILD_TAG`), breaking workflows on operating systems or shells where env vars are case-sensitive.

This PR skips the uppercasing for `env` contexts. Non-env contexts (e.g. `github.actor` → `GITHUB_ACTOR`) are still uppercased.

## Test Plan

Updated `test_context_to_env_var` cases to expect preserved case. Updated `test_template_injection_fix_equivalent_expressions` snapshot accordingly. All 37 template injection tests pass.

Verified manually against the reproducer from #1765: `${{ env.build_tag }}` now produces `${build_tag}` instead of `${BUILD_TAG}`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1559,9 +1559,9 @@ dependencies = [
 
 [[package]]
 name = "jsonschema"
-version = "0.44.1"
+version = "0.45.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f98b64a413c93a1b413dfbaf973b78d271648b9cae50b10302ad88af78991672"
+checksum = "6f29616f6e19415398eb186964fb7cbbeef572c79bede3622a8277667924bbe3"
 dependencies = [
  "ahash",
  "bytecount",
@@ -2260,9 +2260,9 @@ dependencies = [
 
 [[package]]
 name = "referencing"
-version = "0.44.1"
+version = "0.45.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "22952642836711d7a730d23a4dfb0d732e75a85e4c4f5704266d9c8fac278ff1"
+checksum = "b8a618c14f8ba29d8193bb55e2bf13e4fb2b1115313ecb7ae94b43100c7ac7d5"
 dependencies = [
  "ahash",
  "fluent-uri 0.4.1",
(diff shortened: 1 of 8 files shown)


crates/zizmor/src/audit/template_injection.rs
```diff
@@ -1059,7 +1069,7 @@ jobs:
                           echo "User: ${GITHUB_ACTOR}"
                           echo "User: ${GITHUB_ACTOR}"
                           echo "User: ${GITHUB_ACTOR}"
-                          echo "User: ${GITHUB_ACTOR}"
+                          echo "User: ${github_actor}"
```
gcl-sekoia, line 1072: This snapshot change reflects the case-preserving behavior: `env['github_actor']` now produces `${github_actor}` instead of `${GITHUB_ACTOR}`.

I wasn't sure about the best way to update this test. These expressions aren't actually equivalent at runtime: the `env` context only contains variables explicitly set via `env:` blocks or `GITHUB_ENV`, not the default runner environment variables, so the `env.*` variants here all resolve to empty strings ([verified on GHA](https://github.com/gcl-sekoia/zizmor/actions/runs/23151437895/job/67254099801)). Happy to adjust if you'd prefer a different approach.

woodruffw: This seems fine -- this isn't intended to be a real-world workflow, just a minimal set of scenarios exercising the audit's variable logic 🙂

woodruffw (approved):

woodruffw:
Thanks @gcl-sekoia!

gcl-sekoia:
Thank you @woodruffw! Zizmor is a fantastic tool!

PR #1772, opened 2026-03-19
fix(github-actions-models): deserialize dynamic with

### LINKED ISSUE #1769: [BUG]: unable to parse valid yaml/github workflow with dynamic `with` options
### Pre-submission checks

### zizmor version

v1.23.1

### Expected behavior

To not throw an error parsing a valid yaml/workflow file

### Actual behavior

  fatal: no audit was performed
  error: failed to load file://.github/workflows/build.yml as workflow
    |

  Caused by:
      0: failed to load file://.github/workflows/build.yml as workflow
      2: jobs: data did not match any variant of untagged enum Job at line 10 column 3

### Reproduction steps

We build a dynamic `with` options for one of our custom actions:

```yaml
name: Build Packages
run-name: Building Package for ${{ github.ref_name }}

on:
  push:
    tags:
      - "*/v*"

jobs:
  check-configs:
    name: Checking Build Request
    runs-on: ubuntu-24.04

    outputs:
      package: ${{ steps.parameters.outputs.package }}
      version: ${{ steps.parameters.outputs.version }}
      runners: ${{ steps.parameters.outputs.runners }}

    steps:
      - name: Checkout Code
        uses: actions/checkout@v6

      - name: Get Parameters
        id: parameters
        run: |
          package="${GITHUB_REF_NAME%/*}"

          if [[ ! -e "${package}" || ! -e "${package}/package_options.yml" ]]; then
            echo "Invalid Build Request, there is no ${package} folder or 'package_options.yml' is missing"
            exit 1
          fi

          cd "${package}"

          PKG_TYPE=$(yq '.package-type' package_options.yml)
          DEFAULT_VALUE=$([[ "${PKG_TYPE}" == "generic" || "${PKG_TYPE}" == "cockpit" ]] && echo "true" || echo "false")

          # null evaluates to empty string in yq
          OS_IND=$(yq '.os-independent' package_options.yml)
          ARCH_IND=$(yq '.arch-independent' package_options.yml)

          [ "${OS_IND}" == "null" ] && OS_IND="${DEFAULT_VALUE}"
          [ "${ARCH_IND}" == "null" ] && ARCH_IND="${DEFAULT_VALUE}"

          if "${OS_IND}"; then
            runners=(focal-arm64-ephemeral)
          else
            runners=(focal-arm64-ephemeral jammy-arm64-ephemeral noble-arm64-ephemeral)
          fi

          if ! "${ARCH_IND}"; then
            for pkg in "${runners[@]}"; do
              runners+=("$(sed 's/arm64/amd64/' <<< "${pkg}")")
            done
          fi

          # Convert array to JSON array format
          json_runners="[$(printf '"%s"\n' "${runners[@]}" | paste -sd ',' -)]"

          {
            echo "runners=${json_runners}"
            echo "package=${package}"
            echo "version=${GITHUB_REF_NAME#*/[vV]}"
          } | tee -a "${GITHUB_OUTPUT}"

  build-package:
    needs: check-configs
    name: Build Package

    strategy:
      matrix:
        os: ${{ fromJSON(needs.check-configs.outputs.runners) }}

    runs-on: ${{ matrix.os }}

    outputs:
      PREFIX: ${{ steps.setup.outputs.prefix }}
      VERSION: ${{ steps.build_package.outputs.version-number }}

    steps:
      - name: Checkout Code
        uses: actions/checkout@v6
        with:
          sparse-checkout: "/${{ needs.check-configs.outputs.package }}"
          sparse-checkout-cone-mode: false
          fetch-depth: 0


      - name: Package Setup
        id: setup
        env:
          PACKAGE: ${{ needs.check-configs.outputs.package }}
        run: |
          tar cvf - --remove-files -C "${PACKAGE}/" . | tar xvf -
          echo "PREFIX=${PACKAGE}/" | tee -a "${GITHUB_ENV}"
          {
            echo "prefix=${PACKAGE}/"
            echo 'package_options<<EOF'
            yq -o json < package_options.yml
            echo EOF
          } >> "${GITHUB_OUTPUT}"

      - name: Build Package
        id: build_package
        uses: custom/build-action@v1
        with: ${{ fromJson(steps.setup.outputs.package_options) }}
```

Zizmore blows up at the `with` statement, I assume because generally they are supposed to be nested and yet here they must be inline with the `uses` statement.

### Logs

```plain text

```

### Additional context

_No response_

reubenwong97:
## Pre-submission checks

Please check these boxes:

- [x] **Mandatory**: This PR corresponds to an issue: #1769 .

      use that violates the policy.

## Summary

Updates `with` from expecting a literal to something that can be an expression. This is mainly just to prevent zizmor from crashing due to failure to deserialize and does not perform any analysis. Therefore, at various pattern matching sites, it requires matching the `Literal` variant.

## Test Plan

Added a minimal example and ensures it is deserialized by the `test_load_all` correctly.

Closes #1769


reubenwong97:
Hi @woodruffw ,

Can you take a look at this fix attempt when you have some time? Thanks!

crates/github-actions-models/tests/sample-workflows/zizmor-issue-1769.yml
```diff
@@ -0,0 +1,12 @@
...
+name: Test
+on: [push]
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: custom/build-action@v1
+        with:
+          input-a: ${{ fromJSON(steps.setup.outputs.package_options).input-a }}
+          input-b: ${{ fromJSON(steps.setup.outputs.package_options).input-b }}
```
woodruffw: I think this needs to be:

```
with: ${{ fromJSON(steps.setup.outputs.package_options) }}
```

...to properly exercise the expression case here 🙂

docs/release-notes.md
```diff
@@ -28,6 +28,8 @@ of `zizmor`.
 * Fixed a bug where auto-fixes for the [template-injection] audit would fail
   to preserve an environment variable's casing (#1766)
 
+* Fixed a bug where dynamic `with` blocks fail to deserialize (#1772)
```
woodruffw:
```suggestion
* Fixed a bug where `with: ${{ expr }}` clauses would cause a crash (#1772)
```

crates/zizmor/src/audit/artipacked.rs
```diff
@@ -101,7 +104,7 @@ impl Artipacked {
         for step in steps {
             let StepBodyCommon::Uses {
                 uses: Uses::Repository(uses),
-                with,
+                with: LoE::Literal(with),
```
woodruffw: I think we'll need to handle the `LoE::Expr` variants here and below as well -- I _suspect_ we should emit blanket auditor (or maybe pedantic?) level findings when the entire `with:` clause is unanalyzable.

crates/zizmor/src/models/coordinate.rs
```diff
@@ -51,7 +51,7 @@ impl ActionCoordinate {
 
         let StepBodyCommon::Uses {
             uses: Uses::Repository(uses),
-            with,
+            with: LoE::Literal(with),
```
woodruffw: Same here -- for coordinate matching purposes, I think a `with: ${{ expr }}` should be treated as `Usage::ConditionalOptIn`, since it _may_ satisfy the pattern (but we don't know).

reubenwong97:
> Thanks @reubenwong97! Looks good as a start, I flagged some places where we'll need to handle this new variant.
> 
> (We should also add new test cases for each of those locations.)

Hi @woodruffw , can you help check again? Thanks!

crates/zizmor/src/audit/artipacked.rs
```diff
@@ -107,6 +110,27 @@ impl Artipacked {
...
+                    findings.push(
+                        Self::finding()
+                            .severity(Severity::Informational)
+                            .confidence(Confidence::High)
+                            .persona(Persona::Pedantic)
+                            .add_location(
+                                step.location()
+                                    .primary()
+                                    .with_keys(["with".into()])
+                                    .annotated("dynamic `with:` clause cannot be analyzed"),
```
woodruffw: I'm not 100% sure, but I think we probably don't want to propagate the "dynamic `with:`" terminology here and elsewhere. Instead, I think to just say something like:


```suggestion
                                    .annotated("may not set persist-credentials: false"),
```

crates/zizmor/src/audit/artipacked.rs
```diff
@@ -107,6 +110,27 @@ impl Artipacked {
...
 
+            let with = match with {
+                LoE::Literal(with) => with,
+                // Emit blanket pedantic finding if the `with:` block cannot be analyzed
+                LoE::Expr(_) => {
+                    findings.push(
+                        Self::finding()
+                            .severity(Severity::Informational)
+                            .confidence(Confidence::High)
+                            .persona(Persona::Pedantic)
```
woodruffw, line 122: Nit: we should add a location (with subfeature) for the `uses:` clause too, to make the render here friendlier 🙂

crates/zizmor/src/audit/template_injection.rs
```diff
@@ -119,7 +119,7 @@ impl TemplateInjection {
         match step.body() {
             models::StepBodyCommon::Uses {
                 uses: Uses::Repository(uses),
-                with,
+                with: LoE::Literal(with),
```
woodruffw, line 122: Thinking out loud: I kind of wonder if we should flag the entire `with:` clause here (with a pedantic finding?), since an action with injection-prone inputs might be exploitable when the entire `with:` clause comes from an expression.

woodruffw:
Another random thought: maybe it makes sense to flag `with: ${{ expr }}` as another finding in the `obfuscation` audit? It arguably fits the criteria for that audit, since it limits our analysis capabilities. WDYT @reubenwong97?

crates/zizmor/src/audit/artipacked.rs
reubenwong97: Thanks @woodruffw , can you check if this works?

crates/zizmor/src/audit/template_injection.rs
reubenwong97: Flagged it as another finding (change below), does it look alright?

reubenwong97:
> Another random thought: maybe it makes sense to flag `with: ${{ expr }}` as another finding in the `obfuscation` audit? It arguably fits the criteria for that audit, since it limits our analysis capabilities. WDYT @reubenwong97?

Hi @woodruffw , I added a new finding to the `obfuscation` audit as well, but I wasn't confident about the persona for the finding. Could you help to take a look?

For the rest of the changes, I'll update the release notes at the end once they are more or less confirmed, thank you!

crates/zizmor/tests/integration/test-data/template-injection/dynamic-with-clause.yml
reubenwong97, line 1: For the naming for the file, this one contains actions not actually used by the original issue, so I had a different name for it compared to the other test cases..

crates/zizmor/src/audit/artipacked.rs
```diff
@@ -107,6 +111,33 @@ impl Artipacked {
...
+                            .add_location(
+                                step.location()
+                                    .with_keys(["uses".into()])
+                                    .subfeature(Subfeature::new(0, uses.raw()))
+                                    .annotated("this action"),
```
woodruffw:
```suggestion
                                    .annotated("this checkout"),
```

crates/zizmor/src/audit/artipacked.rs
woodruffw: Looks right to me, thanks.

crates/zizmor/src/audit/obfuscation.rs
```diff
@@ -220,6 +220,28 @@ impl Obfuscation {
...
+                        .add_location(
+                            step.location()
+                                .with_keys(["uses".into()])
+                                .annotated("this action"),
+                        )
+                        .add_location(
+                            step.location()
+                                .primary()
+                                .with_keys(["with".into()])
+                                .annotated("expression in with: clause cannot be analyzed"),
```
woodruffw:
```suggestion
                                .annotated("use of an expression for `with:` prevents analysis"),
```

crates/zizmor/src/audit/template_injection.rs
woodruffw: Hmm, I guess this will probably just be too vague/not actionable (and we'll cover it via `obfuscation`). Thanks for trying it out though!

crates/zizmor/src/audit/template_injection.rs
```diff
@@ -573,6 +573,33 @@ impl TemplateInjection {
...
+                    .add_location(
+                        step.location()
+                            .primary()
+                            .with_keys(["with".into()])
+                            .annotated("dynamic with: clause cannot be analyzed"),
+                    )
+                    .build(step)
+                    .map_err(Self::err)?,
+            );
+        }
```
woodruffw:
```suggestion
```

docs/release-notes.md
```diff
@@ -41,12 +31,10 @@ of `zizmor`.
 * Fixed a bug where the [secrets-outside-env] audit would incorrectly
   flag reusable workflows (#1777)
 
-* Fixed a bug where expressions containing `Infinity` or `NaN` would fail to parse (#1778)
-
-* Fixed a bug where some parenthetical forms in expressions would fail to parse (#1779)
-
 ### Changes ⚠️
 
+* Fixed a bug where dynamic `with: ${{ expr }}` clauses would cause a crash (#1772)
```
woodruffw:
```suggestion
* Fixed a bug where `with: ${{ expr }}` clauses would cause a crash (#1772)
```

reubenwong97:
Hi @woodruffw ,

Updated based on your comments! Can you help to take a look again? Thanks!

PR #1820, opened 2026-03-28
feat(unpinned-tools): new audit for actions using underlying tools that are unpinned

### LINKED ISSUE #1775: New audit: unpinned-tools
I've been thinking about this one recently in the context of trivy's [ongoing issues](https://github.com/aquasecurity/trivy/discussions/10425): it's good to hash-pin actions, but a lot of actions internally just fetch whatever "latest" version of a binary is available as a GitHub release. Consequently, even hash-pinned actions can be a vector for distributing malicious builds, since "latest" is always a mutable concept (and without immutable releases, previous tags may also be mutable).

So, we should have an audit for this. Like with `cache-poisoning`, we'll probably want to use the coordinate API and start with a fixed set of well-known actions that are known to install binaries. Some good initial candidates:

- `aquasecurity/trivy-action`
- `taiki-e/install-action`

Each of these will require slightly different coordinates/nuances. For example `trivy-action` has the `version:` input, but `version: latest` should still be flagged. In contrast `install-action` has no `version:` input, but the `tool:` input should have a `@version` suffix.

Those are the relatively easy cases; it might also be nice to be able to flag actions like `astral-sh/setup-uv`, but the wrinkle with those is that they also support a `version-file` input and by policy zizmor doesn't attempt to parse files outside of workflows/actions/etc. for supporting information. So `setup-uv` and similar would be subject to higher false positive rates.

There are also actions that never make sense to flag, like our own `zizmor-action`: `zizmor-action` has a `version:` input that defaults to `latest`, but under the hood `latest` is actually hash-pinned within the repo itself and is therefore immutable and changes only when the action itself changes.

By way of example, here would be some steps that an audit here should flag:

```yaml
- uses: aquasecurity/trivy-action@whatever # FLAG: no `version:` input
- uses: taiki-e/install-action@whatever
  with:
    tool: cargo-hack  # FLAG: no `@version` prefix 
- uses: taiki-e/install-action@whatever
  with:
    tool: cargo-hack@0.5,cargo-minimal-versions # FLAG: we should handle multiple tools correctly here (whitespace-separated is also allowed)
```

## Open questions

- Are there other actions we should flag in the initial version of this audit? The two above seem sufficient for a first stab to me, but there might be other widely used ones.
- Is `unpinned-binary-actions` a good name for this? Maybe `unpinned-tool-actions` or `unpinned-tools`? I kind of like `unpinned-tools`.

CC @reubenwong97, this might be a good one to try and implement if you're wanting to take a stab at a full-sized, nontrivial audit 🙂 -- the biggest open question here is whether we need to make some changes to the coordinate API to support this, I think.

reubenwong97:
## Pre-submission checks

Please check these boxes:

- [x] **Mandatory**: This PR corresponds to an issue: #1775 .

      use that violates the policy.

## Summary

Creates a new audit, `unpinned-tools` that checks `aquasecurity/trivy-action` and `1password/load-secrets-action` usage for pinning of underlying tools.

## Test Plan

Integration test was created.

Closes #1775

### DIFF
--- a/crates/zizmor/src/config/schema.rs
+++ b/crates/zizmor/src/config/schema.rs
@@ -124,7 +124,8 @@ define_audit_rules! {
     archived_uses,
     misfeature,
     superfluous_actions,
-    github_app;
+    github_app,
+    unpinned_tools;
 
     [DependabotCooldownRuleConfig] dependabot_cooldown,
     [ForbiddenUsesRuleConfig] forbidden_uses,
--- a/support/zizmor.schema.json
+++ b/support/zizmor.schema.json
@@ -278,6 +278,9 @@
         "unpinned-images": {
           "$ref": "#/definitions/BaseRuleConfig"
         },
+        "unpinned-tools": {
+          "$ref": "#/definitions/BaseRuleConfig"
+        },
         "unpinned-uses": {
           "$ref": "#/definitions/UnpinnedUsesRuleConfig"
         },


reubenwong97:
Hi @woodruffw ,

This is my first stab at the `unpinned-tools` audit, can you help take a look and let me know what you think? Thanks!

crates/zizmor/src/audit/unpinned_tools.rs
```diff
@@ -0,0 +1,115 @@
...
+
+#[allow(clippy::unwrap_used)]
+static KNOWN_UNPINNED_TOOLS_ACTIONS: LazyLock<Vec<ActionCoordinate>> = LazyLock::new(|| {
+    vec![
+        // https://github.com/aquasecurity/trivy-action/blob/master/action.yaml
+        ActionCoordinate::NotConfigurable("aquasecurity/trivy-action".parse().unwrap()),
+        // https://github.com/1Password/load-secrets-action/blob/main/action.yml
+        ActionCoordinate::NotConfigurable("1password/load-secrets-action".parse().unwrap()),
+    ]
+});
```
reubenwong97: It may be that I don't know how to use the `ActionCoordinate` API well, but in the end, because I wanted to check for `with:` -> `version:` not being present or `with:` -> `version: latest`, I found it easier to do the matching inside the step processing itself..

woodruffw: Hmm, yeah, this unfortunately makes the value of the coordinate API here pretty limited 😅 -- that's not your fault, just me under thinking how useful it would be.

(You can encode those kinds of checks with expressions, but there's no clean way to extract _which_ expression fork satisfied at the moment.)

woodruffw:
Thanks @reubenwong97! This looks like a good start to me; I'll take a closer look at it tonight most likely.

crates/zizmor/src/audit/unpinned_tools.rs
reubenwong97: Hi @woodruffw ,

Since the coordinate API is no longer useful for the current iteration of the audit, I've cleaned up the audit to remove its use. Can you help to take a look when you have time? Thanks!

crates/zizmor/src/audit/unpinned_tools.rs
```diff
@@ -0,0 +1,109 @@
...
+                            .annotated("this action's tool version is not pinned"),
+                    ),
+            ),
+            Some(EnvValue::String(v)) if v == "latest" => Some(
+                Self::finding()
+                    .confidence(Confidence::High)
+                    .severity(Severity::Medium)
+                    .add_location(
+                        step.location()
+                            .with_keys(["uses".into()])
```
woodruffw, line 64: Nit: you can use a subfeature here to make the location even more precise, something like this:


```suggestion
                            .with_keys(["uses".into()])
                            .subfeature(Subfeature::new(0, uses.raw()))
```

crates/zizmor/src/audit/unpinned_tools.rs
```diff
@@ -0,0 +1,109 @@
...
+        else {
+            return Ok(findings);
+        };
+
+        if !KNOWN_UNPINNED_TOOLS_ACTIONS
+            .iter()
+            .any(|action| uses.matches(action))
+        {
+            return Ok(findings);
+        }
```
woodruffw, line 44: Just thinking out loud: this approach is fine for now since all of the current actions have the same "shape" (they each take a `version:` input), but in the long term we'll probably need to support other shapes here. 

That's not something we need to handle for an MVP of this audit, though.

crates/zizmor/src/audit/unpinned_tools.rs
```diff
@@ -0,0 +1,109 @@
...
+                    .confidence(Confidence::High)
+                    .severity(Severity::Medium)
+                    .add_location(
+                        step.location()
+                            .primary()
+                            .with_keys(["uses".into()])
+                            .annotated("this action's tool version is not pinned"),
+                    ),
+            ),
+            Some(EnvValue::String(v)) if v == "latest" => Some(
```
woodruffw, line 58: Nit: I think we also need to handle the case where we have `version: ${{ expr }}`, and emit a low-confidence finding in that case.

Something like `Some(EnvValue::String(v)) if ExtractedExpr::from_fenced(v).is_some()` should work, I think.

woodruffw: (I'm not 100% sure about that though -- maybe that could also just be a TODO comment for the MVP.)

crates/zizmor/src/audit/unpinned_tools.rs
reubenwong97: Thanks for the tip!

crates/zizmor/src/audit/unpinned_tools.rs
reubenwong97: Thanks for the idea! The tests have been updated and this seems reasonable to me.

crates/zizmor/src/audit/unpinned_tools.rs
reubenwong97: Yeap agreed on this. This audit would have looked a lot more different if `taiki-e/install-action` were part of this audit... But I think I'd need more concrete examples before I can design around the final shape of the problem. For now, I think something like this keeps things simple!

reubenwong97:
Hi @woodruffw ,

Really appreciate you taking the time to take another look at this PR! I've updated it based on your comments, could you help to check again? For the audit documentation, I've created an example at https://github.com/woodruffw/gha-hazmat/pull/6, where I've noticed most of the examples are. I can update the documentation with the exact link once that is brought in.

Thank you!

woodruffw:
No problem @reubenwong97, thanks for pushing this through and apologies for my slower review on this one. This LGTM overall, I'm going to push some small tweaks to your branch and then merge.

woodruffw:
Thank you @reubenwong97!

Issue #1818, opened 2026-03-27
Feature: Organize arguments in `--help` under topical headings


epage:
### Pre-submission checks

- [x] I am **not** reporting a bug (crash, false positive/negative, etc). These must be filed via the bug report template.
- [x] I have looked through both the [open](https://github.com/zizmorcore/zizmor/issues?q=is%3Aissue+is%3Aopen+) and [closed](https://github.com/zizmorcore/zizmor/issues?q=is%3Aissue%20state%3Aclosed) issues for a duplicate request.

### What's the problem this feature will solve?

In first using `zizmor`, I was confused by the help output.  I was looking for a feature and didn't see an option for it because it was later in the output.

### Describe the solution you'd like

```
Static analysis for GitHub Actions

Usage: zizmor [OPTIONS] <INPUT>...

Arguments:
  <INPUT>...  The inputs to audit

Lint:
      --fix[=<MODE>]
          Fix findings automatically, when available (EXPERIMENTAL) [possible values: safe, unsafe-only, all]
  -p, --pedantic
          Emit 'pedantic' findings
      --persona <PERSONA>
          The persona to use while auditing [default: regular] [possible values: auditor, pedantic, regular]
      --min-severity <MIN_SEVERITY>
          Filter all results below this severity [possible values: informational, low, medium, high]
      --min-confidence <MIN_CONFIDENCE>
          Filter all results below this confidence [possible values: low, medium, high]
      --collect <COLLECT>...
          Control which kinds of inputs are collected for auditing [default: default] [possible values: all, default,
workflows, actions, dependabot]
      --strict-collection
          Fail instead of warning on syntax and schema errors in collected inputs

Output:
  -v, --verbose...
          Increase logging verbosity
  -q, --quiet...
          Decrease logging verbosity
      --no-progress
          Don't show progress bars, even if the terminal supports them
      --format <NAME>
          The output format to emit. By default, cargo-style diagnostics will be emitted [default: plain] [possible va
lues: plain, json, json-v1, sarif, github]
      --color <WHEN>
          Control the use of color in output [possible values: auto, always, never]
      --render-links <WHEN>
          Whether to render OSC 8 links in the output [env: ZIZMOR_RENDER_LINKS=] [default: auto] [possible values: au
to, always, never]
      --show-audit-urls <WHEN>
          Whether to render audit URLs in the output, separately from any URLs embedded in OSC 8 links [env: ZIZMOR_SH
[...]

PR #1831, opened 2026-03-30
feat: group help options by category

deckstose:
## Pre-submission checks

Please check these boxes:

      one first).

      use that violates the policy.

## Summary

Make `--help` more readable by grouping related options together.

New output looks like this:

```
Static analysis for GitHub Actions

Usage: zizmor [OPTIONS] <INPUTS>...

Arguments:
  <INPUTS>...  The inputs to audit

Lint Options:
      --fix[=<MODE>]
          Fix findings automatically, when available (EXPERIMENTAL) [possible values: safe, unsafe-only, all]
  -p, --pedantic
          Emit 'pedantic' findings
      --persona <PERSONA>
          The persona to use while auditing [default: regular] [possible values: auditor, pedantic, regular]
      --min-severity <MIN_SEVERITY>
          Filter all results below this severity [possible values: informational, low, medium, high]
      --min-confidence <MIN_CONFIDENCE>
          Filter all results below this confidence [possible values: low, medium, high]
      --collect <COLLECT>...
          Control which kinds of inputs are collected for auditing [default: default] [possible values: all, default, workflows, actions, dependabot]

Reporting Options:
  -v, --verbose...
          Increase logging verbosity
  -q, --quiet...
          Decrease logging verbosity
      --no-progress
          Don't show progress bars, even if the terminal supports them
      --format <FORMAT>
          The output format to emit. By default, cargo-style diagnostics will be emitted [default: plain] [possible values: plain, json, json-v1, sarif, github]
      --render-links <RENDER_LINKS>
          Whether to render OSC 8 links in the output [env: ZIZMOR_RENDER_LINKS=] [default: auto] [possible values: auto, always, never]
      --show-audit-urls <SHOW_AUDIT_URLS>
          Whether to render audit URLs in the output, separately from any URLs embedded in OSC 8 links [env: ZIZMOR_SHOW_AUDIT_URLS=] [default: auto] [possible values: auto, always, never]
      --no-exit-codes
          Disable all error codes besides success and tool failure
      --strict-collection
          Fail instead of warning on syntax and schema errors in collected inputs

Network Options:
  -o, --offline                    Perform only offline operations [env: ZIZMOR_OFFLINE=]
      --gh-token <GH_TOKEN>        The GitHub API token to use [env: GH_TOKEN or GITHUB_TOKEN or ZIZMOR_GITHUB_TOKEN]
      --gh-hostname <GH_HOSTNAME>  The GitHub Server Hostname. Defaults to github.com [env: GH_HOST=] [default: github.com]
      --no-online-audits           Perform only offline audits [env: ZIZMOR_NO_ONLINE_AUDITS=]
      --cache-dir <DIR>            The directory to use for HTTP caching. By default, a host-appropriate user-caching directory will be used

Options:
      --lsp                  Run in language server mode (EXPERIMENTAL)
  -c, --config <FILE>        The configuration file to load. This loads a single configuration file across all input groups, which may not be what you intend [env: ZIZMOR_CONFIG=]
      --no-config            Disable all configuration loading
      --completions <SHELL>  Generate tab completion scripts for the specified shell [possible values: bash, elvish, fish, nushell, powershell, zsh]
      --color <WHEN>         Control the use of color in output [possible values: auto, always, never]
      --thanks               Emit thank-you messages for zizmor's sponsors
  -h, --help                 Print help (see a more with '--help')
  -V, --version              Print version
```

And here is a screenshot:

Screenshot_20260330_172733

---
Closes: https://github.com/zizmorcore/zizmor/issues/1818

### DIFF
--- a/docs/release-notes.md
+++ b/docs/release-notes.md
@@ -36,6 +36,10 @@ of `zizmor`.
 * `obfuscation` audit emits a finding that `with: ${{ expr }}` clauses cannot
   be analyzed (#1772)
 
+* `zizmor --help` is now rendered with option groups for improved readability (#1831)
+
+    Many thanks to @deckstose for implementing this improvement!
+
 ### Bug Fixes 🐛
 
 * Fixed a bug where the [concurrency-limits] audit reported findings


crates/zizmor/src/main.rs
```diff
@@ -131,137 +246,73 @@ struct App {
...
-    #[arg(long, value_enum, default_value_t, env = "ZIZMOR_SHOW_AUDIT_URLS")]
-    show_audit_urls: CliShowAuditUrls,
+    /// The directory to use for HTTP caching. By default, a
+    /// host-appropriate user-caching directory will be used.
+    #[arg(
+        long,
+        value_name = "DIR",
+        default_value_t = App::default_cache_dir(),
+        hide_default_value = true,
+        value_hint = ValueHint::DirPath
```
deckstose, line 256: Both here and below for `config` I've added the value name and value hint.

crates/zizmor/src/main.rs
```diff
@@ -100,6 +152,69 @@ struct App {
...
+    ///
+    /// By default, all workflows and composite actions are collected,
+    /// while honoring `.gitignore` files.
+    #[arg(long, default_values = ["default"], num_args=1.., value_delimiter=',')]
+    collect: Vec<CliCollectionMode>,
+}
+
+#[derive(Debug, Args)]
+#[command(next_help_heading = "Reporting Options")]
+struct ReportingArgs {
```
deckstose: I've thought about `Report` and `Reporter` here as well but I'm on the fence what fits better.

crates/zizmor/src/main.rs
```diff
@@ -85,10 +85,62 @@ const STYLES: Styles = Styles::styled()
...
+    /// for a GitHub repository. In the latter case, a `@ref` can be appended
+    /// to audit the repository at a particular git reference state.
+    ///
+    /// Use `-` to read a single input from stdin.
+    #[arg(required = true)]
+    inputs: Vec<String>,
+
     #[command(flatten)]
-    lsp: LspArgs,
+    lint: LintArgs,
```
woodruffw: Nit: Let's call this `AuditArgs` maybe? The "lint" term isn't used elsewhere in the codebase.

crates/zizmor/src/main.rs
```diff
@@ -85,10 +85,62 @@ const STYLES: Styles = Styles::styled()
...
+        })
+        .expect("failed to determine default cache directory")
+        .cache_dir()
+        .try_into()
+        .expect("failed to turn cache directory into a sane path")
+    }
+}
+
+#[derive(Debug, Args)]
+#[command(next_help_heading = "Lint Options")]
```
woodruffw: Per above:

```suggestion
#[command(next_help_heading = "Audit Options")]
```

(and elsewhere below)

crates/zizmor/src/main.rs
woodruffw: Hmm, I'd actually go with `OutputArgs` here -- we don't use the "report" term anywhere else that I can tell, whereas the docs do cover most of these options under "output."

crates/zizmor/src/main.rs
woodruffw: Thank you!

crates/zizmor/src/main.rs
deckstose: That's fine for me. I grouped the "Output" options together with `no_exit_codes`, and in my head that didn't sound right for "output" as in what's displayed on the terminal. Should I move the no_exit_codes flag elsewhere or keep it with output?

woodruffw: Yeah, that's slightly misaligned but I think it's okay to group it there for now. We can always move it if someone complains 🙂

crates/zizmor/src/main.rs
```diff
@@ -100,6 +152,69 @@ struct App {
...
+    /// Only affects `--format=plain` (the default).
+    #[arg(long, value_enum, default_value_t, env = "ZIZMOR_RENDER_LINKS")]
+    render_links: CliRenderLinks,
+
+    /// Whether to render audit URLs in the output, separately from any URLs
+    /// embedded in OSC 8 links.
+    ///
+    /// Only affects `--format=plain` (the default).
+    #[arg(long, value_enum, default_value_t, env = "ZIZMOR_SHOW_AUDIT_URLS")]
+    show_audit_urls: CliShowAuditUrls,
```
epage, line 199: There a reason these didn't get `value_name = "WHEN"`?

crates/zizmor/src/main.rs
```diff
@@ -100,6 +152,69 @@ struct App {
...
+
+    /// Disable all error codes besides success and tool failure.
+    #[arg(long)]
+    no_exit_codes: bool,
+
+    /// Fail instead of warning on syntax and schema errors
+    /// in collected inputs.
+    #[arg(long)]
+    strict_collection: bool,
```
epage, line 208: Is this better next to `collect`?

crates/zizmor/src/main.rs
deckstose: I missed that :fearful:

crates/zizmor/src/main.rs
deckstose: I don't mind it either way. I think it make sense here as it influences the return value and it's grouped with `no_exit_codes`, see https://github.com/zizmorcore/zizmor/pull/1831#discussion_r3010691342.

PR #1834, opened 2026-03-30
fix: Tweak help grouping

epage:
## Pre-submission checks

Please check these boxes:

      one first).

      use that violates the policy.

## Summary

This builds on #1831 with some ideas missed from #1818 as well as
- Trying to provide more `value_name`s
- Realizing "collect" likely best goes with `INPUT`


crates/zizmor/src/main.rs
```diff
@@ -86,17 +86,10 @@ const STYLES: Styles = Styles::styled()
 #[derive(Parser)]
 #[command(about, version, styles = STYLES)]
 #[command(disable_help_flag = true, disable_version_flag = true)]
+#[command(next_display_order = 1)]
```
epage, line 89: Apparently, clap doesn't set display order on positionals, so we have to leave open a space for `INPUT` and then use it

PR #1839, opened 2026-03-31
refactor: Leverage human-panic for CI-specialized logic

### LINKED ISSUE #1838: Use human-panic's new CI-specialization, rather than hand implement
zizmor does
https://github.com/zizmorcore/zizmor/blob/3df1761d07c173903bcee28fc3b5e6ed2fb53ab5/crates/zizmor/src/main.rs#L1052-L1070

I mirrored the idea of this to `human-panic` in rust-cli/human-panic#207.

Wanted to check in first in case there was any further refinement of the output wanted before switching over to exclusively using human-panic.

epage:
## Pre-submission checks

Please check these boxes:

      one first).

      use that violates the policy.

## Summary

I mirrored the idea of what zizmor does in rust-cli/human-panic#207, making it so that this check wouldn't be needed anymore.

Fixes #1838

## Test Plan

Output is tested in `human-panic`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -69,27 +69,12 @@ dependencies = [
 
 [[package]]
 name = "anstream"
-version = "0.6.21"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "43d5b281e737544384e969a5ccad3f1cdd24b48086a0fc1b2a5262a26b8f4f4a"
-dependencies = [
- "anstyle",
- "anstyle-parse 0.2.7",
- "anstyle-query",
- "anstyle-wincon",
- "colorchoice",
- "is_terminal_polyfill",
- "utf8parse",
-]
-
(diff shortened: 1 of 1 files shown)


woodruffw (approved):
Thanks @epage! I tested this locally with a contrived panic and it looks reasonable to me.

Q about human-panic: I noticed it uses exact version constraints (`x.y.z`) for most of its deps, which means that we end up with some duplicated dependencies here. Would you be open to relaxing those to `x.y` so that more of the deps could be unified in downstreams?

epage:
human-panic does not use exact version requirements (`=`) but the default operator which is an implicit `^` operator which is the "semver" operator.

See https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#default-requirements

I specify all version fields combined and verifying the lower bounds are valid (though not if I can go lower).

PR #1849, opened 2026-04-02
feat(ref-version-mismatch): detect missing version comments on SHA-pinned actions

### LINKED ISSUE #1836: Feature: Detect missing version comments on SHA-pinned actions
### Pre-submission checks

When an action is pinned to a commit SHA without a version comment, it hurts readability and may play poorly with tooling (e.g., dependency update tools).

```yaml
# What version is this? No way to tell without looking up the SHA and wasting 30s of my life :p
uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd

# Clear and tool-friendly :D
uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
```

Zizmor today already has some handling for version comments, e.g., validating the accuracy for existing comments with [`ref-version-mismatch`](https://docs.zizmor.sh/audits/#ref-version-mismatch), or generating them for [`unpinned-uses`](https://docs.zizmor.sh/audits/#unpinned-uses) auto-fixes. It seems sensible to add functionality to mandate that a  versioncomment exists at all.

Detect SHA-pinned `uses:` that lack a version comment. Similar to other documentation audits such as [undocumented-permissions](https://docs.zizmor.sh/audits/#undocumented-permissions), this should likely be pedantic only, since it alone does not constitute a security vulnerability.

The auto-fix would use `longest_tag_for_commit` to resolve the SHA to a tag and add the comment via `EmplaceComment`. If no tag exists for the commit, zizmor should skip the finding (as we can't add a comment for something with no tag).

This could be added as a new pedantic finding extending the existing `ref-version-mismatch`,  or as a new, standalone audit (e.g., `missing-version-comment`)

### Additional context

Happy to create a PR if there's support :)

shaanmajid:
## Pre-submission checks

Please check these boxes:

      one first).

- [X] Having read the [AI policy], I hereby disclose the use of an LLM or other
      use that violates the policy.

## Summary

Closes #1836

Extends `ref-version-mismatch` to detect SHA-pinned `uses:` entries that lack a version comment, reported with an autofix that adds `# <tag>` via `longest_tag_for_commit`.

Notes:
- SHA-pinned actions with no tag produce no findings, as there's no version to suggest.
- When a non-version comment already exists, a finding is emitted but no autofix is attached to avoid overwriting user intent.
- Open to changing this, but started this finding as pedantic rather than default, consistent with other documentation-only findings like `undocumented-permissions`. A missing comment is primarily a readability/hygiene issue -- nobody is being misled. By contrast, the existing mismatch finding is more severe, as a wrong comment actively misleads humans and tooling about which version is pinned.

## Test Plan

- Offline unit tests for `CommentVersionState` classification and `EmplaceComment` fix generation
- `gh-token-tests` covering: missing comment (happy path), CRLF, bizarre YAML formatting, untagged SHA (no finding), non-version comment (finding without fix)

### DIFF
--- a/docs/release-notes.md
+++ b/docs/release-notes.md
@@ -54,6 +54,11 @@ of `zizmor`.
 
     Many thanks to @miketheman for implementing this improvement!
 
+* The [ref-version-mismatch] audit now detects missing version comments as well
+  (#1849)
+
+    Many thanks to @shaanmajid for proposing and implementing this improvement!
+
 ### Bug Fixes 🐛
 
 * Fixed a bug where the [concurrency-limits] audit reported findings


crates/zizmor/src/audit/ref_version_mismatch.rs
```diff
@@ -96,10 +123,47 @@ impl RefVersionMismatch {
...
+                let (annotation, tip) = match comment_version_state {
+                    CommentVersionState::Missing => (
+                        "missing version comment",
+                        format!("add version comment '# {}'", tag.name),
+                    ),
+                    CommentVersionState::NonVersionComments => (
+                        "comment does not contain a version",
+                        format!("rewrite comment to include '# {}'", tag.name),
+                    ),
+                    CommentVersionState::Version(_) => unreachable!(),
```
woodruffw, line 150: Nit: I don't love new unreachables in the codebase but I don't have a better idea for avoiding this one (without a much larger refactor of this audit).

PR #1691, opened 2026-03-04
Support languageservices expressions test data

DarkaMaul:
## Pre-submission checks

Please check these boxes:

- [x] **Mandatory**: This PR corresponds to an issue (if not, please create
      one first) (#1688)

- [X] I hereby disclose the use of an LLM or other AI coding assistant in the
      creation of this PR. PRs will not be rejected for using AI tools, but
      *will* be rejected for undisclosed use.

If a checkbox is not applicable, you can leave it unchecked.

## Summary

This PR adds a mechanism to support the KAT from [actions/languageservices](https://github.com/actions/languageservices/tree/main/expressions/testdata) .

## Test Plan

This is a test only PR.

⚠️ This PR should be merged as if because it contains known failure! ⚠️

PR #1857, opened 2026-04-05
KATs for GitHub Actions expressions

woodruffw:
This is a living copy of #1691.

### DIFF
--- a/crates/github-actions-expressions/tests/languageservices_kat.rs
+++ b/crates/github-actions-expressions/tests/languageservices_kat.rs
@@ -4,8 +4,9 @@
 //! See `support/sync-expression-tests.py` for how the KATs themselves
 //! are synchronized from the upstream test suite.
 
-use std::collections::HashMap;
+use std::collections::{HashMap, HashSet};
 use std::path::Path;
+use std::sync::LazyLock;
 
 use github_actions_expressions::{Evaluation, Expr};
 use serde::Deserialize;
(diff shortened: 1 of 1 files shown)


crates/github-actions-expressions/tests/languageservices_kat.rs
```diff
@@ -0,0 +1,224 @@
...
+    }
+}
+
+fn normalize_infinity(s: &str) -> &str {
+    match s {
+        "inf" | "Infinity" => "Infinity",
+        "-inf" | "-Infinity" => "-Infinity",
+        other => other,
+    }
+}
```
woodruffw: Flagging for myself to look at later: I think we want to use `EvaluationSema` here rather than open-coding equality.

crates/github-actions-expressions/tests/languageservices_kat.rs
```diff
@@ -0,0 +1,224 @@
...
+
+/// Group name prefixes to skip
+const SKIP_GROUP_PREFIXES: &[&str] = &[
+    // Depth/memory limit tests: our parser doesn't enforce these limits
+    // and will stack-overflow on deeply nested expressions.
+    "depth-errors",
+    "memory-errors",
+    // case() is not a standard GitHub Actions expression function.
+    "case",
```
woodruffw: Flagging for myself: we support this now, should remove the skip.

PR #1755, opened 2026-03-12
chore(ci): don't run benchmarks on forks

miketheman:
## Pre-submission checks

Please check these boxes:

- [ ] **Mandatory**: This PR corresponds to an issue (if not, please create
      one first).

- [x] Having read the [AI policy], I hereby disclose the use of an LLM or other
      AI coding assistant in the creation of this PR. PRs will not be rejected
      for using AI tools, but *will* be rejected for undisclosed use or
      use that violates the policy.

[AI policy]: https://github.com/zizmorcore/.github/blob/main/AI_POLICY.md

If a checkbox is not applicable, you can leave it unchecked.

## Summary

Prevents failures from forks trying to unsuccessfully upload benchmarks to CodSpeed

PR #1862, opened 2026-04-06
fix(ci): skip benchmarks on fork repos for all event types

### LINKED ISSUE #1861: ci: CodSpeed Benchmark workflow fails for fork repos on pushes to main
(Apologies for ignoring the issue templates, but none of the categories cleanly apply :< )

The CodSpeed benchmark workflow [runs on push to main, labeled PRs, and workflow_dispatch](https://github.com/zizmorcore/zizmor/blob/d9a98ea45ef6bdfe7cf4632c1104bdb7b178cd8e/.github/workflows/benchmark.yml#L3-L16). When a fork syncs its main branch with upstream, this triggers the `push` event, which runs the benchmark job and fails with a 401 from CodSpeed. 

This is the same general issue #1755 attempted to fix by guarding with [`!github.event.pull_request.head.repo.fork`](https://github.com/zizmorcore/zizmor/blob/d9a98ea45ef6bdfe7cf4632c1104bdb7b178cd8e/.github/workflows/benchmark.yml#L30). However, this fails to successfully skip on fork pushes to `main`; that field is `null` (not `false`) on `push` events since there is no pull request context, so `!null` evaluates to `true` and the guard is bypassed.

See [this](https://github.com/shaanmajid/zizmor/actions/runs/24013818939/job/70029637195) job run on my fork for an example failure.

shaanmajid:
## Pre-submission checks

Please check these boxes:

      one first).

      use that violates the policy.

## Summary

Fixes #1861. Follow-up to #1755.

Skips CodSpeed benchmark workflow runs on fork repos for all event types. This should now avoid workflow failures when syncing a fork's main branch (e.g., https://github.com/shaanmajid/zizmor/actions/runs/24013818939/job/70029637195).

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -27,7 +27,7 @@ jobs:
     # PRs only get benchmarked if they have the `run-benchmarks` label.
     # Forks don't have access to CodSpeed secrets, so skip them.
     if: |
-      !github.event.pull_request.head.repo.fork
+      github.repository_owner == 'zizmorcore'
         && (contains(github.event.pull_request.labels.*.name, 'run-benchmarks')
           || github.event_name == 'push'
           || github.event_name == 'workflow_dispatch')


miketheman (approved):
Fair enough. I was trying to avoid passing in the repo name, since this project had already transferred once, who's to say it won't in the future. 😉

PR #1871, opened 2026-04-07
fix(ref-version-mismatch): handle prerelease tags in version comment parsing

### LINKED ISSUE #1869: [BUG]: `ref-version-mismatch` causes persistent false positives for prerelease tags
### Pre-submission checks

### zizmor version

1.23.1

### Expected behavior

If a SHA-pinned `uses:` comment correctly references a prerelease tag (such as `v6-beta`), `ref-version-mismatch` should not report a finding.

(Also, any autofixable finding should converge and be resolved upon being ran with `--fix=all`; it should never enter into an autofix [death spiral](https://en.wikipedia.org/wiki/Ant_mill). In this case, it should not keep reporting a fixable mismatch when the comment text already matches the suggested tag.)

### Actual behavior

`ref-version-mismatch` reports a mismatch for prerelease tags, even when the comment matches and the finding itself indicates such.

`--fix=all` does not resolve the finding; subsequent runs still report the same mismatch and mark it fixable.

### Reproduction steps

1. Save this file as `repro.yml`:

```yaml
name: Beta Tag Repro
on: push
permissions: {}
jobs:
  test:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@71cf2267d89c5cb81562390fa70a37fa40b1305e # v6-beta
        with:
          persist-credentials: false
```

2. Run:

```bash
GH_TOKEN="$(gh auth token)" zizmor path/to/repro.yml
```

3. Run fix mode:

```bash
GH_TOKEN="$(gh auth token)" zizmor --fix=all path/to/repro.yml
```

4. Re-run step 2 or 3 :p

5. Note that comment remains `# v6-beta` and mismatch still appears. :-(

### Logs

```plain text
warning[ref-version-mismatch]: detects commit SHAs that don't match their version comment tags
 --> .../repro.yml:8:75
  |
8 |       - uses: actions/checkout@71cf2267d89c5cb81562390fa70a37fa40b1305e # v6-beta
  |         ---------------------------------------------------------------   ^^ points to commit de0fac2e4500
  |         |
  |         is pointed to by tag v6-beta
  |
  = note: audit confidence -> High
```

### Additional context

Root cause is partial regex matching in `VERSION_COMMENT_PATTERNS` for `ref-version-mismatch`.

Current patterns only capture numeric semver-like prefixes (for example `v\d+(?:\.\d+)*`) and are not bounded to the full token. As a result, comments like `# v6-beta` are parsed as `v6` instead of `v6-beta`.

shaanmajid:
## Pre-submission checks

Please check these boxes:

      one first).

- [X] Having read the [AI policy], I hereby disclose the use of an LLM or other
      use that violates the policy.

## Summary

Fixes #1869 

Modifies the `ref-version-mismatch` comment parser to capture full tag names including prerelease suffixes.

Previously, `VERSION_COMMENT_PATTERNS` used `v\d+(?:\.\d+)*`. This would stop at the last numeric segment, so e.g., `# v6-beta` was parsed as `v6`, causing a perpetual false positive (since the "fix" would correctly update to the same `# v6-beta` comment).

This PR implements two behavioral changes to the regex patterns:

1.` tag=` form: replaces the version-specific regex with `(\S+)`. Since `tag=` is an explicit signal that the next token is a tag name, capturing up to whitespace is appropriate.
2. `# v...` and `version:`/`ver=` forms: added an optional prerelease group `(?:-[\w.-]+)?` to capture suffixes like `-beta`, `-rc.1`, and compound forms like `-beta-2`.

## Test Plan

- Unit tests for prerelease version comment parsing across several pattern forms
- Added integration test with MRE from #1869
- Manually verified autofix behavior end-to-end (`--fix=all` -> re-run -> no findings)

### DIFF
--- a/crates/zizmor/src/audit/ref_version_mismatch.rs
+++ b/crates/zizmor/src/audit/ref_version_mismatch.rs
@@ -30,14 +30,12 @@ audit_meta!(
 #[allow(clippy::unwrap_used)]
 static VERSION_COMMENT_PATTERNS: LazyLock<Vec<Regex>> = LazyLock::new(|| {
     vec![
-        // Matches "# tag=v2.8.0" or "# tag=v1.2.3"
-        Regex::new(r"#\s*tag\s*=\s*(v\d+(?:\.\d+)*(?:\.\d+)?)").unwrap(),
-        // Matches "# v2.8.0"
-        Regex::new(r"#\s*(v\d+(?:\.\d+)*(?:\.\d+)?)").unwrap(),
-        // Matches version without 'v' prefix: "# tag=2.8.0"
-        Regex::new(r"#\s*tag\s*=\s*(\d+(?:\.\d+)*(?:\.\d+)?)").unwrap(),
+        // Matches "# tag=v2.8.0", "# tag=v6-beta", or any non-whitespace tag token.
+        Regex::new(r"#\s*tag\s*=\s*(\S+)").unwrap(),
+        // Matches "# v2.8.0" and prerelease forms like "# v1.2.3-rc.1".
+        Regex::new(r"#\s*(v\d+(?:\.\d+)*(?:-[\w.-]+)?)").unwrap(),
         // More flexible: "# version: 2.8.0"
-        Regex::new(r"#\s*(?:version|ver)\s*[:=]\s*(v?\d+(?:\.\d+)*(?:\.\d+)?)").unwrap(),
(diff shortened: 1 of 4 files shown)


woodruffw (approved):
Thanks a ton @shaanmajid!

shaanmajid:
Thank _you_, Mr. Woodruff :^) !

(also sorry for the default pr title lmfao, fixed)

PR #1900, opened 2026-04-13
ref-version-mismatch: handle version comments without v prefix

### LINKED ISSUE #1899: [BUG]: false positive `ref-version-mismatch` fails to recognise simple versions
### Pre-submission checks

### zizmor version

zizmor 1.24.0

### Expected behavior

I expect zizmor to recognise simple github action versions such as 2.4.0 when referenced after an action

### Actual behavior

The version is rejected with a misleading error message

### Reproduction steps

1. Create a custom action in a private repo and tag it with something like 1.2.0
2. reference that action in another repo with the checksum and comment # 1.2.0 after it
3. Run zizmor -p . 
4. The log says that there is no valid version and gives a suggested fix that's wrong

### Logs

```plain text
help[ref-version-mismatch]: action's hash pin has mismatched or missing version comment
  --> ./.github/workflows/docker-build.yml:72:9
   |
72 |         uses: some-org/custom-actions/some-action@some-checksum # 2.4.0
   |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ comment does not contain a version
   |
   = tip: rewrite comment to include '# 2.4.0'
```

### Additional context

There is a mitigation of prefixing the simple version with `version:` for example `version: 1.2.0`. How ever the tip is wrong in the error log.

woodruffw:
Fixes #1899.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1016,7 +1016,7 @@ checksum = "e629b9b98ef3dd8afe6ca2bd0f89306cec16d43d907889945bc5d6687f2f13c7"
 
 [[package]]
 name = "github-actions-expressions"
-version = "1.24.0"
+version = "1.24.1"
 dependencies = [
  "itertools",
  "pest",
@@ -1030,7 +1030,7 @@ dependencies = [
 
 [[package]]
 name = "github-actions-models"
-version = "1.24.0"
+version = "1.24.1"
 dependencies = [
  "indexmap",
  "insta",
@@ -2926,7 +2926,7 @@ dependencies = [
 
 [[package]]
 name = "subfeature"
-version = "1.24.0"
+version = "1.24.1"
 dependencies = [
  "memchr",
  "regex",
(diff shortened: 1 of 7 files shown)


shaanmajid:
Perhaps worth noting that `v?` will capture some non-version comments that happen to start with a digit (e.g., bare dates like `# 2026-04-13`, issue/PR IDs). Not a concern IMO -- those already emit a pedantic finding for the comment not containing a version -- but it does shift that subset from the pedantic persona to the default.

woodruffw:
Hmm, won't the rest of the pattern filter that out? We'll match `# 4` and `# v4` but beyond the `\d+` production we require a `.`, so `# 2026-04-13` wouldn't match for example. But I might be missing something.

shaanmajid:
It does match I believe; tested on [Rustexp](https://rustexp.lpil.uk/):

Screenshot 2026-04-13 at 12 57 04 PM

`(?:\.\d+)*` is zero-or-more, so the decimal section can match empty and `(?:-[\w.-]+)?` then consumes `-04-13`.

Again, I don't think this is actually a concern -- just wanted to note it

Issue #1903, opened 2026-04-14
[BUG]: Crash in `as_pattern` on parenthesized OR expression used as context accessor


DarkaMaul:
### Pre-submission checks

- [x] I am **not** filing a feature request. These should be filed via the feature request form instead.
- [x] I have checked the [Troubleshooting Guide](https://docs.zizmor.sh/troubleshooting) for my problem.
- [x] I have looked through both the [open](https://github.com/zizmorcore/zizmor/issues?q=is%3Aissue+is%3Aopen+) and [closed](https://github.com/zizmorcore/zizmor/issues?q=is%3Aissue%20state%3Aclosed) issues for a duplicate report.

### zizmor version

main

### Expected behavior

`zizmor` should audit a workflow containing `${{ (github.event.pull_request || github.event.issue).number }}` without crashing. This is a valid GitHub Actions expression that selects `.number` from the result of a grouped OR.

### Actual behavior

`zizmor` panics with `unexpected part in context pattern` in `crates/github-actions-expressions/src/context.rs:108` during the `template-injection` audit.

### Reproduction steps

Save this as `repro.yml`:

```yaml
name: reproducer
on: issue_comment

jobs:
  test:
    runs-on: ubuntu-latest
    steps:
      - run: echo "${{ (github.event.pull_request || github.event.issue).number }}"
```

Run `zizmor` on it:

### Logs

```plain text
Stack trace:


 INFO zizmor: 🌈 zizmor v1.24.1
[00:00:00] ░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░ repro.yml
  ↳ ⠁ audit{input=Workflow(file:///tmp/repro.yml)}                       
thread 'main' (80511) panicked at crates/github-actions-expressions/src/context.rs:108:22:
internal error: entered unreachable code: unexpected part in context pattern
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

### Additional context

Found in the wild in [cupy/cupy](https://github.com/cupy/cupy/blob/main/.github/workflows/pr-updated.yml) with a valid GitHub Actions [run](https://github.com/cupy/cupy/actions/runs/24386277191).

DarkaMaul:
Closed by #1904

Issue #1692, opened 2026-03-04
[BUG]: Expression parser fails on parenthesized compound expressions


DarkaMaul:
### Pre-submission checks

- [x] I am **not** filing a feature request. These should be filed via the feature request form instead.
- [x] I have checked the [Troubleshooting Guide](https://docs.zizmor.sh/troubleshooting) for my problem.
- [x] I have looked through both the [open](https://github.com/zizmorcore/zizmor/issues?q=is%3Aissue+is%3Aopen+) and [closed](https://github.com/zizmorcore/zizmor/issues?q=is%3Aissue%20state%3Aclosed) issues for a duplicate report.

### zizmor version

main

### Expected behavior

zizmor should be able to parse valid GitHub Actions expressions where a parenthesized compound expression is used as an operand to a comparison operator such as:

```
2 <= (3 == true)
0 > (0 < 1)
```

### Actual behavior

zizmor emits `couldn't parse expression` warnings for every audit that attempts to parse these expressions

### Reproduction steps

```yaml
name: Minimal reproducer
on: workflow_dispatch
jobs:
  test:
    runs-on: ubuntu-latest
    steps:
      - run: echo "${{ 2 <= (3 == true) }}"
```

### Logs

```plain text
🌈 zizmor v1.23.0-rc5
 WARN audit:audit{input=Workflow(file:///tmp/repro.yml)}: zizmor::audit::template_injection: couldn't parse expression: ${{ 2 <= (3 == true) }}
 WARN audit:audit{input=Workflow(file:///tmp/repro.yml)}: zizmor::audit::overprovisioned_secrets: couldn't parse expression:  2 <= (3 == true)
 WARN audit:audit{input=Workflow(file:///tmp/repro.yml)}: zizmor::audit::unredacted_secrets: couldn't parse expression:  2 <= (3 == true)
 WARN audit:audit{input=Workflow(file:///tmp/repro.yml)}: zizmor::audit::obfuscation: couldn't parse expression:  2 <= (3 == true)
 WARN audit:audit{input=Workflow(file:///tmp/repro.yml)}: zizmor::audit::secrets_outside_env: couldn't parse expression:  2 <= (3 == true)
 INFO audit: zizmor: 🌈 completed /tmp/repro.yml
warning[excessive-permissions]: overly broad permissions
 --> /tmp/repro.yml:4:3
  |
4 | /   test:
5 | |     runs-on: ubuntu-latest
6 | |     steps:
[...]

PR #1904, opened 2026-04-14
Fix parenthesized compound expression

DarkaMaul:
## Pre-submission checks

Please check these boxes:

      one first). (#1903 )

      use that violates the policy.

## Summary

Fix a crash in `Context::as_pattern` when a workflow contains a parenthesized compound expression in     
context position. The expression parser (since #1692) correctly parses these into a BinOp node, but as_pattern hit an `unreachable!()` on the unexpected variant.

## Test Plan

Test in both context.rs and full regression test.

### DIFF
--- a/crates/github-actions-expressions/src/context.rs
+++ b/crates/github-actions-expressions/src/context.rs
@@ -94,23 +94,22 @@ impl<'src> Context<'src> {
     /// Returns None if the context doesn't have a sensible pattern
     /// equivalent, e.g. if it starts with a call.
     pub fn as_pattern(&self) -> Option<String> {
-        fn push_part(part: &Expr<'_>, pattern: &mut String) -> bool {
+        fn transform_part<'src>(part: &'src Expr<'src>) -> Option<&'src str> {
             match part {
-                Expr::Identifier(ident) => pattern.push_str(ident.0),
-                Expr::Star => pattern.push('*'),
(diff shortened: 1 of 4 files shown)


crates/github-actions-expressions/src/context.rs
```diff
@@ -94,7 +94,7 @@ impl<'src> Context<'src> {
     /// Returns None if the context doesn't have a sensible pattern
     /// equivalent, e.g. if it starts with a call.
     pub fn as_pattern(&self) -> Option<String> {
-        fn push_part(part: &Expr<'_>, pattern: &mut String) {
+        fn push_part(part: &Expr<'_>, pattern: &mut String) -> bool {
```
woodruffw: I think the more idiomatic thing here would be to change the API here from `push_part -> bool` to `transform_part -> Option<&str>`, and then the caller can destructure the `Option` to determine whether or not to push.

woodruffw (changes requested):
Thanks @DarkaMaul, one suggestion.

crates/github-actions-expressions/src/context.rs
DarkaMaul: I've tried to come up with something, let me know what you think

woodruffw (approved):
Thanks @DarkaMaul! That approach looks a lot cleaner to me 🙂

PR #1913, opened 2026-04-15
feat(config): permit remapping rule severities

### LINKED ISSUE #1905: Feature: Remapping lint severities
### Pre-submission checks

We currently use Zizmor where we fail on high findings. We don't quite agree with Zizmor on _some_ lints, e.g. `artipacked` (medium), where we'd rather be safe than sorry. The upstream levels make sense, which is why we'd prefer to remap them if possible.

It could be as simple as `--remap artipacked=high`. A matching config format style might also make sense, something like:

```yaml
remap:
  artipacked: high
```

or:

```yaml
rules:
  artipacked:
    remap-severity: high
```

### Additional context

This can be worked around with `--format json` and parsing it, but that makes the solution hard to tell to internal customers (ie engineers) as they can't just be told: "Run `zizmor . --min-severity high --remap artipacked=high`"

If you think the config looks fine, I'm happy to look at implementing this and PRing :)

Sorry if this already exists; I know this can be formulated a million different ways. I tried my best to search but know that can have fallen short, unfortunately.

Proximyst:
## Pre-submission checks

Please check these boxes:

      one first).

      use that violates the policy.
  - This was AI-assisted with Claude Code and some good ol' fashioned manual work.

## Summary

This implements a new configuration option to remap severities of individual rules, as was discussed.

Please note that this contribution is provided by my employer Grafana Labs, not me as an individual. The code was created on work equipment & during working hours. We contribute it to the open-source project knowing it is not proprietary or similar. Happy to sign a CLA if need be!

Fixes: #1905

## Test Plan

There are new integration tests for this already, but also manually tested just to be safe:

```
$ ./target/debug/zizmor crates/github-actions-models/tests/sample-workflows/pwn-requests.yml --min-severity=high | grep artipacked
 INFO zizmor: 🌈 zizmor v1.24.1
 INFO audit: zizmor: 🌈 completed crates/github-actions-models/tests/sample-workflows/pwn-requests.yml

$$ ./target/debug/zizmor crates/github-actions-models/tests/sample-workflows/pwn-requests.yml --min-severity=high -c test-cfg.yml | grep artipacked
 INFO zizmor: 🌈 zizmor v1.24.1
 INFO audit: zizmor: 🌈 completed crates/github-actions-models/tests/sample-workflows/pwn-requests.yml
error[artipacked]: credential persistence through GitHub Actions artifacts
   = help: audit documentation → https://docs.zizmor.sh/audits/#artipacked

$ cat test-cfg.yml
rules:
  artipacked:
    remap:
      severity: high
```

### DIFF
--- a/crates/zizmor/src/finding.rs
+++ b/crates/zizmor/src/finding.rs
@@ -60,14 +60,9 @@ pub(crate) enum Confidence {
 
 #[derive(Copy, Clone, Debug, Eq, Hash, Ord, PartialOrd, PartialEq, Serialize, Deserialize)]
 pub(crate) enum Severity {
-    // We use aliases here to accept input in lowercase.
-    #[serde(alias = "informational")]
     Informational,
-    #[serde(alias = "low")]
     Low,
-    #[serde(alias = "medium")]
     Medium,
-    #[serde(alias = "high")]
     High,
 }
 
--- a/crates/zizmor/src/finding.rs
+++ b/crates/zizmor/src/finding.rs
@@ -60,9 +60,14 @@ pub(crate) enum Confidence {
 
 #[derive(Copy, Clone, Debug, Eq, Hash, Ord, PartialOrd, PartialEq, Serialize, Deserialize)]
 pub(crate) enum Severity {
+    // We use aliases here to accept input in lowercase.
+    #[serde(alias = "informational")]
     Informational,
+    #[serde(alias = "low")]
     Low,
+    #[serde(alias = "medium")]
     Medium,
+    #[serde(alias = "high")]
     High,
 }
 
--- a/crates/zizmor/src/registry.rs
+++ b/crates/zizmor/src/registry.rs
@@ -145,7 +145,15 @@ impl<'a> FindingRegistry<'a> {
     pub(crate) fn extend(&mut self, results: Vec<Finding<'a>>) {
         // TODO: is it faster to iterate like this, or do `find_by_max`
         // and then `extend`?
-        for finding in results {
+        for mut finding in results {
+            if let Some(remapped) = self
+                .input_registry
+                .get_config(finding.input_group())
+                .severity_remap(&finding)
+            {
+                finding.determinations.severity = remapped;
+            }
+
             if self.persona > finding.determinations.persona {
                 self.suppressed.push(finding);
             } else if finding.ignored
--- a/crates/zizmor/src/registry.rs
+++ b/crates/zizmor/src/registry.rs
@@ -146,13 +146,11 @@ impl<'a> FindingRegistry<'a> {
         // TODO: is it faster to iterate like this, or do `find_by_max`
         // and then `extend`?
         for mut finding in results {
-            if let Some(remapped) = self
+            finding.determinations.severity = self
                 .input_registry
                 .get_config(finding.input_group())
                 .severity_remap(&finding)
-            {
-                finding.determinations.severity = remapped;
-            }
+                .unwrap_or(finding.determinations.severity);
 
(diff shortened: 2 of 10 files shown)


woodruffw:
Thank you @Proximyst! I'll hopefully have time for review in the coming days.

crates/zizmor/src/registry.rs
```diff
@@ -145,7 +145,15 @@ impl<'a> FindingRegistry<'a> {
...
         // and then `extend`?
-        for finding in results {
+        for mut finding in results {
+            if let Some(remapped) = self
+                .input_registry
+                .get_config(finding.input_group())
+                .severity_remap(&finding)
+            {
+                finding.determinations.severity = remapped;
+            }
```
woodruffw: Nit: maybe `severity = (...).severity_remap(...).unwrap_or(severity)` instead? I personally find that a bit easier to read than `if let` 🙂

crates/zizmor/tests/integration/config.rs
```diff
@@ -284,3 +284,69 @@ fn test_invalid_configs() -> anyhow::Result<()> {
 
     Ok(())
 }
+
+/// Ensures that severity remapping changes the displayed severity of a finding.
+/// artipacked normally produces Medium; remapped to High here.
+#[test]
+fn test_severity_remap() -> anyhow::Result<()> {
+    insta::assert_snapshot!(
```
woodruffw, line 292: Nit: might be good to have a `--no-config` snapshot in this function too, just to assert that the remap doesn't get honored in that case 🙂

docs/configuration.md
```diff
@@ -172,6 +172,26 @@ Per-audit configuration, where `id` is the audit's name, e.g.
 
 Not all audits are configurable. See each audit's documentation for details.
 
+### `rules.<id>.remap`
+
+*Type*: `object`
+
+Remap parts of the rule. Currently, only severity is supported.
```
woodruffw, line 175: Nit: I think an `important` or `note` callout here would be helpful too, telling users that remapping is applied in a blanket fashion even if the individual finding severities aren't uniform.

Something roughly like:


```suggestion
Remap parts of the rule. Currently, only severity is supported.

!!! note

    Remappings are applied in a blanket fashion. For example, if you remap an audit's severity to `high`, both `medium` _and_ `low` findings will be re-graded as `high`.
```

crates/zizmor/src/finding.rs
```diff
@@ -60,9 +60,14 @@ pub(crate) enum Confidence {
 
 #[derive(Copy, Clone, Debug, Eq, Hash, Ord, PartialOrd, PartialEq, Serialize, Deserialize)]
 pub(crate) enum Severity {
+    // We use aliases here to accept input in lowercase.
+    #[serde(alias = "informational")]
```
woodruffw: It's kind of unfortunate that we need aliases here 😅 -- that's my fault from letting the internal (cased) representation bleed into the "v1" JSON format and other public interfaces.

Rather than aliasing here, perhaps we can repurpose the `SchemaRemapSeverity` type you added? That could go under `config` as `RemapSeverity` and the `impl From<RemapSeverity> for Severity` as a lossless transform.

woodruffw (commented):
Thanks @Proximyst, this looks really great! Mostly nitpicks on my part.

Proximyst:
Thanks for the review, @woodruffw! I think this should address your comments well. 

I also took the liberty in cdc94ee to make the severity an `Option`, as I realised we might not want it to have a default remap in the future if the remapping options are ever expanded 😅

woodruffw (approved):
Thank you @Proximyst!

PR #1953, opened 2026-04-28
fix: clarify available fix reporting

uncenter:
## Pre-submission checks

Please check these boxes:

      one first).

**Discussed on Discord.**

      use that violates the policy.

## Summary

Generally clarifies the reporting of available fixes, a) by outputting which mode (the current) is causing fixes to be filtered and suggesting flags to remedy this and b) by reporting fix counts split by disposition (safe or unsafe) rather than the total available fix count.

### DIFF
--- a/crates/zizmor/src/main.rs
+++ b/crates/zizmor/src/main.rs
@@ -2,7 +2,7 @@
 
 use std::{
     collections::HashSet,
-    env,
+    env, fmt,
     io::{Write, stdout},
     process::ExitCode,
 };
@@ -689,6 +689,16 @@ pub(crate) enum FixMode {
     UnsafeOnly,
     /// Apply all fixes, both safe and unsafe.
     All,
+}
+
+impl fmt::Display for FixMode {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        match *self {
+            FixMode::Safe => write!(f, "safe"),
+            FixMode::UnsafeOnly => write!(f, "unsafe-only"),
+            FixMode::All => write!(f, "all"),
+        }
+    }
 }
 
 /// State used when collecting input groups.
--- a/crates/zizmor/src/output/fix.rs
+++ b/crates/zizmor/src/output/fix.rs
@@ -49,8 +49,13 @@ pub fn apply_fixes(
 
     if fixes_by_input.is_empty() {
         if total_fixes > 0 {
+            let suggestion = match fix_mode {
+                FixMode::Safe => "Use --fix=unsafe or --fix=all to apply unsafe fixes.",
+                FixMode::UnsafeOnly => "Use --fix=safe or --fix=all to apply safe fixes.",
+                FixMode::All => unreachable!(),
+            };
             anstream::eprintln!(
-                "No fixes available to apply ({total_fixes} held back by fix mode)."
+                "No fixes available to apply ({total_fixes} held back by {fix_mode} mode). {suggestion}",
             );
         } else {
             anstream::eprintln!("No fixes available to apply.");
--- a/crates/zizmor/src/output/fix.rs
+++ b/crates/zizmor/src/output/fix.rs
@@ -50,12 +50,13 @@ pub fn apply_fixes(
     if fixes_by_input.is_empty() {
         if total_fixes > 0 {
             let suggestion = match fix_mode {
-                FixMode::Safe => "Use --fix=unsafe or --fix=all to apply unsafe fixes.",
-                FixMode::UnsafeOnly => "Use --fix=safe or --fix=all to apply safe fixes.",
-                FixMode::All => unreachable!(),
+                FixMode::Safe => Some("Use --fix=unsafe or --fix=all to apply unsafe fixes."),
+                FixMode::UnsafeOnly => Some("Use --fix=safe or --fix=all to apply safe fixes."),
+                FixMode::All => None,
             };
             anstream::eprintln!(
-                "No fixes available to apply ({total_fixes} held back by {fix_mode} mode). {suggestion}",
+                "No fixes available to apply ({total_fixes} held back by {fix_mode} mode).{}",
+                suggestion.map(|s| format!(" {s}")).unwrap_or_default()
             );
         } else {
             anstream::eprintln!("No fixes available to apply.");
--- a/crates/zizmor/src/output/plain.rs
+++ b/crates/zizmor/src/output/plain.rs
@@ -1,5 +1,6 @@
 //! "plain" (i.e. cargo-style) output.
 
+use itertools::Itertools;
 use std::collections::{HashMap, hash_map::Entry};
 
 use annotate_snippets::{Annotation, AnnotationKind, Group, Level, Renderer, Snippet};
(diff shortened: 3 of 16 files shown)


crates/zizmor/src/main.rs
```diff
@@ -691,6 +691,16 @@ pub(crate) enum FixMode {
...
 
+impl fmt::Display for FixMode {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        match *self {
+            FixMode::Safe => write!(f, "safe"),
+            FixMode::UnsafeOnly => write!(f, "unsafe-only"),
+            FixMode::All => write!(f, "all"),
+        }
+    }
+}
```
uncenter, line 702: Theoretically `clap` should already know what the kebab-case serializations of this enum are since it outputs them in `--help` but I couldn't figure out how to access those and reimplemented the Display trait here.

crates/zizmor/src/output/fix.rs
```diff
@@ -49,8 +49,13 @@ pub fn apply_fixes(
 
     if fixes_by_input.is_empty() {
         if total_fixes > 0 {
+            let suggestion = match fix_mode {
+                FixMode::Safe => "Use --fix=unsafe or --fix=all to apply unsafe fixes.",
+                FixMode::UnsafeOnly => "Use --fix=safe or --fix=all to apply safe fixes.",
+                FixMode::All => unreachable!(),
+            };
```
uncenter, line 56: I think it is worth it to suggest to the user how to apply the other fixes, but this is subjective and I can remove if you disagree.

crates/zizmor/src/output/fix.rs
```diff
@@ -49,8 +49,13 @@ pub fn apply_fixes(
 
     if fixes_by_input.is_empty() {
         if total_fixes > 0 {
+            let suggestion = match fix_mode {
+                FixMode::Safe => "Use --fix=unsafe or --fix=all to apply unsafe fixes.",
+                FixMode::UnsafeOnly => "Use --fix=safe or --fix=all to apply safe fixes.",
+                FixMode::All => unreachable!(),
```
uncenter: This feels icky, to have unreachable. Is there a better way I can do this?

crates/zizmor/src/output/plain.rs
```diff
@@ -126,11 +127,20 @@ pub(crate) fn render_findings(
...
 
-    let nfixable = findings.fixable_findings().count();
-    if nfixable > 0 {
+    let fixes_by_disposition: HashMap<FixDisposition, usize> = findings
+        .fixable_findings()
+        .flat_map(|finding| &finding.fixes)
+        .map(|fix| fix.disposition)
+        .counts();
+    let mut sorted_fixes_by_disposition: Vec<_> = fixes_by_disposition.iter().collect();
+    sorted_fixes_by_disposition.sort_by_key(|a| a.0);
```
uncenter, line 136: We sort so that the output is deterministic (do we list the safe fixes or the unsafe fixes first), e.g. for snapshot testing. It is a hashmap of only two entries (there are two dispositions), so it likely doesn't matter — however, I'm not very well versed in the data structure usage here. I went with a sorted Vec but I think a BTreeMap also works as a sorted alternative?

crates/zizmor/src/output/plain.rs
```diff
@@ -126,11 +127,20 @@ pub(crate) fn render_findings(
...
+    for (disposition, count) in sorted_fixes_by_disposition {
         qualifiers.push(format!(
-            "{nfixable} fixable",
-            nfixable = nfixable.bright_green()
+            "{} {disposition} fixes",
+            match disposition {
+                FixDisposition::Safe => count.bright_green().to_string(),
+                FixDisposition::Unsafe => count.bright_red().to_string(),
```
uncenter, line 142: Rust kept complaining about the match arms not having the same type if I don't use `.to_string()` for both here. I also tried `.color(BrightGreen)` and `.color(BrightRed)`, as well as `count.color(match { ... })` where the arms gave `BrightGreen` and `BrightRed` to no avail.

crates/zizmor/src/output/fix.rs
woodruffw: Hmm, one option is to have it be an empty string, i.e. `=> ""`.

But I think we could also probably specialize this a bit further with a custom tip, so no need to change anything here while I think about it 🙂

crates/zizmor/src/output/plain.rs
woodruffw: `BTreeMap` or `IndexMap` would work, yeah. I believe we already use `IndexMap` elsewhere so that would be a good pick here IMO.

crates/zizmor/src/output/plain.rs
woodruffw: Yeah, that's an unfortunate side effect of `owo_colors`'s API. Using `to_string` is fine!

crates/zizmor/src/main.rs
woodruffw: Makes sense, I think this is part of `ValueEnum` but not exposed publicly. Implementing `Display` seems fine!

crates/zizmor/src/output/fix.rs
woodruffw: Oh, a perhaps cleaner way would be to have this `match` product `Option<&'static str>`, and then `FixMode::All` will produce `None`. Then you can emit the suggestion only if `Some(_)`.

woodruffw (commented):
Thanks @uncenter, this looks good to me as a start! Just two small suggestions.

crates/zizmor/src/output/plain.rs
uncenter: Is that worth using here in this case though? I don't think we can do a conversion from itertools' `counts` `HashMap` into an `IndexMap` very easily from what I'm finding online, given that the IndexMap is sorted by insertion order. Alternatively, we could just manually get the counts for each of the (two) dispositions into separate variables and print them a specific way.

crates/zizmor/src/output/fix.rs
uncenter: Done!

crates/zizmor/src/output/fix.rs
```diff
@@ -49,8 +49,14 @@ pub fn apply_fixes(
...
     if fixes_by_input.is_empty() {
         if total_fixes > 0 {
+            let suggestion = match fix_mode {
+                FixMode::Safe => Some("Use --fix=unsafe or --fix=all to apply unsafe fixes."),
+                FixMode::UnsafeOnly => Some("Use --fix=safe or --fix=all to apply safe fixes."),
+                FixMode::All => None,
+            };
             anstream::eprintln!(
-                "No fixes available to apply ({total_fixes} held back by fix mode)."
+                "No fixes available to apply ({total_fixes} held back by {fix_mode} mode).{}",
```
uncenter, line 58:
```suggestion
                "No fixes available to apply ({total_fixes} held back in {fix_mode} mode).{}",
```

Teeny thing: this alternate phrasing might be slightly clearer.

woodruffw:
Merged with #1971, thanks @uncenter!

PR #1954, opened 2026-04-29
fix(cache-poisoning): evaluate certain cache control expressions

### LINKED ISSUE #1940: [BUG]: zizmor currently flags maturin's generated workflow for `sccache: ${{ !startsWith(github.ref, 'refs/tags/') }}` being vulnerable to a cache poisoning attack
### Pre-submission checks

### zizmor version

1.24.1

### Expected behavior

It was reported in this [issue](https://github.com/PyO3/maturin/issues/2425#issuecomment-2571654034) that the error shown by `zizmor` for `sccache: ${{ !startsWith(github.ref, 'refs/tags/') }}` could be a potential false-positive.

### Actual behavior

`zizmor` currently reports `sccache: ${{ !startsWith(github.ref, 'refs/tags/') }}` as being vulnerable to a cache poisoning attack.

### Reproduction steps

1. Generate a workflow using [`maturin generate-ci`](https://www.maturin.rs/distribution.html?highlight=generate-ci#github-actions)
2. Run `uvx zizmor` on the generated workflow.

### Logs

```plain text
error[cache-poisoning]: runtime artifacts potentially vulnerable to a cache poisoning attack
   --> ./.github/workflows/build-and-release.yaml:126:9
    |
  8 | / on:
  9 | |   push:
 10 | |     branches:
 11 | |       - master
 12 | |     tags:
 13 | |       - '*'
    | |___________- generally used when publishing artifacts generated at runtime
...
126 |           uses: PyO3/maturin-action@e83996d129638aa358a18fbd1dfb82f0b0fb5d3b # v1.51.0
    |           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ this step
127 | /         with:
128 | |           target: ${{ matrix.platform.target }}
129 | |           args: --release --out dist --find-interpreter
130 | |           sccache: ${{ !startsWith(github.ref, 'refs/tags/') }}
    | |_______________________________________________________________- may enable caching here
    |
    = note: audit confidence → Low
```

### Additional context

If it helps this is the [workflow](https://github.com/py-econometrics/pyfixest/blob/master/.github/workflows/build-and-release.yaml) I ran `zizmor` on, and got the error. Additionally locally I also ran `pinact` to pin the commit SHAs, before running `zizmor` on the workflow.

reubenwong97:
## Pre-submission checks

Please check these boxes:

- [x] **Mandatory**: This PR corresponds to an issue #1940 .

      use that violates the policy.

## Summary

Parses known patterns in cache control expressions to determine if cache is enabled.

## Test Plan

Issue reproduction integration test.

Closes #1940


reubenwong97:
Hi @woodruffw ,

Can you help to take a look at this when you have time? This is one of the fixes I'm less confident about in my approach.. Broadly speaking I attempted to parse the inner expression into nodes that we recognize and are relevant (in this specific case, it is very targeted to `${{ !startsWith(github.ref, 'refs/tags/') }}`. Unrecognized expressions are passed through to the rest of the audit.

I am also not too pleased with the many `let`-`else`'s I have, but I'm not sure if there is a better way around this..

Thank you!

woodruffw:
Thanks @reubenwong97, I'll be able to review this in the next day or so.

woodruffw (commented):
Thanks @reubenwong97! This approach looks good to me overall, I'm just going to push some tweaks + docs.

crates/zizmor/src/audit/cache_poisoning.rs
```diff
@@ -282,28 +392,42 @@ audit_meta!(
 );
 
 impl CachePoisoning {
-    fn trigger_used_when_publishing_artifacts(&self, trigger: &Trigger) -> bool {
+    fn trigger_used_when_publishing_artifacts(&self, trigger: &Trigger) -> Option<ReleaseTrigger> {
```
woodruffw: Unfortunately I think we need to generalize this even further: technically a workflow can have multiple "release" triggers in the way we mean here, e.g.:

```yaml
on:
  release:
  push:
    tags:
      - "v*"
```

To capture this I think we need to turn this from `Option<ReleaseTrigger>` to `Vec<ReleaseTrigger>` and then specialize our heuristic below to apply only when our _only_ release trigger is `ReleaseTrigger::TagPush`.

woodruffw: [`5202a79` (#1954)](https://github.com/zizmorcore/zizmor/pull/1954/commits/5202a79bd70e2175511a91bad6e9bd05592074fa)

woodruffw (approved):
Thanks @reubenwong97! I think there's a lot of other heuristics we could apply here over time (like for other triggers), but this is a great foundation.

Issue #1958, opened 2026-04-30
Feature: support --fix to reduce excessive permissions


vmvarela:
### Pre-submission checks

- [x] I am **not** reporting a bug (crash, false positive/negative, etc). These must be filed via the bug report template.
- [x] I have looked through both the [open](https://github.com/zizmorcore/zizmor/issues?q=is%3Aissue+is%3Aopen+) and [closed](https://github.com/zizmorcore/zizmor/issues?q=is%3Aissue%20state%3Aclosed) issues for a duplicate request.

### What's the problem this feature will solve?

I need to correct the permissions of more than a thousand repositories in my Github organization, so it's a personal necessity.

### Describe the solution you'd like

Add --fix support to the excessive-permissions audit.

woodruffw:
@vmvarela Please affirm that you didn't use an LLM to generate this issue. This has many of the same hallmarks as your previous issue.

vmvarela:
This is my need and I have opened a PR for this.
I hope this will be useful for somebody else.

NOTE: I have used LLM to write the code.

woodruffw:
Sorry, but that's not an affirmation. Please don't do this; your PR also violates our [AI policy](https://github.com/zizmorcore/.github/blob/main/AI_POLICY.md) in multiple ways:

- You didn't wait for this issue to be greenlit;
- Your PR exceeds 750 lines;
- You didn't use the PR template, which explicitly requires you to disclose the use of LLMs *in* the PR's submission as well.

The underlying idea here is fine and I want to accommodate your use case, but this is not a considerate way to collaborate on this project.

To reset: can you say more about your use case? What is your organization, and why do you have 1000 repositories that need to have their permissions dropped automatically? For most people, this is a relatively low-priority fix if their repositories are private, since there's much less external exposure.

vmvarela:
We need to secure 1000+ existing repositories within a private GitHub Enterprise organization, without any "public" repositories. For too long, hundreds of workflows have been created without security review, and a comprehensive audit is necessary, for which "zizmor" could be useful. One area for improvement in our workflows is specifying the minimum permissions required, as global permissions are too often used. "zizmor" detects this, but it would be great if it could also fix it (at least in clear cases).

vmvarela:
Sorry for the inconvenience. It's clear I didn't read the instructions.

woodruffw:
It's okay, just please read the policy carefully in the future!

[...]

PR #1959, opened 2026-04-30
feat(excessive-permissions): infer minimum GITHUB_TOKEN permissions from used actions

vmvarela:
## Summary

Solves #1958

Adds permission inference to the `excessive-permissions` audit: for each job, walks its `uses:` steps, looks up each action in a built-in KB, and emits an unsafe auto-fix with the minimum required `GITHUB_TOKEN` scopes.

## What's new

- **Inference engine** — computes minimum per-job permissions from known actions; falls back to `{}` when any action is unknown
- **Built-in KB** (`data/action-permissions.json`, ~58 entries) embedded at compile time
- **`--action-kb <FILE>` / `ZIZMOR_ACTION_KB`** — override/extend the built-in KB
- **`yamlpatch`** — fix block-mapping rendering so permission patches produce valid YAML
- **`Permission`** — derive `Clone + Copy` (required by the inference engine)

## Fixes behaviour

| Scenario | Fix |
|---|---|
| `write-all` / `read-all` at workflow level | `{}` |
| `write-all` / default at job level, all actions known | Inferred minimum scope map |
| `write-all` / default at job level, unknown action | `{}` |
| Explicit `write` scope at workflow level | No fix (move to job manually) |

All fixes are `FixDisposition::Unsafe` and never applied automatically.


woodruffw:
Please follow our [AI policy](https://github.com/zizmorcore/.github/blob/main/AI_POLICY.md) when opening PRs.

(There are multiple significant changes here that would require design agreement and consensus, like the "KB" concept.)

PR #1962, opened 2026-05-01
Add a hidden span to cache poisoning findings

woodruffw:
See https://github.com/zizmorcore/zizmor/issues/1940#issuecomment-4357491560.

Very similar to #1945.

h/t @2bndy5 for the report 🙂


2bndy5:
I tested on the snippet posted in the comment/report, and it works! Thanks the speedy response/action. 🥇

woodruffw:
Glad to hear it!

PR #1977, opened 2026-05-06
fix(ref-version-mismatch): allow name/ver refs

### LINKED ISSUE #1938: BUG: `ref-version-mismatch` should not flag weird-but-valid refs
### Pre-submission checks

In pedantic mode, Zizmor currently flags `name/vz.y.z` tags as invalid.

```
help[ref-version-mismatch]: action's hash pin has mismatched or missing version comment
  --> .github/workflows/update-pnpm.yml:44:9
   |
44 |         uses: grafana/shared-workflows/actions/create-github-app-token@ae92934a14a48b94494dbc06d74a81d47fe08a40 # create-github-app-...
   |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ comment does not contain a version
   |
   = tip: rewrite comment to include '# create-github-app-token/v0.2.2'
   = help: audit documentation → https://docs.zizmor.sh/audits/#ref-version-mismatch
```

It'd be nice to be able to centrally exclude specific actions using those tags from the audit, via zizmor.yml, rather than needing to exclude each usage across many workflow files. Similar to what `forbidden-uses` allows you to configure: https://docs.zizmor.sh/audits/#rulesforbidden-usesconfigallowdeny

### Additional context

It could maybe be considered a bug that in the above output provided, the `tip` recommends the exact comment that the line already has?

reubenwong97:
## Pre-submission checks

Please check these boxes:

- [x] **Mandatory**: This PR corresponds to an issue #1938 .

      use that violates the policy.

## Summary

Adds a new regex to match for `<name>/<version>` refs. `<name>` allows for any level of nesting of `/`.

## Test Plan

Added an integration test with the issue reproduction.

Closes #1938

### DIFF
--- a/crates/zizmor/src/audit/ref_version_mismatch.rs
+++ b/crates/zizmor/src/audit/ref_version_mismatch.rs
@@ -11,7 +11,7 @@ use crate::{
     config::Config,
     finding::{
         Confidence, Finding, Fix, Persona, Severity,
-        location::{Comment, Feature, Location, Routable},
+        location::{Feature, Location, Routable},
     },
     github,
     models::{StepCommon, action::CompositeStep, uses::RepositoryUsesExt, workflow::Step},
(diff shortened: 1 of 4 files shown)


reubenwong97:
Hi @woodruffw ,

Can you take a look when you have some time? I believe all that is missing is a regex pattern matching this type of comment. I'm just unsure if the pattern matched here may be a little too broad.. Thanks!

woodruffw:
Thanks @reubenwong97! 

I actually think we want a slightly different approach here -- instead of matching patterns like `name/vers`, I think we want to collect _everything_ that might be a valid ref name and add it to the `CommentVersionState::NonVersionComments` variant. Then, if any of those match the resolved name, we have a match and can skip the rest of the audit.

(In practice I think that means we don't need to pattern match the comments at all, since almost anything can be a valid ref name. We should just stash them and compare them against the resolved ref.)

reubenwong97:
> (In practice I think that means we don't need to pattern match the comments at all, since almost anything can be a valid ref name. We should just stash them and compare them against the resolved ref.)

Hi @woodruffw ,

Sorry I have been busy the past few days. Would the above suggestion imply something like:
```rust
#[derive(Clone, Copy, Debug, Eq, PartialEq)]
enum CommentVersionState<'doc> {
    Missing,
    RefCandidates(&'doc [Comment<'doc>]),
}
```
instead, and end up getting rid of the regexes completely?

Thanks!

woodruffw:
No worries at all. Yeah, that's close to what I was thinking -- either we can drop the version regexes entirely (and only use them later on for the actual patch applicability part) _or_ we can add data to the `NonVersionComments` variant. Either seems fine to me.

reubenwong97:
Hi @woodruffw ,

Can you take a look at this approach and see if its fine? I'll loop through all comments, treating them as potential refs. I think the regex is still required as some comments like `# tag=v3.0.0` need to extract out the version first.. Thanks!

woodruffw:
Thanks @reubenwong97, I'll be able to look at this in the next few days.

woodruffw:
Noting: I took another look at this, but I want to do a last cleanup pass on it before merging. I think the overall pattern logic here can be simplified significantly (not the fault of this PR, just how the audit has accreted changes over time).

PR #1985, opened 2026-05-11
New audit: typosquat-uses

### LINKED ISSUE #1989: Feature: new audit for action typosquatting
### Pre-submission checks

- [x] I have looked through both the open and closed issues for a duplicate request. The closest prior issue is https://github.com/zizmorcore/zizmor/issues/1115 (container registry typosquatting, closed as dupe of https://github.com/zizmorcore/zizmor/issues/737). Action-name typosquatting is unclaimed.

Workflows can be redirected to attacker-controlled code by a single mistyped owner in a `uses:` reference. `action/checkout` instead of `actions/checkout`, `dokcer/login-action` instead of `docker/login-action`. If the misspelled namespace is registered by someone else, the workflow fetches and runs their action with whatever permissions the job has.

zizmor catches archived actions and impostor commits but does nothing about look-alike slugs. A scan of ~150k repositories surfaced about 15 near-misses, two of which resolved to live repos under a different owner. The yield is small but the failure mode is total compromise of the job, so it seems worth a check.

A new `typosquat-uses` audit that:

- Compares each `uses:` slug against a baked-in corpus of popular actions using the [typomania](https://github.com/rustfoundation/typomania) library (omission, transposition, repetition, homoglyph, keyboard typos).
- Treats the owner as the package author so near-misses within the same org (e.g. `actions/chckout`) are suppressed. Same-org typos are harmless because the legitimate owner already controls the namespace and the step just 404s at runtime.
- Reports at low confidence offline.
- When a GitHub token is available, checks whether the misspelled slug resolves to a live repository and raises confidence to high if it does.

A draft implementation is up at https://github.com/zizmorcore/zizmor/pull/1985.

### Additional context

Open questions worth resolving in review:

- Whether a hand-seeded ~50-entry corpus is the right shape or whether it should be generated by a `support/popular-actions.py` script the way `archived-repos.txt` is refreshed.
- Whether to ship as Regular persona (current draft) or start at Pedantic given how rarely it fires.

andrew:
## Pre-submission checks

- [x] **Mandatory**: This PR corresponds to an issue (if not, please create one first).
- [x] Having read the [AI policy], I hereby disclose the use of an LLM or other AI coding assistant in the creation of this PR. PRs will not be rejected for using AI tools, but *will* be rejected for undisclosed use or use that violates the policy.

Closes #1989. Developed with Claude Code (Opus 4.7); I've reviewed the diff and can explain every change without referring back to the tool.

## Summary

Adds a `typosquat-uses` audit that flags `uses:` references one edit away from a popular action but owned by a different account, e.g. `action/checkout` or `dokcer/login-action`.

Uses the [typomania](https://github.com/rustfoundation/typomania) library against a baked-in corpus of ~50 popular actions in `crates/zizmor/data/popular-actions.txt`. The owner is treated as the package author so near-misses within the same org (e.g. `actions/chckout`) are ignored, since that namespace is already controlled and the step just 404s at runtime.

Works offline at low confidence. With a GitHub token it checks whether the misspelled repo actually exists and bumps confidence to high when it does. Adds `Client::repo_exists` for the lookup.

In a scan of ~150k repos this pattern surfaced ~15 near-misses, almost all harmless same-org typos. Only two resolved to live repos under a different owner. So the value here is less about finding active squats and more about catching the dangerous case where a typo'd owner name points at a namespace someone else could register.

The corpus is hand-seeded for now. Happy to add a `support/popular-actions.py` generator along the lines of `archived-repos` if the approach looks right.

## Test Plan

Integration test at `crates/zizmor/tests/integration/audit/typosquat_uses.rs` with a fixture covering omission, transposition, homoglyph, repetition, and owner-transposition positive cases plus two negative cases (same-owner near-miss, unrelated slug). Snapshot is inline. The online (high-confidence) branch is not covered by a test since it needs a live foreign-owned squat; tested manually that the offline path produces correct findings.


crates/zizmor/src/audit/typosquat_uses.rs
```diff
@@ -0,0 +1,286 @@
...
+    }
+
+    fn get(&self, name: &str) -> typomania::Result<Option<&dyn Package>> {
+        Ok(self.0.get(name).map(|p| p as &dyn Package))
+    }
+}
+
+struct ActionSlug {
+    owners: HashSet<String>,
+}
```
woodruffw: Can we use the existing `RepositoryUsesPattern` API for this, rather than a new  `ActionSlug` type? I think the two have similar purposes.

crates/zizmor/src/audit/typosquat_uses.rs
```diff
@@ -0,0 +1,286 @@
...
+
+impl ActionSlug {
+    fn new(slug: &str) -> Self {
+        let owner = slug.split('/').next().unwrap_or(slug).to_lowercase();
+        Self {
+            owners: HashSet::from([owner]),
```
woodruffw: Nit (assuming we need `ActionSlug` instead of `RepositoryUsesPattern`): I _think_ this could be `owner: String`?

crates/zizmor/src/github.rs
```diff
@@ -604,6 +604,19 @@ impl Client {
             .map_err(Into::into)
     }
 
+    #[instrument(skip(self))]
+    pub(crate) async fn repo_exists(&self, owner: &str, repo: &str) -> Result<bool, ClientError> {
+        let url = format!("{api_base}/repos/{owner}/{repo}", api_base = self.api_base);
```
woodruffw: Unfortunately I think this won't scale super well for large repos -- GitHub will ratelimit it pretty quickly, plus (IIUC) it returns a largish JSON blob over the wire.

I _think_ we could probably use the Git API directly for this, i.e. tweak our current `Client::list_refs` API to return `None` when the response 404s. That should be lighter in terms of response + ratelimits, plus (in the future) we can synthetically cache it since we're already heavily using that route for ref listing in numerous audits.

docs/audits.md
```diff
@@ -1728,6 +1728,59 @@ shell quoting/expansion rules.
...
+
+=== "After"
+
+    ```yaml title="typosquat.yml" hl_lines="3"
+    - name: checkout
+      uses: actions/checkout@v4
+    ```
+
+If the reference is intentional (for example, a legitimate fork with a similar
+name), suppress the finding with an [ignore comment](./usage.md#ignoring-results).
```
woodruffw: Nit: I think we can remove this, since it's the same guidance as for all false positives 🙂 


```suggestion
```

crates/zizmor/src/registry.rs
```diff
@@ -75,6 +75,7 @@ impl AuditRegistry {
         register_audit!(audit::dependabot_cooldown::DependabotCooldown);
         register_audit!(audit::concurrency_limits::ConcurrencyLimits);
         register_audit!(audit::archived_uses::ArchivedUses);
+        register_audit!(audit::typosquat_uses::TyposquatUses);
```
woodruffw, line 78: Nit: Make sure to add this to `config/schema.rs` as well and run `make generate-schema` after to bump the `zizmor.yml` config schema!

crates/zizmor/src/audit/typosquat_uses.rs
```diff
@@ -0,0 +1,286 @@
...
+        StepCommon as _,
+        action::CompositeStep,
+        workflow::{ReusableWorkflowCallJob, Step},
+    },
+    state::AuditState,
+};
+
+const ALPHABET: &str = "abcdefghijklmnopqrstuvwxyz0123456789-_./";
+
+static TYPOS: &[(char, &[&str])] = &[
```
woodruffw, line 28: Just making sure I understand (since I'm not super familiar with typomania): is this a table of commit edit errors, based on keyboard layouts?

crates/zizmor/src/audit/typosquat_uses.rs
```diff
@@ -0,0 +1,286 @@
...
+                    Confidence::High,
+                    format!("{slug} {squat} and resolves to a live repository"),
+                ),
+                Ok(false) => (
+                    Confidence::Low,
+                    format!("{slug} {squat} (currently unregistered)"),
+                ),
+                Err(_) => (Confidence::Low, format!("{slug} {squat}")),
+            },
+            None => (Confidence::Low, format!("{slug} {squat}")),
```
woodruffw: Nit: I think it'd be cool/nice to add a tip to this that the user can enable an online audit to increase the finding's confidence.

(See `FindingBuilder::tip` for that.)

crates/zizmor/src/audit/typosquat_uses.rs
woodruffw: Alternatively if that doesn't work we could _probably_ store just the `RepositoryUses::slug` here.

woodruffw (commented):
Thanks @Andrew! I agree this would be a great audit to have, and thanks a ton for working on it.

Mostly nits from me; the only general question I have is around case handling -- right now the `uses:` clauses get folded to lowercase; are there any concerns/risks with doing that? The answer is probably 'no' but I'm not 100% sure since I'm not sure how typomania works 🙂

crates/zizmor/src/audit/typosquat_uses.rs
andrew: I had a look but `RepositoryUsesPattern` is a matcher enum and doesn't map cleanly onto typomania's `Package` trait (which wants a single author identity for the shared-author check). Kept `ActionSlug` but simplified it to hold just `owner: String` per the next comment.

crates/zizmor/src/audit/typosquat_uses.rs
andrew: Done in 79e8f0f.

crates/zizmor/src/github.rs
andrew: Good call. Reworked in 79e8f0f to call `list_refs` and map `RepoMissingOrPrivate` to `Ok(false)`. That reuses the existing ref cache so a hit from another audit means this is free, and a miss is the lighter Git protocol path rather than the REST blob.

docs/audits.md
andrew: Removed in 79e8f0f.

crates/zizmor/src/registry.rs
andrew: Added and regenerated in 79e8f0f.

crates/zizmor/src/audit/typosquat_uses.rs
andrew: Yes, it's keyboard-adjacency substitutions (QWERTY) plus a few visual confusables like `m`→`rn` and `1`→`l`. The table is lifted from typomania's example which in turn comes from the crates.io typosquat checker. Added a comment explaining that in 79e8f0f.

crates/zizmor/src/audit/typosquat_uses.rs
andrew: Done in 79e8f0f. The tip only shows in the offline branch.

andrew:
Thanks for the review, all addressed in 79e8f0f.

On case folding: GitHub treats owner and repo names as case-insensitive (`Actions/Checkout` resolves to `actions/checkout`), so lowercasing both the corpus and the candidate before handing them to typomania is correct and stops mixed-case slugs from slipping past. typomania itself does byte-wise comparison so without the fold `Actions/checkout` wouldn't match the corpus entry.

crates/zizmor/src/audit/typosquat_uses.rs
```diff
@@ -0,0 +1,287 @@
...
+    fn contains_name(&self, name: &str) -> typomania::Result<bool> {
+        Ok(self.0.contains_key(name))
+    }
+
+    fn get(&self, name: &str) -> typomania::Result<Option<&dyn Package>> {
+        Ok(self.0.get(name).map(|p| p as &dyn Package))
+    }
+}
+
+struct ActionSlug {
```
woodruffw: Nit: I think this type might be cleaner as `ActionOwner` or similar, since we use the term "slug" elsewhere to refer to the full `owner/repo` slug but this just carries the owner.

crates/zizmor/src/audit/typosquat_uses.rs
```diff
@@ -0,0 +1,287 @@
...
+        &self,
+        uses: &RepositoryUses,
+    ) -> Option<(FindingBuilder<'doc>, String)> {
+        let slug = format!(
+            "{}/{}",
+            uses.owner().to_lowercase(),
+            uses.repo().to_lowercase()
+        );
+
+        let candidate: Box<dyn Package> = Box::new(ActionSlug::new(&slug));
```
woodruffw: I _think_ this works and saves a clone + saves a string munge.

```suggestion
        let candidate: Box<dyn Package> = Box::new(ActionSlug::new(&uses.owner()));
```

woodruffw (approved):
Thanks @Andrew! I'll rebase and address some nits on my end, but this looks good to me.

andrew:
I've also got a few improvement prs open on typomania, so hopefully there will be a new release of that before 1.2.6 comes around

woodruffw:
(This also reminded me that I need to break `RepositoryUses` up a bit -- it should be renamed to `ActionUses` and should have an `ActionSlug` sub-type for just `owner/repo` pairs.)

woodruffw:
Thank you @Andrew!

PR #1995, opened 2026-05-13
Add opt-in warning mode for inaccessible GitHub repos

### LINKED ISSUE #1350: UX: Poor experience across multiple private repos
See https://github.com/zizmorcore/zizmor-action/issues/55 for an instance of this.

TL;DR: some people use zizmor in private repos, which in turn reference actions/reusable workflows from _other_ private repos. For example, `ci.yml` in `secret-org/secret-repo` might reference `secret-org/secret-action@v1`.

When users do this in CI/CD (e.g. with zizmor-action), they get snarled: by default zizmor-action tries to do online audits, which then fails on 404s because the default `secrets.GITHUB_TOKEN` doesn't have cross-org private repo access. This results in confusing failures for users where zizmor refuses to continue because it can't access remote resources that it needs to check.

See https://github.com/orgs/community/discussions/46566 for details on that `secrets.GITHUB_TOKEN` limiation.

Some potential solutions:

1. Make zizmor more permissive by default of 404s against remote resources. I'm very hesitant to do this, because IME users ignore warnings about missing resources/failed requests and never go back to confirm that something important wasn't missed.
2. Nudge users more heavily towards using PATs in this scenario. I don't love this because (1) PATs are annoying, (2) it requires another layer of special case documentation ("you only need PATs if..."), and (3) users have a tendency to overscope PATs (whereas zizmor only ever needs read-only access).
3. Add an *opt in* CLI flag, something like `--allow-github-api-failures` (this is a terrible name). With this users would get a warning instead of a hard error, so (1) but not by default. I'm okay with this I think, but it's going to require a better CLI flag name.

Overall I think (3) is the right path forwards, at least for now. But this is also pretty low-priority at the moment, since this veers towards the enterprise side of the feature spectrum _and_ users can always unblock themselves by running zizmor locally or with a PAT (or entirely offline).

rafael81:
## Summary

Closes #1350.

This adds an opt-in `--warn-on-github-api-failures` audit option for private-repo deployments where online audits encounter referenced GitHub repositories that are missing or inaccessible to the configured token.

Behavior is intentionally narrow: the new option downgrades only audit failures caused by `RepoMissingOrPrivate` into warnings, then lets the remaining audits continue. Other GitHub/API failures still fail normally, so transient outages, rate limits, and unrelated client errors are not hidden.

## Validation

- `cargo fmt --check`
- `cargo test -p zizmor audit::tests::audit_error`
- `cargo clippy -p zizmor --bin zizmor -- -D warnings`
- `cargo test -p zizmor test_stdin_workflow`
- `cargo test -p zizmor warn_on_min_severity_unknown`
- `GH_TOKEN=$(gh auth token) cargo test -p zizmor issue_1286 --features gh-token-tests`
- `GH_TOKEN=$(gh auth token) cargo test -p zizmor issue_1350_warn_on_github_api_failures --features gh-token-tests`
- Manual repro: `GH_TOKEN=$(gh auth token) cargo run -p zizmor -- --no-progress --show-audit-urls=never --warn-on-github-api-failures /Users/user/project/4dollor/zizmor/crates/zizmor/tests/integration/test-data/issue-1286.yml`

Disclosure: this PR was prepared with AI assistance, then manually reviewed and tested before submission.

### DIFF
--- a/crates/zizmor/src/audit/mod.rs
+++ b/crates/zizmor/src/audit/mod.rs
@@ -224,6 +224,13 @@ impl AuditError {
     pub(crate) fn ident(&self) -> &'static str {
         self.ident
     }
+
+    pub(crate) fn is_repo_missing_or_private(&self) -> bool {
+        self.source.chain().any(|err| {
+            err.downcast_ref::<crate::github::ClientError>()
+                .is_some_and(crate::github::ClientError::is_repo_missing_or_private)
+        })
+    }
 }
 
 /// Auditing trait.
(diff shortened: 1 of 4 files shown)


woodruffw:
Please follow our [AI Policy](https://github.com/zizmorcore/.github/blob/main/AI_POLICY.md) and make sure to use our PR template when filing PRs. PRs that don't use the template will not be reviewed.

PR #1998, opened 2026-05-14
impostor-commit: handle tag SHAs properly

### LINKED ISSUE #1997: [BUG]: impostor-commit burns thousands of REST calls per non-commit SHA pin, exhausting the 1,000/hr actions GITHUB_TOKEN in a single run
### Pre-submission checks

### zizmor version

v1.24.1

### Expected behavior

When a `uses: owner/repo@<sha>` pin's SHA doesn't resolve as a commit (annotated tag-object SHA, typo, deleted commit) I expect `impostor-commit` to flag it and move on, at a cost of a few of REST calls per offending pin.

### Actual behavior

A single such pin causes `impostor-commit` to walk every branch and every tag in the upstream repo via `compare_commits`, multiplied by the retry middleware that treats most 4xx as retryable.

For `github/codeql-action` (349 branches + 544 tags = 893 refs at time of writing), one bad pin produces ~3,572 REST calls per invocation (893 × 4: one initial call + three retries on 404).

The default `GITHUB_TOKEN` inside Actions is 1,000 REST requests / hour / repo; a GitHub App installation token is 5,000 / hour. So one bad pin overshoots a default token by ~3.5x in a single zizmor run, and chews ~71% of an App-installation budget.

Two pieces of code combine:

1. The retry classifier in `crates/zizmor/src/github.rs` (around lines 272–294) marks essentially every 4xx retryable, capped at 3 retries. PR #1609 narrows this to "every 4xx except 403" so 403 rate-limits can be surfaced cleanly. 404 is still retried 3×, but a 404 from `compare/A...B` meaning "A or B isn't a commit in this repo" can never succeed on retry. Each one costs four calls.

2. `compare_commits` (same file, ~lines 620–629) catches 404 specifically and maps it to `Ok(None)`. `named_ref_contains_commit` in `crates/zizmor/src/audit/impostor_commit.rs` then treats `None` as "diverged refs, keep looking", so the slow-path loop at the bottom of `impostor()` (~lines 123–139) runs to completion against every branch then every tag (even though every call is a 404 against the same bogus head SHA).

Net cost per bad pin: `4 x (branches + tags in upstream repo)` REST calls. Popular targets (`github/codeql-action`, `actions/checkout`, etc.) make this very expensive.

### Reproduction steps

1. In a workflow, pin an action to its **annotated tag-object SHA** instead of the commit SHA the tag dereferences to. Real example using `github/codeql-action`'s `v4.35.4`:
   ```yaml
   uses: github/codeql-action/upload-sarif@3ce22a6e336a7fcc318bc58ae1986395bdc83ba7 # v4.35.4
   ```
   `3ce22a6e…` is the tag-object SHA; the underlying commit is `68bde559dea0fdcac2102bfdf6230c5f70eb485e`. Verify:
   ```
   gh api repos/github/codeql-action/git/ref/tags/v4.35.4
     # .object.type == "tag"  (annotated)
     # .object.sha  == 3ce22a6e...
   gh api repos/github/codeql-action/commits/3ce22a6e336a7fcc318bc58ae1986395bdc83ba7
     # 422 No commit found for SHA
   gh api repos/github/codeql-action/compare/main...3ce22a6e336a7fcc318bc58ae1986395bdc83ba7
     # 404
   ```
2. Run `zizmor --format=sarif .github/` against that workflow with a fresh token.
3. Read `X-RateLimit-Used` before and after, or watch `gh api rate_limit`. With a `codeql-action`-sized target it climbs by ~3,500 from the one pin.

### Logs

Full Actions run from the incident: https://github.com/contextbridge/planbridge/actions/runs/25812188244/job/75831162270

The symptom is downstream rate-limit exhaustion on subsequent API calls in the workflow, not a zizmor stderr error.

### How I hit this in the wild
I asked a coding agent to convert every `uses: foo/bar@v1.2.3` in my workflows to its SHA-pinned form. For the annotated tags in `github/codeql-action` the agent read `GET /repos/{owner}/{repo}/git/ref/tags/{tag}` and pasted `.object.sha` without dereferencing the annotated-tag object underneath. The resulting pin runs fine in Actions — the runner accepts the tag-object SHA — so there was no visible signal that anything was wrong.

What I noticed instead was that my CI sat rate-limited for an hour. I'd already routed the zizmor step through a GitHub App token (5,000/hr installation budget) because the default `GITHUB_TOKEN` had been getting exhausted on busy PR days; with the bad pin in place, that App token got exhausted too. Two one-hour windows total before I traced it back to the pin.

### Ideas

Two cheap mitigations, each of which independently breaks the cascade:

- **Extend #1609's retry-classifier narrowing.** Once #1609 lands, only 403 is excluded. A safer default is to retry only 408, 429, and 5xx. Any other 4xx is "the server understood the request and the request was bad" — retrying cannot succeed and just multiplies the wasted-call count.
- **Short-circuit `impostor()` when the head SHA isn't a commit.** Before entering the slow-path branch/tag loop, do a single `GET /repos/{owner}/{repo}/commits/{sha}` (or `GET .../git/commits/{sha}`) on the head SHA. If GitHub returns 404/422, the SHA isn't a commit in this repo and `impostor()` can return `Ok(true)` immediately. One call instead of `4 × (branches + tags)`. This complements #1417's fast path (which targets the case where the SHA *is* a commit somewhere in the fork network).

Either fix alone would have made my one bad pin cost ~5 calls instead of ~3,500.

Relation to prior issues, restated for the triager:
- #151: false-positive *finding* is acceptable; this issue is about the API-call *cost* of producing the false positive.
- #1417: added the `branch_commits` fast path. When that endpoint errors, the audit falls through to the slow path, which is what blows the budget here.
- #1252 / PR #1609: same shape of "specific 4xx isn't really retryable", but narrower carve-out (403 only). This issue argues for broadening it and for an orthogonal short-circuit in `impostor()`.

**AI Disclaimer :robot:**: I used Claude Opus 4.7 to help diagnose the problem and draft this issue. I human-reviewed all this content before opening the issue.

woodruffw:
Closes #1997.

This fixes the _core_ of #1997, which is that if a user hash-pinned with a tag SHA instead of a commit SHA we'd (1) miss all of our fast paths, and (2) incorrectly flag it as an impostor since we'd end up doing branch comparisons with the tag SHA rather than a commit SHA.

~~I still need to optimize this more -- the downside to this approach is that we now do at least one (and in the worst case several) GitHub API calls before we ever take the fast paths, which exist primarily to avoid GitHub's slow publishing REST APIs.~~ Okay, this is now roughly as fast as it was before.

### DIFF
--- a/crates/zizmor/src/audit/impostor_commit.rs
+++ b/crates/zizmor/src/audit/impostor_commit.rs
@@ -44,6 +44,18 @@ audit_meta!(
     "commit with no history in referenced repository"
 );
 
+/// An intermediate result type for the impostor check. This is used to
+/// encode our fallthroughs from fastest -> fast -> slow paths for
+/// impostor checks.
+enum IntermediateDetermination {
+    /// Definitely an impostor.
+    NotImpostor,
+    /// Definitely not an impostor.
+    Impostor,
+    /// Not sure yet.
+    Indeterminate,
+}
+
 impl ImpostorCommit {
     async fn named_ref_contains_commit(
         &self,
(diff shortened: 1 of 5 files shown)


crates/zizmor/src/audit/impostor_commit.rs
```diff
@@ -77,6 +79,22 @@ impl ImpostorCommit {
             return Ok(false);
         };
 
+        // `head_ref` might actually been the object ID for a tag,
+        // or even further indirected (e.g. an annotated tag), so we
+        // need to peel it to find the real underlying commit SHA
+        // for the checks below.
+        // TODO: We really shouldn't block the fast paths below
+        // on this, since it's a relatively uncommon case.
```
woodruffw: Noting for myself: I think the trick here is going to be having a queue of `head_ref_candidates`, which we'll try the fast paths on. Then, we'll only add to the queue if the fast paths fail on the initial `head_ref` candidate.

woodruffw: Actually, I guess that's really overkill given that at most we'll only have two candidates.

PR #2017, opened 2026-05-15
Make unpinned-tools version-aware for actions

### LINKED ISSUE #2007: [BUG]: unpinned-tools false positive for aquasecurity/trivy-action v0.36.0
### Pre-submission checks

### zizmor version

1.25.0

### Expected behavior

I would expect zizmor to not flag an `unpinned-tools` finding for the following line:

```yaml
uses: aquasecurity/trivy-action@ed142fd0673e97e23eac54620cfb913e5ce36c25 # v0.36.0
```

### Actual behavior

Zizmor reports `action installs an unpinned external tool: action implictly uses an unpinned latest version`

### Reproduction steps

San [this workflow](https://github.com/martincostello/website/blob/5aecdbb8b0cd956fdcc095cc2bc93d510315808b/.github/workflows/container-scan.yml#L70-L70) with zizmor 1.25.0

### Logs

```plain text

```

### Additional context

Since the recent Trivy incident, the aquasecurity/trivy-action action now pins the tool version if it is not explicitly specified ([action.yaml@v0.36.0](https://github.com/aquasecurity/trivy-action/blob/ed142fd0673e97e23eac54620cfb913e5ce36c25/action.yaml#L98-L101)).

I _used_ to pin Trivy with an explicit version, but after that incident I removed it to effectively slow down release by tying it to the action knowing that the action now pins the underlying trivy version.

woodruffw:
WIP, I'm not very happy with this approach yet (it's slow, plus it's going to confuse offline users).

Fixes #2007.

### DIFF
--- a/crates/zizmor/src/audit/unpinned_tools.rs
+++ b/crates/zizmor/src/audit/unpinned_tools.rs
@@ -1,21 +1,37 @@
+use std::sync::LazyLock;
+
 use github_actions_models::common::{EnvValue, Uses, expr::LoE};
 use subfeature::Subfeature;
 
 use crate::audit::{Audit, AuditError, audit_meta};
 use crate::config::Config;
-use crate::finding::{Confidence, Finding, Severity};
+use crate::finding::{Confidence, Finding, Persona, Severity};
+use crate::github::Client;
 use crate::models::StepBodyCommon;
 use crate::models::uses::RepositoryUsesExt;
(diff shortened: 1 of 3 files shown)


woodruffw:
Notes: 

- I think we can make this a faster path by assuming that the version comment is correct, since `ref-version-mismatch` will enforce that separately.

PR #2029, opened 2026-05-18
feat(shellcheck): add shellcheck wrapper audit rule

guillp:
## Pre-submission checks

Please check these boxes:

- [x] **Mandatory**: This PR corresponds to an issue (if not, please create one first). #1595

- [x] Having read the [AI policy], I hereby disclose the use of an LLM or other AI coding assistant in the creation of this PR. PRs will not be rejected for using AI tools, but *will* be rejected for undisclosed use or use that violates the policy. AI tool was used to generate part of the code, which I then manually reviewed, cleaned from slop and improved.

## Summary

This audit rule runs `shellcheck` on `run:` blocks that are considered to contain shell scripts, as discussed in #1595.

[Shellcheck](https://www.shellcheck.net/) is the de-facto open-source SAST tool for shell scripts. It can detect a large range of security/quality/readability issues.

Such shell scripts are embedded into YAML syntax, so it is difficult to "extract" them then run them trough SAST without using proper tools. `zizmor` is a good candidate for such tools since it is already a dedicated SAST for GitHub Actions.

Implementing the same checks and rules that `shellcheck` already supports into `zizmor`  would be inneficient and a duplication of efforts.

Note that `actionlint` already wraps shellcheck. The reporting of found issues, however, is less readable than those produced using this PR, which points to the exact location of the issue:

```
info[shellcheck]: shellcheck finding in shell run block
  --> crates/zizmor/tests/integration/test-data/shellcheck/positive.yml:17:19
   |
14 |     runs-on: ubuntu-latest
   |     ------- shell implied by runner
...
17 |         run: echo $FOO
   |                   ^^^^ SC2086: Double quote to prevent globbing and word splitting.
   |
   = tip: shellcheck rule reference: https://www.shellcheck.net/wiki/SC2086

1 finding: 1 informational, 0 low, 0 medium, 0 high
```

The main difficulty is to make sure that shellcheck is ran only on shell scripts and nothing else. Since the shell is often implicitly derived from the configured `run-on:` parameter, we cannot always assume that the shell is bash.
So by default, shells that are not explicitly set to bash or bash compatible are ignored and this audit is skipped.
A config option `check-unknown-shell` is available to consider unknown shell as bash, thus triggering the shellcheck rule.
Explicitly setting a `shell: bash` for each job is another good option (or a [default run shell](https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#defaultsrunshell) is a good approach, may be worth another audit?

## Test Plan

Added specific test cases for `shellcheck`:
- if shellcheck is not available in PATH, rule is not triggered
- if shellsheck is available, check that it is correctly triggered and findings are appropriately reported
- 2 test cases with unknown (var-derived) shells, with and without config option `check-unknown-shell`

### DIFF
--- a/crates/zizmor/src/audit/mod.rs
+++ b/crates/zizmor/src/audit/mod.rs
@@ -45,6 +45,7 @@ pub(crate) mod ref_version_mismatch;
 pub(crate) mod secrets_inherit;
 pub(crate) mod secrets_outside_env;
 pub(crate) mod self_hosted_runner;
+pub(crate) mod shellcheck;
 pub(crate) mod stale_action_refs;
 pub(crate) mod superfluous_actions;
 pub(crate) mod template_injection;
--- a/crates/zizmor/src/audit/shellcheck.rs
+++ b/crates/zizmor/src/audit/shellcheck.rs
@@ -198,14 +198,25 @@ impl ShellcheckAudit {
                 vec![]
             });
 
-        tracing::debug!("shellcheck diagnostics: {diagnostics:#?}");
+        let run_location = step.location().primary().with_keys(["run".into()]);
+
+        // Determine the indentation offset of the run: block.
+        // It will be used to calculate the exact offset of findings below.
+        let run_point = run_location
+            .clone()
+            .concretize(step.document())
+            .map_err(Self::err)?
+            .concrete
+            .location
+            .start_point;
 
         diagnostics
             .into_iter()
             .map(|diagnostic| {
+                // find the code fragment inside the inspected shell script
                 let offset = Self::diagnostic_span(run, &diagnostic);
                 let end_offset = if diagnostic.end_line == diagnostic.line {
-                    let line_start = offset - diagnostic.column.saturating_sub(1);
+                    let line_start = offset.saturating_sub(diagnostic.column.saturating_sub(1));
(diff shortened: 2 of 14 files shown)


woodruffw:
@guillp JFYI, I'm unlikely to accept this PR: zizmor intentionally doesn't depend on any external tools, even optional ones, and this use case is already well-handed by tools  like [actionlint](https://github.com/rhysd/actionlint).

(This may change in the distant future, but for the time being this is an intentional restriction as part of keeping zizmor fast and focused on security audits, rather than correctness audits.)

guillp:
@woodruffw: thanks for the feedback and for your work on `zizmor`. I understand the (sane) design rationale. Just wanted to publish in case you or another user would be interested.

To be fair, shellcheck is not only about correctness, it also helps a lot with security.
Actionlint is good but its reporting of shellcheck findings only points to the `run:` block, not to the specific issue in the shell code. That is what motivated me to implement this.

If it suits you better, I can update the PR to disable the rule by default and only enable it by explicit configuration. Otherwise, feel free to close.

guillp:
FYI: shuck is a shellcheck alternative implemented in Rust https://github.com/ewhauser/shuck/
It is able to check run: blocks inside Actions workflows. 
So I am closing this. Thanks!

PR #2037, opened 2026-05-19
fix(ci): correct typw=raw typo in release-docker.yml

shaanmajid:
## Pre-submission checks

Please check these boxes:

      one first).
  - Apologies for skipping this, but there's no option for project internal issues, and filing a full bug report seemed overkill for this. (No option to skip the template either).
      use that violates the policy.

## Summary
The `latest` tag entry in `release-docker.yml` had `typw=raw` instead of `type=raw`. This does not appear to be a functional issue; `docker/metadata-action` ignores unknown attributes and defaults a typeless line to raw, so the `:latest` tag was still emitted correctly when `inputs.latest` was true. Fixing just for consistency, but I this should be a no-op.

## Test Plan
N/A

### DIFF
--- a/.github/workflows/release-docker.yml
+++ b/.github/workflows/release-docker.yml
@@ -135,7 +135,7 @@ jobs:
           images: "${{ env.ZIZMOR_IMAGE }}"
           tags: |
             type=raw,value=${{ github.event.inputs.version }}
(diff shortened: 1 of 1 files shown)


woodruffw (approved):
Oops, thanks!

PR #2042, opened 2026-05-20
Remove pest and replace with a hand-written parser

### LINKED ISSUE #1883: Replace our pest usage with a hand-written parser
Filing this for myself to do. `pest` has served us really well (it's an incredible project), but I think we can get better build- and runtime performance by switching to a hand-written parser for GitHub expressions. The GitHub expression language is also not super easy to write as a pest grammar, as can be seen in the numerous hacks our grammar has accreted over time.

Now that we have KAT tests per #1857 / #1694, doing this with a high degree of confidence should be a lot easier.

woodruffw:
This is mainly an experiment in trying to remove our pest grammar, which both adds to build times and has required various ugly bolt-on hacks.

I've replaced it with a vanilla recursive descent parser. ~~Not sure if it's faster at runtime yet.~~ 

Closes #1883.


woodruffw:
Unfortunately this seems like mostly a wash in terms of runtime performance. But the build-time improvement is worth it, I think.

PR #2053, opened 2026-05-23
Add another cache control heuristic

### LINKED ISSUE #2050: Feature: cache-poisoning (finer grained)
### Pre-submission checks

I've done my best to read all tickets related to cache poisoning, I hope I didn't miss any.

There are some cases where caches can be acceptable, while other cases make cache poisoning much easier. With Zizmor's cache-poisining rule, it's currently all or nothing:
- either the rule is enabled and prevents all caches everywhere
- or the rule is disabled (via comments or globally) and caches aren't checked at all

I'd like to be able to use things like:

```yaml
cache: ${{ !(github.event_name == 'push' && github.ref_type == 'tag') }}
```
which, I believe would ensure that tag builds use no cache at all (but less secure builds may use cache).

This means: I'd like zizmor to consider it an error if I **don't** specify a `cache:` value but be more lenient on the value itself (as long as it's not `true`)

Optionally, it would be interesting to be able to provide a list that limits the valid cache values, e.g. to the value `${{ !(github.event_name == 'push' && github.ref_type == 'tag') }}` (keeping in mind that it's most probably different from workflow to workflow)

I'm perfectly open to a decision that partial caching is still much too risky, but I wanted to open the discussion.

In other words, currently, the rules states 2 things: there must be a cache value, and it must be false. I like the first part and would like to keep it, and I think there is a world where the second part is more lenient while still being secure.

### Additional context

_No response_

woodruffw:
This ended up being very large due to refactoring (mostly moving `BinExpr` into its own proper type and changing how origin tests are represented), but the core change is only a hundred lines or so.

TL;DR: we can now detect (with acceptable generality) patterns like `github.event_name == 'push' && github.ref_type == 'tag'` (and their negation). Needs more tests still.

Fixes #2050.

### DIFF
--- a/crates/github-actions-expressions/src/lib.rs
+++ b/crates/github-actions-expressions/src/lib.rs
@@ -10,7 +10,7 @@ use crate::{
     context::Context,
     identifier::Identifier,
     literal::Literal,
-    op::{BinOp, UnOp},
+    op::{BinExpr, BinOp, UnOp},
 };
 
 pub mod call;
@@ -73,7 +73,11 @@ impl<'a> Origin<'a> {
 /// was parsed from, with no surrounding whitespace. A parenthesized expression
 /// includes its parentheses, so every span is a balanced, self-contained slice
 /// of the source.
-#[derive(Debug, PartialEq)]
+///
+/// NOTE: `SpannedExpr` has a manual [`PartialEq`] implementation that only considers
+/// the underlying expression, not the span. In opther words, two `SpannedExpr` instances
+/// are equal if their ASTs are equal, even if their source spans are not.
+#[derive(Debug)]
 pub struct SpannedExpr<'src> {
     /// The expression's source origin.
     pub origin: Origin<'src>,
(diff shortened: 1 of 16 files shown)


crates/github-actions-expressions/src/lib.rs
```diff
@@ -402,6 +407,38 @@ impl<'src> Expr<'src> {
...
     }
+
+    /// Returns whether this expression 'commutatively matches' the given expression.
+    ///
+    /// For most expressions, this is the same as equivalence (i.e. `==`).
+    /// For binary expressions that are also commutative, this takes commutivity into account.
+    /// For example, `a == b` is considered to match `b == a`, but `a > b` is not
+    /// considered to match `b > a`. This check is recusive, i.e. two nested binary expressions
+    /// will be fully checked for commutative equivalence.
+    pub fn commutative_matches(&self, other: &Expr<'src>) -> bool {
```
woodruffw, line 418: This is one of the key changes: we now have this helper for comparing `BinExpr`s for commutative equivalence. We don't yet do anything more special here, like De Morgan's laws for equivalence between distinct forms.

PR #2055, opened 2026-05-23
dependabot-cooldown: honor config when fixing

### LINKED ISSUE #2047: [BUG]: rules.dependabot-cooldown.config.days ignored
### Pre-submission checks

### zizmor version

1.25.2

### Expected behavior

When configuring `rules.dependabot-cooldown.config.days`, I expect that value to be the minumum value that the audit and autofix will check for and apply.  

### Actual behavior

`rules.dependabot-cooldown.config.days` seems to be ignored, and 7 days is the default value always applied.

### Reproduction steps

Setting `rules.dependabot-cooldown.config.days` to a value such as 14 has no effect on the value that audit checks or autofix applies.  Seems to be hardcoded to 7.

Example zizmor config
```yaml
# .github/zizmor.yml
rules:
  dependabot-cooldown:
    config:
      days: 14
```

This file passes the audit
```yaml
# .github/dependabot.yml
version: 2

updates:
  - package-ecosystem: pip
    directory: /
    schedule:
      interval: daily
    cooldown:
      default-days: 7
```

If we provide a Dependabot config without a value set and enable autofix and the same Zizmor config from above
```yaml
# .github/dependabot.yml
version: 2

updates:
  - package-ecosystem: pip
    directory: /
    schedule:
      interval: daily
```

...it results in a file with a 7 day cooldown (rather than the expected/configured 14)

```yaml
# .github/dependabot.yml
version: 2

updates:
  - package-ecosystem: pip
    directory: /
    schedule:
      interval: daily
    cooldown:
      default-days: 7
```

### Logs

```plain text

```

### Additional context

_No response_

woodruffw:
Previously we only honored the config when auditing, not auto-fixing.

Fixes #2047. Closes #2048.

### DIFF
--- a/crates/zizmor/src/audit/dependabot_cooldown.rs
+++ b/crates/zizmor/src/audit/dependabot_cooldown.rs
@@ -34,9 +34,10 @@ impl DependabotCooldown {
     /// Creates a fix that increases an insufficient default-days value
     fn create_increase_default_days_fix<'doc>(
         update: crate::models::dependabot::Update<'doc>,
+        minimum_days: u64,
     ) -> Fix<'doc> {
         Fix {
-            title: "increase default-days to 7".to_string(),
+            title: format!("increase default-days to {minimum_days}"),
             key: update.location().key,
             disposition: FixDisposition::Safe,
             patches: vec![Patch {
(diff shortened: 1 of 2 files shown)


woodruffw:
(This reminds me that we still need a better way to integration-test fixes than the current unit-level approach.)

PR #2060, opened 2026-05-24
audit: add `unpinned-packages` rule

### LINKED ISSUE #2057: Feature: Rule to ban install of packages outside the context of a lockfile (`npm install pkg`, `gem install pkg`)
### Pre-submission checks

I see a decent number of CI setups wherein a specific package, gem, etc. is installed and used by a workflow.

For example, a job to release a gem to RubyGems.org:

```yml
name: Release gems

jobs:
  release:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v6
      - uses: ruby/setup-ruby@v1
      - uses: rubygems/configure-rubygems-credentials@v2
      - name: Publish to RubyGems
        run: |
          gem install gem-release # <- this installs the `gem-release` gem's latest version, and the latest versions of gems it depends on
          rake release
``` 

This opens the workflow up to a few different attacks:

- Compromised packages are more problematic, as their newest version always gets installed automatically when the workflow runs, which can happen dozens or hundreds of times a day in large projects.
- Determining if a compromise has occurred after such an attack can be considerably more difficult without a lockfile, since the job logs don't necessarily include specific versions of every package installed this way.
- SHA hashes - if your package manager/lockfile supports them - would not be enforced as there is no lockfile to check against.

Even if `gem install gem-release:1.2.3` were used, the gems it depends on are still loosely-defined and install the newest versions (that are allowed by the gem's ranges, at least), and SHA hashes are still not enforced if RubyGems.org itself were compromised.

I'd like a zizmor rule that would catch usages of commands like

- `gem install gem-name`
- `npm install package-name`
- `npx -y package-name`
- `npx --yes package-name`
- `pip install package-name`
- `composer require package-name`

Because they avoid going through a lockfile, they're always pulling the latest release of a package (and all of its dependencies).

The concern I have is just that the pattern makes the likelihood of compromise _a lot_ higher if any given package or its dependents were compromised. So it'd be good to warn on all of these, and suggest the user add them to the Gemfile/package.json/etc. if possible.

### Additional context

I saw #2044, #1607, and #1405 which are similar, but none of them seem to be what I'm suggesting here specifically.

It could be called `unpinned-packages`, or something like that?

thebigbone:
## Pre-submission checks

Please check these boxes:

      one first).

      use that violates the policy.

## Summary

Detects direct package installs outside lockfile context (`npm install <pkg>`, `gem install <pkg>`, `pip install <pkg>`, `npx -y`, `composer require <pkg>`) in workflow `run:` steps.

closes #2057

### DIFF
--- a/crates/zizmor/src/audit/mod.rs
+++ b/crates/zizmor/src/audit/mod.rs
@@ -51,6 +51,7 @@ pub(crate) mod template_injection;
 pub(crate) mod typosquat_uses;
 pub(crate) mod undocumented_permissions;
 pub(crate) mod unpinned_images;
+pub(crate) mod unpinned_packages;
 pub(crate) mod unpinned_tools;
 pub(crate) mod unpinned_uses;
 pub(crate) mod unredacted_secrets;
--- a/crates/zizmor/src/audit/unpinned_packages.rs
+++ b/crates/zizmor/src/audit/unpinned_packages.rs
@@ -1,0 +1,324 @@
+use crate::{
+    audit::{Audit, AuditError, AuditLoadError, audit_meta},
+    config::Config,
+    finding::{Confidence, Finding, Severity},
+    models::{StepBodyCommon, StepCommon, action::CompositeStep, workflow::Step},
+    state::AuditState,
+    utils::once::static_regex,
+};
+
+audit_meta!(
+    UnpinnedPackages,
+    "unpinned-packages",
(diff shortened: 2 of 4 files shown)


woodruffw:
This appears to overlap with #2061, but doesn't follow the project's idioms correctly (we use tree-sitter for semantic-aware shell audits, wherever possible).

PR #2063, opened 2026-05-25
Document node `cache-poisoning` remediation

### LINKED ISSUE #2062: Feature: update docs for #1153
### Pre-submission checks

Would be nice to update docs after #1153 so the right remediation is easier to find.

Update the docs

### Additional context

_No response_

mxr:
## Pre-submission checks

Please check these boxes:

      one first).

      use that violates the policy.

## Summary

Document node `cache-poisoning` remediation. Resolves #2062

## Test Plan

CI

### DIFF
--- a/docs/audits.md
+++ b/docs/audits.md
@@ -303,7 +303,8 @@ intended to publish build artifacts:
 * Disable cache-aware actions with an `#!yaml if:` condition based on the trigger at
   the step level, *or*
 * Set an action-specific input to disable cache restoration when appropriate,
-  such as `lookup-only` in @Swatinem/rust-cache.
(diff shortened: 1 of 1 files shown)


woodruffw:
My main hesitancy with doing this is that there are a _lot_ of cache-aware actions, and listing them all in prose is going to make these docs very hard to read. I think a cleaner approach here would be to add a table, something like:

```markdown
| Action | Input |
| ------ | ----- |
| @Swatinem/rust-cache | `lookup-only: true` |
| @actions/setup-node | `package-manager-cache: false` |
```

I'll do this in a distinct PR so you don't need to cycle through my nitpicks 🙂

PR #2065, opened 2026-05-25
Skip statically-false `if:` steps in audits (#2059)

### LINKED ISSUE #2059: unpinned-tools: false positive on steps with if: false (allowlist-registration pattern)
### Summary

The `unpinned-tools` audit flags steps that are syntactically `if: false` — i.e., guaranteed never to execute. In repos that use a composite action as a declarative allowlist registry (where every `- uses:` entry is `if: false` purely so dependabot tracks the SHA for inclusion in an approved-actions list), this produces a false positive: the audit's premise ("when this step runs, it'll install an unpinned tool") doesn't hold for steps that don't run.

### Reproducer

Minimal composite action:

```yaml
runs:
  using: "composite"
  steps:
    - uses: 1Password/load-secrets-action@<sha>  # v4.0.0
      if: false
```

zizmor v1.25.2 raises `warning[unpinned-tools]` here, even though the step has no runtime effect.

Real-world example: https://github.com/apache/infrastructure-actions/blob/main/.github/actions/for-dependabot-triggered-reviews/action.yml — 105 `if: false` entries serving as the approved-actions allowlist; one currently triggers `unpinned-tools`.

### Suggested handling

Skip `unpinned-tools` (and arguably similar runtime-behavior audits — `bot-conditions`, `cache-poisoning`, etc.) on steps where `if` is a statically-falsey literal: `if: false`, `if: ${{ false }}`. Detecting more sophisticated never-runs conditions (e.g. `if: ${{ github.actor == 'never-this-user' }}`) is harder and out of scope; literal-false detection alone covers the common allowlist-registry pattern.

This wouldn't weaken coverage of any step that could actually run.

### Workaround in the meantime

Inline `# zizmor: ignore[unpinned-tools]` per offending step — works, but each new entry that triggers the audit needs its own ignore, which scales poorly across allowlists of 100+ entries.

Happy to discuss / send a PR if there's interest.

potiuk:
Steps whose `if:` condition is provably false at static-analysis time (e.g. `if: false`, `if: ${{ false }}`, `if: ${{ true && false }}`) cannot execute, so any "what happens when this step runs" audit (unpinned-tools, bot-conditions, cache-poisoning, etc.) would always be a false positive on them.

This is filtered at the iterator layer (`Steps`, `CompositeSteps`), so all step-level audits — and `conditions()`-based audits — inherit the behavior. Non-Boolean falsy values (`${{ '' }}`, `${{ 0 }}`) are kept out of scope; only conditions that const-evaluate to `Boolean(false)` are skipped.

Fixes #2059.

## Pre-submission checks

Please check these boxes:

      one first).

      use that violates the policy.

## Summary

Removes the "static-analysis-time" `if: false` conditioned actions from auditing.

## Test Plan

Unit tests were updated as part of the change.

### DIFF
--- a/crates/zizmor/src/models.rs
+++ b/crates/zizmor/src/models.rs
@@ -1,7 +1,7 @@
 //! Enriching/context-bearing wrappers over GitHub Actions models
 //! from [`github_actions_models`].
 
-use github_actions_expressions::context;
+use github_actions_expressions::{Evaluation, Expr, context};
 use github_actions_models::common;
 use github_actions_models::common::Env;
 use github_actions_models::common::expr::LoE;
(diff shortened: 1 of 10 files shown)


woodruffw:
Thanks a ton @potiuk! This looks great to me overall; I'm just going to push up some small nitpicks.

crates/zizmor/src/models.rs
```diff
@@ -1,14 +1,40 @@
...
+/// This recognizes:
+///
+/// 1. The literal `if: false` (which deserializes as [`common::If::Bool(false)`]).
+/// 2. Bare or fenced expressions that const-evaluate to a falsy boolean,
+///    e.g. `if: ${{ false }}`, `if: false && something`.
+///
+/// More elaborate "never runs" conditions (e.g. `if: github.actor == 'no-one'`)
+/// are intentionally out of scope; only conditions that we can prove falsy
+/// without any runtime context are recognized.
+pub(crate) fn if_is_statically_false(cond: &common::If) -> bool {
```
woodruffw, line 26: Nit: Could probably be an instance method on `If`.

woodruffw: Oops, just realized that won't work because of cross-crate boundaries. Never mind...

PR #2085, opened 2026-05-31
Detect falsy true arms in actions pseudo-ternaries

### LINKED ISSUE #746: Feature: wrong value in ternary pattern
### Pre-submission checks

In GitHub Actions expressions, using a ternary pattern like this:

```yaml
env:
  VARIABLE: ${{ condition && '' || 'value' }}
```
does not work as intended when the true value is an empty string.
Because '' is treated as false, the expression returns 'value' even when the condition is true.

Can you confirm this is the case and if so, then please add as an audit.

The ternary pattern is inverted, no problem.

### Additional context

_No response_

terror:
## Pre-submission checks

Please check these boxes:

      one first).
      use that violates the policy.

## Summary

Resolves https://github.com/zizmorcore/zizmor/issues/746

This diff adds an `obfuscation` audit check for GitHub Actions expressions that use the common `condition && value || fallback` pseudo-ternary pattern when `value` statically evaluates as falsy. In those cases, GitHub Actions short-circuit semantics return the fallback even when the condition is true, so expressions like `foo && '' || 'bar'` do not behave like ternaries.

The check reports the falsy true arm directly and handles chained `||` expressions so later pseudo-ternaries are not missed. It covers falsy constants such as empty strings, `0`, `false`, `null`, and constant-foldable falsy expressions, while avoiding fine forms like `!condition && 'value' || ''` and non-falsy true arms.

**n.b.** This does _not_ add an auto-fix yet. A reasonable follow-up would be to rewrite eligible patterns to the official `case(...)` form, for example converting `condition && true_value || fallback` into
`case(condition, true_value, fallback)`. I'm open to getting this in here, but it may make the diff larger than ideal 🤔 

## Test Plan

`cargo test -p zizmor`

### DIFF
--- a/docs/audits.md
+++ b/docs/audits.md
@@ -2445,6 +2445,72 @@ Other resources:
                 if: contains(fromJSON('["refs/heads/main", "refs/heads/develop"]'), github.ref)
         ```
 
+## `unsound-ternary`
+
+| Type     | Examples              | Introduced in | Works offline | Auto-fixes available | Configurable |
+|----------|-----------------------|---------------|---------------|--------------------|--------------|
+| Workflow, Action | [issue-746-repro.yml] | v1.26.0      | ✅            | ❌                 | ❌           |
+
+[issue-746-repro.yml]: https://github.com/zizmorcore/zizmor/blob/main/crates/zizmor/tests/integration/test-data/unsound-ternary/issue-746-repro.yml
+
+Detects GitHub Actions expressions that use `&&` and `||` as a pseudo-ternary
+when the "true" value can evaluate to a falsy value.
+
+It's common to imitate the behavior of a ternary in GitHub Actions expressions
+like this:
+
+```yaml
+${{ condition && value || fallback }}
+```
+
+This only behaves like a ternary when `value` is truthy. If `value` is falsy,
+the expression falls through to `fallback` even when `condition` is true. For
+example, the following expression evaluates to `bar` when `foo` is truthy:
+
+```yaml
+${{ foo && '' || 'bar' }}
+```
+
+This audit detects pseudo-ternaries where the true arm statically evaluates to
+a falsy value, including `#!yaml ''`, `#!yaml 0`, `#!yaml false`, and
+`#!yaml null`.
+
+### Remediation
+
+Use the official `case(...)` function when a value may be falsy:
+
+```yaml
+${{ case(condition, value, fallback) }}
+```
+
+Alternatively, rewrite the expression so that the true arm cannot be falsy.
+
+!!! example
+
+    === "Before :warning:"
+
+        ```yaml title="unsound-ternary.yml" hl_lines="7"
+        jobs:
+          build:
+            runs-on: ubuntu-latest
+            steps:
+              - run: echo "value=${VALUE}"
+                env:
+                  VALUE: ${{ foo && '' || 'bar' }}
+        ```
+
+    === "After :white_check_mark:"
+
+        ```yaml title="unsound-ternary.yml" hl_lines="7"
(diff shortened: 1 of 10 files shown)


woodruffw:
Thanks for opening this @terror. I'm curious what you think about doing this in a new audit (something like `unsound-ternary`?) -- it's arguably not obfuscation per se, but an expression footgun like the other `unsound-*` audits.

terror:
> Thanks for opening this @terror. I'm curious what you think about doing this in a new audit (something like `unsound-ternary`?) -- it's arguably not obfuscation per se, but an expression footgun like the other `unsound-*` audits.

I think that makes sense! I put it under obfuscation mostly because of the earlier issue discussion and because the existing audit already walks fenced expressions, but I agree that this is less about obfuscation and more about confusing semantics. 

I'll look to scope this into a new `unsound-ternary` audit.

docs/audits.md
```diff
@@ -2445,6 +2445,78 @@ Other resources:
...
+|----------|-----------------------|---------------|---------------|--------------------|--------------|
+| Workflow, Action | [issue-746-repro.yml] | v1.26.0      | ✅            | ❌                 | ❌           |
+
+[issue-746-repro.yml]: https://github.com/zizmorcore/zizmor/blob/main/crates/zizmor/tests/integration/test-data/unsound-ternary/issue-746-repro.yml
+
+Detects GitHub Actions expressions that use `&&` and `||` as a pseudo-ternary
+when the "true" value can evaluate to a falsy value.
+
+GitHub Actions expressions do not have a `?:` ternary operator. A common
+substitute is to write an expression like this:
```
woodruffw: Nit: we mention `case(...)` in the remediation below and it's effectively GitHub's official ternary option, so maybe a tweak:


```suggestion
It's common to imitate the behavior of a ternary in GitHub Actions expressions like this:
```

docs/audits.md
```diff
@@ -2445,6 +2445,78 @@ Other resources:
...
+
+This audit detects pseudo-ternaries where the true arm statically evaluates to
+a falsy value, including `#!yaml ''`, `#!yaml 0`, `#!yaml false`, `#!yaml null`,
+and constant-foldable expressions that produce a falsy value.
+
+!!! note
+
+    Fully static expressions, such as `#!yaml ${{ true && '' || 'bar' }}`, are
+    handled by the [`obfuscation`](#obfuscation) audit's constant-reduction
+    finding instead.
```
woodruffw: I think this is safe to remove 🙂

docs/audits.md
```diff
@@ -2445,6 +2445,78 @@ Other resources:
...
+the expression falls through to `fallback` even when `condition` is true. For
+example, the following expression evaluates to `bar` when `foo` is truthy:
+
+```yaml
+${{ foo && '' || 'bar' }}
+```
+
+This audit detects pseudo-ternaries where the true arm statically evaluates to
+a falsy value, including `#!yaml ''`, `#!yaml 0`, `#!yaml false`, `#!yaml null`,
+and constant-foldable expressions that produce a falsy value.
```
woodruffw: Nit: shorten since "statically evaluates" is already mentioned at the head of the sentence:

```suggestion
a falsy value, including `#!yaml ''`, `#!yaml 0`, `#!yaml false`, and `#!yaml null`.
```

woodruffw (commented):
Thanks for working on this @terror! I did a quick skim of the docs, just a few small nitpicks. I'll do a full review of the audit's logic tomorrow as well.

woodruffw (approved):

woodruffw:
Thanks a ton @terror!

Issue #2046, opened 2026-05-21
Feature: Cannot disambiguate filename when multiple files have the same name


AntoineSebert:
### Pre-submission checks

- [x] I am **not** reporting a bug (crash, false positive/negative, etc). These must be filed via the bug report template.
- [x] I have looked through both the [open](https://github.com/zizmorcore/zizmor/issues?q=is%3Aissue+is%3Aopen+) and [closed](https://github.com/zizmorcore/zizmor/issues?q=is%3Aissue%20state%3Aclosed) issues for a duplicate request.

### What's the problem this feature will solve?

Sample Github Action repository:
```
.
├── sample-folder-0
│   ├── sample-action-0
│   │   ├── action.yml
│   │   └── README.md
│   └── sample-action-1
│       ├── action.yml
│       └── README.md
└── sample-folder-1
    ├── sample-action-2
    │   ├── action.yml
    │   └── README.md
    └── sample-action-3
        ├── action.yml
        └── README.md
```

Sample `zizmor` worklflow run (taken from actual run):
```sh
[17](https://github.com/my-org/my-repo/actions/runs/26211415953/job/77122969824?pr=122#step:4:119)
 INFO audit: zizmor: 🌈 completed ./sample-folder-0/sample-action-0/action.yml
 INFO audit: zizmor: 🌈 completed ./sample-folder-0/sample-action-1/action.yml
Warning: action.yaml:63: action installs an unpinned external tool: action implictly uses an unpinned latest version
Error: Process completed with exit code 13.
```

The location of the issue is `action.yaml:63`, however this is not sufficient to pinpoint the file it originates from.
We can filter out audits marked as "completed" (in our example, `sample-folder-0/sample-action-0` and `sample-folder-0/sample-action-1`), but we cannot determine if it lies in `sample-folder-1/sample-action-2` or `sample-folder-1/sample-action-3`.

This is especially an issue in repositories that contain many more `action.yml`.

### Describe the solution you'd like

Sample `zizmor` worklflow run where the audit filepath is shown:
```sh
[17](https://github.com/my-org/my-repo/actions/runs/26211415953/job/77122969824?pr=122#step:4:119)
 INFO audit: zizmor: 🌈 completed ./sample-folder-0/sample-action-0/action.yml
 INFO audit: zizmor: 🌈 completed ./sample-folder-0/sample-action-1/action.yml
[...]

PR #1748, opened 2026-03-11
fix(github-format): preserve path

reubenwong97:
## Pre-submission checks

Please check these boxes:

- [x] **Mandatory**: This PR corresponds to an issue #1745 .

- [ ] Having read the [AI policy], I hereby disclose the use of an LLM or other
      AI coding assistant in the creation of this PR. PRs will not be rejected
      for using AI tools, but *will* be rejected for undisclosed use or
      use that violates the policy.

[AI policy]: https://github.com/zizmorcore/.github/blob/main/AI_POLICY.md

If a checkbox is not applicable, you can leave it unchecked.

## Summary

For the github format, uses the presentation path with prefix stripping instead of the sarif path.

## Test Plan

Added snapshot integration tests for the scenarios described in the issue.

Closes #1745

Issue #1745, opened 2026-03-10
[BUG]: Part of the path is missing when displayed in `GitHub` format


Danil42Russia:
### Pre-submission checks

- [x] I am **not** filing a feature request. These should be filed via the feature request form instead.
- [x] I have checked the [Troubleshooting Guide](https://docs.zizmor.sh/troubleshooting) for my problem.
- [x] I have looked through both the [open](https://github.com/zizmorcore/zizmor/issues?q=is%3Aissue+is%3Aopen+) and [closed](https://github.com/zizmorcore/zizmor/issues?q=is%3Aissue%20state%3Aclosed) issues for a duplicate report.

### zizmor version

1.23.1

### Expected behavior

```
::error file=.github/workflows/build.yml,line=17,title=excessive-permissions::build.yml:17: overly broad permissions: contents: write is overly broad at the workflow level
::warning file=.github/workflows/build.yml,line=287,title=archived-uses::build.yml:287: action or reusable workflow from archived repository: repository is archived
```

### Actual behavior

```
::error file=workflows/build.yml,line=17,title=excessive-permissions::build.yml:17: overly broad permissions: contents: write is overly broad at the workflow level
::warning file=workflows/build.yml,line=287,title=archived-uses::build.yml:287: action or reusable workflow from archived repository: repository is archived
```

### Reproduction steps

```bash
zizmor .github --format=github
```

### Logs

```plain text

```

### Additional context

Discovered when adding `Zizmor` to private repo, via `zizmorcore/zizmor-action`

Because `.github` disappears from the path, `GitHub Annotations` cannot be attached to changes and therefore are not displayed :(

woodruffw:
Hi @Danil42Russia, thanks for the report. I suspect that's happening because you're running zizmor against a subdirectory, rather than your repo root -- could you try `zizmor .` and see if the paths in the annotations are correct?

(This is a bug regardless, but that should be the fast fix.)

Danil42Russia:
Yes, this happens if I specify the `.github` subdirectory instead of the project root `.`:

```bash
zizmor . --format=github
zizmor .github --format=github
```

```
🌈 zizmor v1.23.1
 INFO audit: zizmor: 🌈 completed ./.github/workflows/ci.yml
::warning file=.github/workflows/ci.yml,line=9,title=excessive-permissions::ci.yml:9: overly broad permissions: default permissions used due to no permissions: block
🌈 zizmor v1.23.1
[...]

PR #2095, opened 2026-06-02
Add InputGroup::root

woodruffw:
Every input group now tracks its "root", i.e. the repository root it belongs to (if it has one). We then use this to opportunistically provide a better "SARIF path" for both the SARIF and GH annotations output formats.

~~WIP, not plumbed yet. The idea here is to address a class of recurring problems where users run `zizmor .github` or similar and get relative paths in their outputs that they don't expect (or that their tools don't expect).~~

TODO:

- [x] Plumb this into the GitHub annotations format as well
- [x] Clean up more

Context: #1745 #2046 #2113 and probably several others. Augments the initial fix in #1748.

### DIFF
--- a/crates/zizmor/src/registry/input.rs
+++ b/crates/zizmor/src/registry/input.rs
@@ -447,6 +447,7 @@ impl InputGroup {
                 break;
             }
 
+            // TODO: Handle worktrees?
             tracing::trace!("checking if {candidate} is a Git repository root");
             if candidate.join(".git").is_dir() {
                 return Some(candidate);
--- a/crates/zizmor/src/registry/input.rs
+++ b/crates/zizmor/src/registry/input.rs
@@ -1,10 +1,11 @@
 //! Input registry and associated types.
 
 use std::{
-    collections::{BTreeMap, btree_map},
+    collections::{BTreeMap, HashSet, btree_map},
     io::Read as _,
     path::PathBuf,
     str::FromStr as _,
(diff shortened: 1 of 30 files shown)


crates/zizmor/src/registry/input.rs
```diff
@@ -368,6 +380,42 @@ impl InputGroup {
         Ok(())
     }
 
+    /// Given a path to an input file, attempt to discover the Git repository root that it belongs
+    /// to, if any.
+    ///
+    /// This is a rough approximation of what `git rev-parse --show-toplevel` does.
+    ///
+    /// Returns `None` if the path is not within a Git repository or if the root can't be determined for any reason.
+    pub(crate) fn discover_root(path: &Utf8Path) -> Option<Utf8PathBuf> {
```
woodruffw, line 420: I think we'll need to support `GIT_CEILING_DIRECTORIES` in this.

crates/zizmor/src/registry/input.rs
```diff
@@ -368,6 +380,42 @@ impl InputGroup {
...
+        let Some(mut candidate) = if path.is_file() {
+            path.parent()?.to_path_buf()
```
woodruffw: This returns `Some("")` for `foo.yml`, which won't do what we expect.

woodruffw:
I don't love this, but it's better than what we had before.

Issue #2097, opened 2026-06-03
[BUG]: unpinned version false positive, when using matrix builds (GitHub Actions)


staabm:
### Pre-submission checks

- [x] I am **not** filing a feature request. These should be filed via the feature request form instead.
- [x] I have checked the [Troubleshooting Guide](https://docs.zizmor.sh/troubleshooting) for my problem.
- [x] I have looked through both the [open](https://github.com/zizmorcore/zizmor/issues?q=is%3Aissue+is%3Aopen+) and [closed](https://github.com/zizmorcore/zizmor/issues?q=is%3Aissue%20state%3Aclosed) issues for a duplicate report.
- [x] I affirm that I have followed the [AI policy](https://github.com/zizmorcore/.github/blob/main/AI_POLICY.md) and that this report constitutes my own words and understanding.

### zizmor version

1.25.2

### Expected behavior

we are feeding a pinned docker image identifier via matrix strategy into a github actions `jobs.services` definition.
since we do not want all our jobs to take the perf hit of setting up the docker image, we set this image to `''` for most jobs.

### Actual behavior

zizmor doesn't like our job config

<img width="1580" height="744" alt="Image" src="https://github.com/user-attachments/assets/fbc01cf6-bf50-4624-9d67-5d8ac2b9d707" />

### Reproduction steps

have a github action job definition like

```
jobs:
  integration-tests:
    name: "Integration - ${{ matrix.name }}"
    runs-on: "ubuntu-latest"

    services:
      database:
        # see https://hub.docker.com/_/mysql
        image: "${{ matrix.mysql || '' }}"
        env:
          MYSQL_ROOT_PASSWORD: root
        ports:
          - 3306:3306
        options: '--health-cmd="mysqladmin ping" --health-interval=10s --health-timeout=5s --health-retries=3'

    strategy:
      fail-fast: false
      matrix:
        include:
         # a job which depends on mysql and therefore defines a pinned version for the service
          - php-version: 8.5
            name: phpstan-dba tests
            mysql: 'mysql:8.4.9@sha256:c36050afdca850f23cef85703f84c7531a5ae155a11b5ee1c60acb09937c4084' # 8.4.9

         # ... some more jobs which don't define the mysql value
          - php-version: 8.5
            name: phpstan-dba tests
```

### Logs

```plain text

```

### Additional context

the above construct was implemented in https://github.com/phpstan/phpstan/pull/14760/files and works like expected (from a github actions point of view).

[...]

PR #2098, opened 2026-06-03
unpinned-uses: handle matrix indirections more generally

woodruffw:
This performs an overdue refactor of how we classify Docker images in `unpinned-uses`, fixing #2097 in the process.


staabm:
Thank you 🙏

PR #2108, opened 2026-06-08
fix(config): treat empty or comment-only config files as valid

### LINKED ISSUE #2111: [BUG]: Empty or comment-only zizmor.yml causes fatal configuration error
### Pre-submission checks

### zizmor version

1.25.2

### Expected behavior

A comment-only file should be treated as a valid configuration equivalent to `RawConfig::default()` - i.e., no rules overridden, audit proceeds normally.

Other edge cases to consider:
* An empty file,
*  a file only containing `{}`

### Actual behavior

An empty or comment-only `.github/zizmor.yml` config file causes a fatal error instead of being treated as a valid (no-op) configuration:

```
   fatal: no audit was performed
   error: configuration error in .github/workflows
     = help: invalid configuration syntax
     = help: missing field `rules` at line 3 column 1
```

While the error message is easily understood, it came as a surprise while I was progressively commenting out rules while enabling zizmor.

### Reproduction steps

```bash
# Create an empty config file
touch .github/zizmor.yml
zizmor .
# => fatal: no audit was performed

echo "# Only Comments" > .github/zizmor.yml
zizmor .
# => fatal: no audit was performed
```

### Logs

```plain text

```

### Additional context

_No response_

serval-frenchie:
## Pre-submission checks

Please check these boxes:

      one first).

      use that violates the policy.

## Summary

An empty or comment-only `.github/zizmor.yml` currently produces a fatal error:

```
fatal: no audit was performed
error: configuration error in .github/workflows
  = help: invalid configuration syntax
  = help: missing field `rules` at line 3 column 1
```

This fixes that & allows for empty or comment-only config files.

Two changes to `RawConfig`:

1. **`#[serde(default)]` on `rules`** — a mapping without a `rules` key (e.g. `{}`) deserializes to an empty `HashMap` instead of erroring.
2. **Null-document guard in `RawConfig::load`** — an empty file or a comments-only file produces a YAML null document; we now return `RawConfig::default()` instead of attempting to deserialize null into a struct.

Fixes #2111 

## Test Plan

- 3 unit tests in `config/mod.rs`: empty file, comment-only file, empty mapping (`{}`)
- 2 integration tests with fixture config files covering the same cases end-to-end

🤖 Generated with [Claude Code](https://claude.com/claude-code)


serval-frenchie:
Reading the PR backlog, this seems potentially related to #1010

woodruffw:
Hi @serval-frenchie, thanks for opening a PR. PRs must use our [PR template](https://github.com/zizmorcore/.github/blob/main/.github/PULL_REQUEST_TEMPLATE.md) and follow our [AI policy](https://github.com/zizmorcore/.github/blob/main/AI_POLICY.md). Once both of those are true, this PR will be reviewable.

serval-frenchie:
Thanks for the quick feedback @woodruffw and apologies for the oversight - #1010 is somewhat related, but not precisely. I opened issue #2111 for this specific case.

crates/zizmor/tests/integration/config.rs
woodruffw, line 1: I don't think we need these integration test changes, the unit tests covering these config scenarios are enough.

crates/zizmor/src/config/mod.rs
```diff
@@ -795,7 +803,29 @@ impl Config {
...
+        let config = RawConfig::load(contents).expect("comment-only config should be valid");
+        assert!(config.rules.is_empty());
+    }
+
+    #[test]
+    fn test_load_rules_missing_key() {
+        // A mapping with no `rules` key should parse as a default config.
+        let config = RawConfig::load("{}").expect("empty mapping should be valid");
+        assert!(config.rules.is_empty());
+    }
```
woodruffw, line 824: Nit: this should probably be one `#[test]` written in a table-driven style instead.

serval-frenchie: much nicer! thanks for the suggestion.

crates/zizmor/src/config/mod.rs
```diff
@@ -194,12 +194,20 @@ pub(crate) struct AuditRuleConfig {
...
     rules: HashMap<String, AuditRuleConfig>,
 }
 
 impl RawConfig {
     fn load(contents: &str) -> Result<Self, ConfigErrorInner> {
-        yaml_serde::from_str(contents).map_err(ConfigErrorInner::Syntax)
+        // An empty or comment-only config file produces a YAML null document.
+        // Treat that the same as an empty mapping (no rule overrides).
+        let value: yaml_serde::Value =
+            yaml_serde::from_str(contents).map_err(ConfigErrorInner::Syntax)?;
```
woodruffw: I think we need to do something different here -- parsing into a `Value` means that we lose line context on errors when transforming the value into our actual config type. I think we should probably parse into an `Option<RawConfig>` instead?

serval-frenchie: Thanks for the feedback. I'm still pretty rusty when it comes to Rust 🥁 💥 , definitely appreciate the tips!

PR #1000, opened 2025-07-01
refactor: simplify and correct Remove op

woodruffw:
WIP; this is going to require some re-thinking within `yamlpath`.

Follows #999.

PR #2121, opened 2026-06-14
fix: make Remove patch op sound for flow collections and empty parents

### LINKED ISSUE #2122: [BUG]: Remove patch op corrupts flow collections and leaves dangling parents in fix mode
## Pre-submission checks

> Note: this report was drafted with AI assistance at the contributor's direction. The AI-policy affirmation box is intentionally left unchecked pending the human contributor's own review and disclosure on the accompanying PR.

## zizmor version

1.25.2 (main)

## Expected behavior

The `Remove` patch op (used by fix mode, e.g. `insecure-commands`) should remove only the targeted value and keep the surrounding YAML well-formed, regardless of whether the value lives in a block or flow collection.

For example, removing `ACTIONS_ALLOW_UNSECURE_COMMANDS` from a flow-style mapping should preserve the sibling:

```yaml
env: { ACTIONS_ALLOW_UNSECURE_COMMANDS: true, FOO: bar }
```

should become:

```yaml
env: { FOO: bar }
```

## Actual behavior

The `Remove` op deletes the entire line(s) spanning the targeted value. This is unsound for flow collections and for sole members:

- Removing one member of a flow mapping/sequence deletes the whole line, e.g. `{ a: 1, b: 2 }` with `b` removed corrupts the document.
- Removing the only key under a block mapping leaves a dangling null parent, e.g. removing the sole key under `env:` leaves `env:` with no value.

A prior fix attempt (#1020) detected the container by sniffing for `{`/`[` tokens, which is brittle because it cannot distinguish nesting such as `{[ ... ]}` from `[{ ... }]`.

## Reproduction steps

1. Create a workflow with a flow-style `env` that mixes the insecure flag with another variable:

```yaml
on: push
jobs:
  test:
    runs-on: ubuntu-latest
    env: { ACTIONS_ALLOW_UNSECURE_COMMANDS: true, FOO: bar }
    steps:
      - run: echo hi
```

2. Run zizmor in fix mode over the workflow.
3. Observe that the `env:` line is clobbered as a whole rather than leaving `env: { FOO: bar }`.

## Logs

_N/A_

## Additional context

This tracks the soundness of the `Remove` patch op in `yamlpath`/`yamlpatch`. Related: META #876, draft #1000, and prior attempt #1020. A fix is proposed in #2121, which resolves container kind from the parse tree (not token sniffing), handles flow separators, avoids following a terminal alias, and collapses dangling empty block parents.

mostafa:
## Pre-submission checks

Please check these boxes:

      one first).

      use that violates the policy.

## Summary

Closes #2122. Reworks the `Remove` patch op so it is sound for flow collections and nested structures, addressing the soundness concern raised in #1020 and following the "re-think within `yamlpath`" direction of #1000. Relates to META #876.

Previously `Remove` deleted whole lines, which:

- corrupted flow mappings and sequences (removing `b` from `{a: 1, b: 2}` deleted the whole line), and
- left dangling null parents (removing the only key under `env:` left `env:` with no value).

The earlier flow handling in #1020 detected the container by sniffing for `{`/`[` tokens, which is brittle because it cannot distinguish nesting like `{[ ... ]}` from `[{ ... }]`.

Removal span computation now lives in `yamlpath` as `Document::removal_span`, which uses the parse tree rather than token heuristics:

- Dispatches on the value's actual container kind (block vs flow) as reported by tree-sitter, so nested mixtures like `[{a: 1}, {b: 2}]` and `{x: [1, 2]}` resolve correctly.
- For flow members, removes the element plus a single adjacent comma and the whitespace up to its neighbor, and collapses a sole member cleanly to `{}` or `[]`.
- Resolves the element at its own lexical site, so a terminal alias is not followed: removing `b: *x` deletes `b: *x` rather than the anchor definition `a: &x 1` it points to.
- Collapses dangling empty block parents upward to the first ancestor that still has other content. Flow containers are left as valid `{}`/`[]`, and block containers only nest inside block containers, so collapse never crosses styles.

`yamlpatch`'s `Op::Remove` is now a thin call into `Document::removal_span`.

Verified outputs match the (draft) expectations in #1000: `foo: { a: a, b: b }` with `/foo/b` removed gives `foo: { a: a }`; an empty key `foo/baz` removal preserves the rest; and `foo: [1, 2, 3: {abc: def}]` with `/foo/2` removed gives `foo: [1, 2]`.

Design note: #1000 leaned toward extending `QueryMode::Pretty` in `query_node` and deleting that span. I added a dedicated `removal_span` instead, because removal has needs that pretty extraction does not (separator handling, sole-member collapse, parent collapse) and `query_pretty` is shared with `Op::Replace`. Happy to fold this into `query_pretty` if you prefer.

## Test Plan

- `cargo test -p yamlpath -p yamlpatch` (added snapshot coverage for block/flow mappings and sequences, nested flow collections, multiline flow, alias safety, and block parent collapse)
- `cargo test -p zizmor --bins insecure_commands` (the only current `Op::Remove` consumer)
- `cargo fmt --check`
- `cargo clippy -- --deny warnings`

### DIFF
--- a/crates/yamlpatch/tests/unit_tests.rs
+++ b/crates/yamlpatch/tests/unit_tests.rs
@@ -1432,6 +1432,45 @@ jobs:
 }
 
 #[test]
+fn test_remove_first_step_key_slides_nested_sibling_onto_marker() {
+    // The sibling that slides onto the `- ` marker keeps its own nested
+    // block correctly indented.
+    let yaml = "\
+steps:
+  - env:
+      X: 1
+    uses:
+      a: b
+";
+    insta::assert_snapshot!(remove(yaml, route!("steps", 0, "env")), @"
+    --- PATCH ---
+    steps:
+      - uses:
+          a: b
+
+    --- END PATCH ---
+    ");
+}
+
+#[test]
+fn test_remove_non_first_step_key_keeps_marker() {
+    // Removing a key that does not share the `- ` line uses ordinary
+    // whole-line removal, leaving the marker (and first key) intact.
+    let yaml = "\
+steps:
+  - env: prod
+    run: echo hi
+";
+    insta::assert_snapshot!(remove(yaml, route!("steps", 0, "run")), @"
+    --- PATCH ---
+    steps:
+      - env: prod
+
+    --- END PATCH ---
+    ");
+}
+
+#[test]
 fn test_multiple_operations_preserve_comments() {
     let original = r#"
 # Main configuration
--- a/crates/yamlpatch/tests/unit_tests.rs
+++ b/crates/yamlpatch/tests/unit_tests.rs
@@ -1401,6 +1401,37 @@ fn test_remove_sole_flow_member_does_not_collapse_parent() {
 }
 
 #[test]
+fn test_remove_sole_env_key_preserves_step_sequence_marker() {
+    // Regression: removing the only key under a *step-level* `env:` collapses
+    // `env` (good), but here `env:` is the first key of a `-` step item, so
+    // whole-line block removal also eats the `- ` marker. That turns the
+    // `steps:` list into a mapping and silently corrupts the document.
+    let yaml = "\
+jobs:
+  build:
+    steps:
+      - env:
+          ACTIONS_ALLOW_UNSECURE_COMMANDS: true
+        run: echo hi
+";
+    insta::assert_snapshot!(
+        remove(
+            yaml,
+            route!("jobs", "build", "steps", 0, "env", "ACTIONS_ALLOW_UNSECURE_COMMANDS")
+        ),
+        @"
+    --- PATCH ---
+    jobs:
+      build:
+        steps:
+          - run: echo hi
+
+    --- END PATCH ---
+    "
+    );
+}
+
+#[test]
 fn test_multiple_operations_preserve_comments() {
     let original = r#"
 # Main configuration
--- a/crates/yamlpatch/tests/unit_tests.rs
+++ b/crates/yamlpatch/tests/unit_tests.rs
@@ -1094,6 +1094,312 @@ permissions:
     ");
 }
 
+/// Apply a single `Op::Remove` at `route` to `yaml` and return the
+/// resulting source wrapped with `format_patch` for snapshotting.
+fn remove(yaml: &str, route: yamlpath::Route) -> String {
+    let document = yamlpath::Document::new(yaml).unwrap();
+    let result = apply_yaml_patches(
+        &document,
+        &[Patch {
+            route,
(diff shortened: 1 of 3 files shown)


woodruffw:
Thanks @mostafa! I'm hoping to give this a review in the next day or so.

woodruffw:
"Next day or so" apparently means next week for me, but I'm reviewing this now...

crates/yamlpatch/tests/unit_tests.rs
```diff
@@ -1400,6 +1400,37 @@ fn test_remove_sole_flow_member_does_not_collapse_parent() {
...
+    --- PATCH ---
+    jobs:
+      build:
+        steps:
+          - run: echo hi
+
+    --- END PATCH ---
+    "
+    );
+}
```
woodruffw, line 1471: I added this as a (currently) failing reproducer -- our current `Op::Remove` deletes the leading `-` when deleting a now-empty mapping, which incorrectly converts the parent type from a list into a mapping.

I _think_ the cleanest thing to do here is either handle the removal of a single remaining mapping member as a transformation to `mapping: {}` _or_ just make sure the list type survives when deleting the entire mapping.

So option 1:

```yaml
jobs:
  build:
    steps:
      - env: {}
      - run: echo hi
```

or option 2:

```yaml
jobs:
  build:
    steps:
      - run: echo hi
```

(2) is arguably nicer, but I suspect (1) might be a little easier to implement.

mostafa: Went with option 2 (preserve the list). Your failing testcase passes now.

I detect this structurally rather than textually: when the removed key is a `block_mapping_pair` whose container `block_mapping` is the inline value of a `block_sequence_item` (`block_sequence_item -> block_node -> block_mapping`) and the key shares the `- ` marker's line, I remove from the key's start up to the next sibling key's start. That keeps the `- ` and slides the following key onto its line.

I added two regression tests next to yours: one where the sliding sibling has its own nested block (indentation is preserved), and one removing a non-first key (ordinary whole-line removal, marker untouched).

One edge I left as-is: if a standalone comment sits between the removed first key and the next key, it gets absorbed into the removal. Preserving it would force the dangling-dash form (`- ` on its own line), which is uglier than the result you asked for. Happy to revisit if you'd prefer comment preservation there.

mostafa: Addressed in https://github.com/zizmorcore/zizmor/pull/2121/commits/8570c75caf56bef06d5edf22462ae7949f00505b.

woodruffw (approved):
Thank you @mostafa!

Issue #2123, opened 2026-06-14
Feature: Display folder name for composite actions instead of just "action.yml" to be able to determine what specific compose action has issues


PrinsFrank:
### Pre-submission checks

- [x] I am **not** reporting a bug (crash, false positive/negative, etc). These must be filed via the bug report template.
- [x] I have looked through both the [open](https://github.com/zizmorcore/zizmor/issues?q=is%3Aissue+is%3Aopen+) and [closed](https://github.com/zizmorcore/zizmor/issues?q=is%3Aissue%20state%3Aclosed) issues for a duplicate request.
- [x] I affirm that I have followed the [AI policy](https://github.com/zizmorcore/.github/blob/main/AI_POLICY.md) and that this request constitutes my own words and understanding.

### What's the problem this feature will solve?

Right now, When a repository has many composite actions like this:
```
.github/actions/setup-a/action.yml
.github/actions/setup-b/action.yml
```

The output only shows the filename:
`action.yml:24: code injection (....)`
This makes it impossible to determine what specific file has the issue without going over all composite actions and checking if the issues occur in each one.

### Describe the solution you'd like

When analyzing a composite action. include the path starting at `.github/actions`, so including the foldername

woodruffw:
Hi @PrinsFrank, thank you for opening an issue. In the future, please double check for duplicates! See: #2113, #2046, #1745. 

This is already fixed on HEAD but not in a release yet. #2095 will further improve it.

PR #2124, opened 2026-06-14
Use filepath instead of filename in error/warning output on Github Actions

PrinsFrank:
## Pre-submission checks

Please check these boxes:

      one first). #2123 

      use that violates the policy.

## Summary

This PR changes the output for github actions to display the full path inline instead of only the filename. There can be many workflows with the same name in a single repository. For example, for composite actions, the structure of files is always `composite-action-name/action.yml`. when there are many such composite actions, all errors will have the prefix `action.yml`, which is not enough determine in what specific composite action the error occurs in.

This is my first ever contribution to a rust repo. Any help would be appreciated. ;)

### DIFF
--- a/crates/zizmor/src/output/github.rs
+++ b/crates/zizmor/src/output/github.rs
@@ -34,8 +34,7 @@ impl Finding<'_> {
         let title = self.ident;
 
         let message = format!(
-            "{filename}:{start_line}: {desc}: {annotation}",
-            filename = primary.symbolic.key.filename(),
+            "{filepath}:{start_line}: {desc}: {annotation}",
             desc = self.desc,
             annotation = primary.symbolic.annotation,
         );


crates/zizmor/src/output/github.rs
```diff
@@ -34,8 +34,7 @@ impl Finding<'_> {
         let title = self.ident;
 
         let message = format!(
-            "{filename}:{start_line}: {desc}: {annotation}",
-            filename = primary.symbolic.key.filename(),
+            "{filepath}:{start_line}: {desc}: {annotation}",
```
woodruffw, line 37: The `filename` here is _not_ the filename that controls the annotation itself: that's in `file={filepath}` below, which already uses the more correct path.

woodruffw:
Thanks for opening this @PrinsFrank, but this was already fixed in #1748 (just not released yet). #2095 will further improve it by ensuring we always emit a repo-relative path, regardless of the input path used by the user.

PR #2127, opened 2026-06-15
feat(unpinned-uses): pin to full version when fixing a major ref

### LINKED ISSUE #2093: Should --fix for unpinned-uses pin to the full version, when given the major version?
### Discussed in https://github.com/orgs/zizmorcore/discussions/2092

<sup>Originally posted by **inahga** June  2, 2026</sup>
Consider
```
- uses: actions/checkout@v6
```
This is appropriately flagged, and --fix=all will expand it to:
```
- uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
```
However, leaving the tag at `v6` is not very useful. When actions/checkout updates, and moves its v6 tag forward, this workflow will start to fail lints with `ref-version-mismatch`.

It would be better if zizmor resolved v6 to the latest available full version tag, and used that.
```
- uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
```

Note that `ref-version-mismatch` has an effective autofix which will expand the version tag, but that only applies once the action is updated and we have failing CI. It would be better if the autofix just fixed it right in the first place.

(This is also particularly annoying because I have 20 or so repos to blast zizmor across :sob:)

Happy to contribute a changeset, just checking if the maintainers are OK with this first.</div>

david-wagih:
Closes #2093.

## What

When `zizmor --fix` pins an unpinned action that was referenced by a
major-version tag (e.g. `actions/checkout@v6`), the version comment now
reflects the **full** resolved version instead of the major ref:

```yaml
# before
- uses: actions/checkout@de0fac…83dd # v6

# after
- uses: actions/checkout@de0fac…83dd # v6.0.2
```

## Why

Pinning the comment to the major ref means that once upstream moves the `v6`
tag forward, the pinned SHA no longer matches what `v6` points at, and
`ref-version-mismatch` flags the workflow — requiring a second fix. Resolving
to the full version up front avoids that cascade.

## How

After resolving the ref to a commit SHA, `UnpinnedUses::attempt_fix` now
resolves that SHA back to its longest tag via the existing
`Client::longest_tag_for_commit` (the same primitive `ref-version-mismatch`
uses, which prefers `v1.2.3` over `v1`). It falls back to the original ref if
the commit has no matching tag. The hash pin and the fix title are unchanged.

## Tests

Adds a `gh-token-tests` snapshot test using `actions/checkout@v1` — an
end-of-life major tag that no longer moves, so the resolved SHA and full
version (`# v1.2.0`) are stable. Existing full-ref snapshot tests are
unchanged, confirming no regression for already-full refs.

## AI usage disclosure

Per the project's AI policy: this contribution was made with AI assistance
(Claude). The author has reviewed and understands the change.

### DIFF
--- a/crates/zizmor/src/audit/unpinned_uses.rs
+++ b/crates/zizmor/src/audit/unpinned_uses.rs
@@ -70,6 +70,20 @@ impl UnpinnedUses {
             }
         };
 
+        // Resolve the commit back to its longest tag so the version comment
+        // reflects the full version (e.g. `v6.0.2`) rather than the major ref
+        // (e.g. `v6`) the user wrote. Pinning to the full version avoids a
+        // later `ref-version-mismatch` finding when the major tag is moved
+        // forward by the upstream maintainers.
+        // Fall back to the original ref if the commit has no matching tag.
+        let comment_ref = match client
+            .longest_tag_for_commit(uses.owner(), uses.repo(), &commit)
+            .await
+        {
+            Ok(Some(tag)) => tag.name,
+            _ => uses.git_ref().to_string(),
+        };
+
         let action = if let Some(subpath) = uses.subpath() {
             format!("{}/{}", uses.slug(), subpath)
         } else {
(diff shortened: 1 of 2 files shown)


woodruffw (commented):
Thanks @david-wagih, this looks reasonable to me. In the future please make sure to use one of the PR templates.

woodruffw (approved):

david-wagih:
> Thanks @david-wagih, this looks reasonable to me. In the future please make sure to use one of the PR templates.

Sure thing, thanks for the feedback ❤️

PR #2131, opened 2026-06-17
docs: pin manual uvx zizmor examples

shaanmajid:
## Pre-submission checks

Please check these boxes:

      one first).

      use that violates the policy.

## Summary

Closes #2129.

Notes: 
 - I considered using the current released zizmor version in the examples, but that would need release-time docs bumping to avoid going stale. Since the project doesn't currently appear to have release tooling for version-bearing docs (e.g., [cargo-release](https://github.com/crate-ci/cargo-release), [rooster](https://github.com/zanieb/rooster)), this uses an explicit placeholder instead.
- Feel free to push any style/wording tweaks -- no need to roundtrip :^)

## Test Plan

N/A; docs-only change

### DIFF
--- a/docs/integrations.md
+++ b/docs/integrations.md
@@ -86,7 +86,7 @@ GitHub Actions setup:
 
     jobs:
       zizmor:
-        name: zizmor latest via PyPI
+        name: zizmor via PyPI
         runs-on: ubuntu-latest
         permissions:
           security-events: write # Required for upload-sarif (used by zizmor-action) to upload SARIF files.
@@ -102,7 +102,7 @@ GitHub Actions setup:
             uses: astral-sh/setup-uv@08807647e7069bb48b6ef5acd8ec9567f424441b # v8.1.0
 
           - name: Run zizmor 🌈
-            run: uvx zizmor --format=sarif . > results.sarif # (2)!
+            run: uvx 'zizmor@<version>' --format=sarif . > results.sarif # (2)!
             env:
               GH_TOKEN: ${{ secrets.GITHUB_TOKEN }} # (1)!
 
@@ -115,9 +115,12 @@ GitHub Actions setup:
 
     1. Optional: Remove the `env:` block to only run `zizmor`'s offline audits.
 
-    2. This installs the [zizmor package from PyPI], since it's pre-compiled
-       and therefore completes much faster. You could instead compile `zizmor`
-       within CI/CD with `cargo install zizmor`.
+    2. This installs a pinned [zizmor package from PyPI], since it's
+       pre-compiled and therefore completes much faster. You could instead
+       compile `zizmor` within CI/CD with
+       `cargo install 'zizmor@<version>'`. Replace `<version>` with the release
+       you want, and keep the pin current with your usual dependency update
+       tooling.
 
     For more inspiration, see `zizmor`'s own [repository workflow scan], as well
     as GitHub's example of [running ESLint] as a security workflow.
@@ -158,7 +161,7 @@ GitHub Actions setup:
 
     jobs:
       zizmor:
-        name: zizmor latest via PyPI
+        name: zizmor via PyPI
         runs-on: ubuntu-latest
         permissions:
           contents: read # Only needed for private repos. Needed to clone the repo.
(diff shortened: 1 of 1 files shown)


docs/integrations.md
```diff
@@ -102,7 +102,7 @@ GitHub Actions setup:
             uses: astral-sh/setup-uv@08807647e7069bb48b6ef5acd8ec9567f424441b # v8.1.0
 
           - name: Run zizmor 🌈
-            run: uvx zizmor --format=sarif . > results.sarif # (2)!
+            run: uvx 'zizmor@<version>' --format=sarif . > results.sarif # (2)!
```
woodruffw: I think it might be nice to make this copy-pasteable, by having something like this instead:

```
env:
  ZIZMOR_VERSION: 1.25.2

...

run: |
  uvx "zizmor@${ZIZMOR_VERSION}"
```

(i.e. the env could go at the top of the workflow.)

This would also be pretty easy for me to maintain, since I could just substitute in each newly-released version.

docs/integrations.md
```diff
@@ -115,9 +115,12 @@ GitHub Actions setup:
 
     1. Optional: Remove the `env:` block to only run `zizmor`'s offline audits.
 
-    2. This installs the [zizmor package from PyPI], since it's pre-compiled
-       and therefore completes much faster. You could instead compile `zizmor`
-       within CI/CD with `cargo install zizmor`.
+    2. This installs a pinned [zizmor package from PyPI], since it's
```
woodruffw, line 121: I actually think we should remove the `cargo install` language entirely since it isn't a flow I technically support.

So I think this tooltip would be safe to remove entirely 🙂

docs/integrations.md
```diff
@@ -172,16 +175,19 @@ GitHub Actions setup:
...
             env:
               GH_TOKEN: ${{ secrets.GITHUB_TOKEN }} # (1)!
     ```
 
     1. Optional: Remove the `env:` block to only run `zizmor`'s offline audits.
 
-    2. This installs the [zizmor package from PyPI], since it's pre-compiled
-       and therefore completes much faster. You could instead compile `zizmor`
-       within CI/CD with `cargo install zizmor`.
+    2. This installs a pinned [zizmor package from PyPI], since it's
```
woodruffw, line 187: (Same here.)

docs/integrations.md
```diff
@@ -191,6 +197,11 @@ GitHub Actions setup:
         be rendered; all subsequent findings will be logged in the actions log but
         **will not be rendered** as annotations.
 
+!!! tip
```
woodruffw: I think this tip is correct, but might be confusing to a user who is integrating zizmor but doesn't know much about Python packaging. I think my preference is to leave it out for now, and maybe add a new section (like "Python packaging") in a follow-up PR.

woodruffw (commented):
Thanks @shaanmajid, just a few nitpicks.

docs/integrations.md
shaanmajid: Ah, if you're fine with manually keeping this up-to-date: do you strongly prefer this over just an inline `uvx zizmor@1.25.2`? The reason I used a placeholder was to avoid you having to maintain it across bumps, but if you're open to that, then I think in-line might be simplest for consumers?

docs/integrations.md
shaanmajid: Updated in f15ad67.

docs/integrations.md
shaanmajid: Updated in [449aa21](https://github.com/zizmorcore/zizmor/pull/2131/commits/449aa21e0e0fa71d304e0e2f54b54b703162eda4)

docs/integrations.md
shaanmajid: Ditto :^)

docs/integrations.md
shaanmajid: Went ahead with your env-var suggestion for now :^); updated in 17cae4b.

woodruffw (approved):
Thanks @shaanmajid!

docs/integrations.md
woodruffw: Yeah, I'm fine with keeping it updated -- I need to automate the version rewriting across docs more generally, and this is another good motivator for doing so.

(I personally like the env style because I believe Renovate and some other tools support it well.)

Issue #1865, opened 2026-04-07
[BUG]: False positive `superfluous-actions` for `dtolnay/rust-toolchain` on self-hosted runner


AntoineSebert:
### Pre-submission checks

- [x] I am **not** filing a feature request. These should be filed via the feature request form instead.
- [x] I have checked the [Troubleshooting Guide](https://docs.zizmor.sh/troubleshooting) for my problem.
- [x] I have looked through both the [open](https://github.com/zizmorcore/zizmor/issues?q=is%3Aissue+is%3Aopen+) and [closed](https://github.com/zizmorcore/zizmor/issues?q=is%3Aissue%20state%3Aclosed) issues for a duplicate report.

### zizmor version

1.23.1

### Expected behavior

No `superfluous-actions` when the functionality provided by the action is not present on the runner:
```
zizmor --gh-token=$(gh auth token) -c ~/zizmor.yml .github/workflows/test.yaml
🌈 zizmor v1.23.1
 INFO audit: zizmor: 🌈 completed .github/workflows/test.yaml
No findings to report. Good job! (2 suppressed)
```

### Actual behavior

The audit reports a `superfluous-actions`:
```
zizmor --gh-token=$(gh auth token) -c ~/zizmor.yml .github/workflows/test.yaml
🌈 zizmor v1.23.1
 INFO audit: zizmor: 🌈 completed .github/workflows/test.yaml
help[superfluous-actions]: action functionality is already included by the runner
  --> .github/workflows/test.yaml:28:15
   |
28 |       - uses: dtolnay/rust-toolchain@stable
   |         ------^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |         |     |
   |         |     use `rustup` and/or `cargo` in a script step
   |         this step
   |
   = note: audit confidence → High

2 findings (1 suppressed): 0 informational, 1 low, 0 medium, 0 high
```

Once replaced `- uses: dtolnay/rust-toolchain@stable` by `- run: rustup install stable && cargo init`, the job run fails:
```
act -s GITHUB_TOKEN="$(gh auth token)" -P liquid-github-runner-set=catthehacker/ubuntu:act-24.04 -W .github/workflows/test.yaml
INFO[0000] Using docker host 'unix:///var/run/docker.sock', and daemon socket 'unix:///var/run/docker.sock' 
[Test/Test] ⭐ Run Set up job
[Test/Test] 🚀  Start image=catthehacker/ubuntu:act-24.04
[...]

Issue #22, opened 2024-09-22
Template injection: don't flag on `matrix` expansions when the matrix is statically defined


woodruffw:
This will be 100% FP.

h/t @alex

VorpalBlade:
I believe I have come across another instance of this issue, though I'm not sure so I'll post my example here:

```
    |
105 |     runs-on: ${{ matrix.os }}
    |     ------------------------- note: expression may expand into a self-hosted runner
```

where

```yaml
    strategy:
      matrix:
        include:
          - target: aarch64-unknown-linux-gnu
            os: ubuntu-latest
            rustflags: -Clink-arg=-Wl,--compress-debug-sections=zlib
          - target: aarch64-unknown-linux-musl
            os: ubuntu-latest
            rustflags: -Clink-arg=-Wl,--compress-debug-sections=zlib
          - target: armv7-unknown-linux-gnueabihf
            os: ubuntu-latest
            rustflags: -Clink-arg=-Wl,--compress-debug-sections=zlib
          - target: armv7-unknown-linux-musleabihf
            os: ubuntu-latest
            rustflags: -Clink-arg=-Wl,--compress-debug-sections=zlib
          - target: i686-unknown-linux-gnu
            os: ubuntu-latest
            rustflags: -Clink-arg=-Wl,--compress-debug-sections=zlib
          - target: i686-unknown-linux-musl
            os: ubuntu-latest
            rustflags: -Clink-arg=-Wl,--compress-debug-sections=zlib
          - target: x86_64-unknown-linux-gnu
            os: ubuntu-latest
            rustflags: -Clink-arg=-Wl,--compress-debug-sections=zlib
          - target: x86_64-unknown-linux-musl
            os: ubuntu-latest
            rustflags: -Clink-arg=-Wl,--compress-debug-sections=zlib
          - target: x86_64-pc-windows-msvc
            os: windows-latest
            rustflags:
          - target: x86_64-apple-darwin
            os: macos-latest
            rustflags:
          - target: aarch64-apple-darwin
            os: macos-latest
            rustflags:
```

All the OSes are hosted runners by github, so this is a false positive.

woodruffw:
Thanks @VorpalBlade! That does indeed look like a lingering false positive -- would you be able to share your entire workflow so that I can fully reproduce the bug here?

VorpalBlade:
Sure, that would be https://github.com/VorpalBlade/chezmoi_modify_manager/blob/main/.github/workflows/release.yml

[...]

PR #2140, opened 2026-06-22
Foundation for a better 'self-hosted' check

woodruffw:
This is towards #1865.

The idea is to have a better (more sound) "self-hosted" check than our current one, which in turn will allow me to implement something like `NormalJob::is_self_hosted` for reuse.


woodruffw:
TODOs:

- [ ] `NormalJob::is_self_hosted()` probably needs to return `Vec<SelfHosted>` instead, where `SelfHosted { Yes(hints: SymbolicLocation), Indeterminate }`. An empty `Vec` will then signal that the job is definitely not self hosted, while one or more `SelfHosted` results will signal _how_ it (might) be self-hosted.
- [ ] I probably need to make the self-hosted check more precise in general, since this is revealing a number of places where we were previously unsound. For example in the repro for #22 we have: `runner: ${{ github.repository_owner == 'python' && 'ubuntu-24.04-aarch64' || 'ubuntu-24.04' }}` and this _should_ probably flag a self-hosted runner (for `ubuntu-24.04-aarch64`).

crates/zizmor/data/github-runners.txt
```diff
@@ -0,0 +1,28 @@
...
+ubuntu-26.04
+ubuntu-26.04-arm
+ubuntu-latest
+ubuntu-slim
+windows-11-arm
+windows-11-vs2026-arm
+windows-2022
+windows-2025
+windows-2025-vs2026
+windows-latest
```
rmuir, line 28: hmm, but this isn't just "either-or". 

I'm making a lot of use of github-hosted runners with the [azure private networking](https://docs.github.com/en/organizations/managing-organization-settings/about-azure-private-networking-for-github-hosted-runners-in-your-organization) feature, which means my runners get strange names and will get flagged as self-hosted if they don't match this list? They aren't self-hosted though.

I just wanted to make sure you were aware of the false-positive there.

rmuir: another use-case where I end out with wacky runner names, yet they are still github-hosted: [custom images](https://docs.github.com/en/actions/how-tos/manage-runners/larger-runners/use-custom-images)

woodruffw: Ah, that's unfortunate. There doesn't really seem to be a static solution to this, at least not without querying a bunch of APIs that I don't want zizmor to have permission to access.

The "good" news here is that this audit will remain in the auditor persona, so it shouldn't really affect most users.

rmuir: actionlint has a check that you are using a valid runner name, I configure the custom runners in the `actionlint.yml` as if they were "self-hosted" like this:

```yaml
# These aren't really self-hosted, but configure actionlint as if they were.
# This way runner name can still be validated by actionlint without disabling checks.
self-hosted-runner:
  # Labels of self-hosted runner in array of strings.
  labels:
    # corporate access with cached resources
    - azure-custom-runner
    # image generation with corporate access
    - image-gen-azure
```

PR #2149, opened 2026-06-25
feat: support parallel steps (background, wait, wait-all, cancel, parallel)

devantler:
## Summary

[GitHub released parallel steps on 2026-06-25](https://github.blog/changelog/2026-06-25-actions-steps-can-now-be-run-in-parallel/), adding five new step keys: `background`, `wait`, `wait-all`, `cancel`, and `parallel`. Today zizmor can't parse a workflow that uses the new *step kinds* and bails out instead of auditing it:

```console
$ zizmor --offline pl.yml
 WARN ... failed to validate file://pl.yml as workflow: input does not match expected validation schema
fatal: no audit was performed
error: no inputs collected
```

(`background` alone already parsed, since serde ignores unknown fields; the `wait`/`wait-all`/`cancel`/`parallel` steps fail because they match neither the `Uses` nor `Run` body variant.)

With this change zizmor parses and audits them normally.

## Changes

In `github-actions-models` (`workflow/job.rs`), modeled after [`workflow-syntax.md`](https://github.com/github/docs/blob/main/content/actions/reference/workflows-and-actions/workflow-syntax.md):

- **`background`** → a new optional boolean field on `Step` (mirrors `continue-on-error`).
- **`wait`/`wait-all`**, **`cancel`**, **`parallel`** → new `StepBody` variants. Per the spec, `wait` is a single ID **or** a list, `cancel` is a **single** ID, and `parallel` holds a nested `Vec<Step>`.
- **`StepBody` now hand-implements `Deserialize`** (dispatching on key presence) instead of deriving an `untagged` enum. This is necessary because the `wait-all:` step's value is the YAML **null**, which serde's untagged buffering can't match against a struct variant (a derived untagged variant keyed on a null-valued field never matches; verified that the equivalent *non*-untagged struct does). Dispatching on key presence sidesteps that entirely.

In `zizmor`:

- The new control steps map to a new `StepBodyCommon::Other`, so the existing `uses:`/`run:` audits skip them.
- **Audits descend into `parallel:` groups.** When a job's steps are enumerated, the steps nested in a `parallel:` group are flattened in with the correct yamlpath route (e.g. `steps[0].parallel[1]`), so findings inside a parallel block point at the right location.

## Test plan

- New `github-actions-models` test (`test_parallel_steps`) covering `background`, `wait` (single + list), `wait-all`, `cancel`, and nested `parallel` — including the bare `wait-all:` form.
- New `zizmor` test (`test_steps_descend_into_parallel`) asserting `Steps` flattens the nested `parallel:` steps.
- `cargo test -p github-actions-models` (26) and `cargo test -p zizmor --bins` (118) green; `cargo fmt --check` clean; clippy clean on the changed files.
- Verified end-to-end: `template-injection` fires on a dangerous expression both at top level **and** in a nested `parallel:` step, each pointing at the correct line.

---
🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/crates/github-actions-models/src/workflow/job.rs
+++ b/crates/github-actions-models/src/workflow/job.rs
@@ -1,7 +1,7 @@
 //! Workflow jobs.
 
 use indexmap::IndexMap;
-use serde::Deserialize;
+use serde::{Deserialize, Deserializer};
 use yaml_serde::Value;
 
 use crate::common::expr::{BoE, LoE};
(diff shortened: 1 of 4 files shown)


woodruffw:
@devantler please see our [AI policy](https://github.com/zizmorcore/.github/blob/main/AI_POLICY.md). In particular, PRs should be accompanied by an issue, need to be greenlit before contribution, and must use our PR template. This doesn't do any of those.

(If you had opened an issue first, I would have told you that this is not a good candidate for AI contribution: it involves too many subtle parts of how pathing and modeling work in zizmor.)

devantler:
@woodruffw sorry about that. Look forward to see it supported when you have had time to look through it's spec, and what it requires.

Creating the issue now instead 👍

Issue #2151, opened 2026-06-25
Feature: Add support for parallel steps


devantler:
### Pre-submission checks

- [x] I am **not** reporting a bug (crash, false positive/negative, etc). These must be filed via the bug report template.
- [x] I have looked through both the [open](https://github.com/zizmorcore/zizmor/issues?q=is%3Aissue+is%3Aopen+) and [closed](https://github.com/zizmorcore/zizmor/issues?q=is%3Aissue%20state%3Aclosed) issues for a duplicate request.
- [x] I affirm that I have followed the [AI policy](https://github.com/zizmorcore/.github/blob/main/AI_POLICY.md) and that this request constitutes my own words and understanding.

### What's the problem this feature will solve?

[GitHub released parallel steps on 2026-06-25](https://github.blog/changelog/2026-06-25-actions-steps-can-now-be-run-in-parallel/), adding five new step keys: `background`, `wait`, `wait-all`, `cancel`, and `parallel`. Today zizmor can't parse a workflow that uses the new *step kinds* and bails out instead of auditing it. This is a feature request to support parsing of these new keys.

### Describe the solution you'd like

Parsing of the new keys so zizmor can scan workflows that adopt the feature.

devantler:
https://github.com/zizmorcore/zizmor/pull/2149 was an attempt at this, and it might inspire parts of a final implementation albeit it breaking the AI policy (which I mistakenly overlooked).

I am willing to help with the implementation or finalizing the solution direction.

woodruffw:
Thanks for opening this @devantler. This is not a good candidate for an AI contribution for the reasons mentioned in the PR you opened before.

(I'm currently experimenting with this new feature to understand how it works, since the docs aren't very clear. Given that it's brand new I'm not going to rush to get a half-working version of support within zizmor out.)

woodruffw:
Leaving some notes:

- Schema update mentions composite actions but GitHub's announcement does not: https://github.com/SchemaStore/schemastore/pull/5845 (ref #2152 on our side)
- Nested parallel blocks are not allowed (thankfully)

woodruffw:
This is done with #2153, but will be considered experimental in the next release.

PR #2153, opened 2026-06-26
Initial parallel steps modeling

woodruffw:
See #2151.

This is part 1 of N. ~~I haven't tried to add `parallel:` with this PR, just the other keys.~~

NB: This has the unfortunate side effect of introducing some optionality to the `StepCommon` trait, since composite actions don't (yet) appear to have these same keys. If they gain that support in the future, we'll be able to return to non-optional `StepBodyCommon` access.

### DIFF
--- a/crates/github-actions-models/src/workflow/job.rs
+++ b/crates/github-actions-models/src/workflow/job.rs
@@ -121,7 +121,7 @@ pub struct Step {
 }
 
 #[derive(Deserialize, Debug)]
-#[serde(rename_all = "kebab-case", untagged)]
+#[serde(rename_all_fields = "kebab-case", untagged)]
 pub enum StepBody {
     Uses {
         /// The GitHub Action being used.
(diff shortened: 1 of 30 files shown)


crates/github-actions-models/src/workflow/job.rs
```diff
@@ -106,6 +106,11 @@ pub struct Step {
     #[serde(default)]
     pub continue_on_error: BoE,
 
+    /// Whether the step runs asynchronously, i.e. does not block its successor from running.
```
woodruffw, line 109: The changes here made me realize that the APIs here probably need to be inverted: the current `struct Step > enum StepBody` approach is inflexible.

What we probably want is `enum Step` where each `Step` variant has a `flatten`ed `StepBody` or similar.

crates/zizmor/src/models.rs
```diff
@@ -63,8 +63,17 @@ pub(crate) enum StepBodyCommon<'s> {
...
+    /// within the same job. Ordering across jobs is not defined.
+    ///
+    /// At the moment, a step's ordering is defined lexically, i.e.
+    /// a step that appears lexically before another step is considered
+    /// first, even if step-level parallelism may change their execution order.
+    ///
+    /// This is an API rather than a derivation for each `StepCommon`
+    /// so that each implementation doesn't need to derive the entire
+    /// stack of traits implied by ordering (`PartialEq`, `PartialOrd`, etc).
+    fn ord(&self) -> impl Ord;
```
woodruffw, line 76: This was the cleanest solution I could think of to `parallel:` changing the straightforward "natural" ordering of steps by index within their job.

Issue #2146, opened 2026-06-22
[BUG]: Windows path separator inconsistency when auditing explicit file paths


chirizxc:
```bash
❯ zizmor --version
zizmor 1.26.1
```

OS: Windows 10

If we run zizmor using the command `zizmor .github`, the result is as follows:

```bash
❯ zizmor .github
 INFO zizmor: 🌈 zizmor v1.26.1                                                                                                                
 INFO audit: zizmor: 🌈 completed .github\actions\build-pgo-wheel\action.yaml
 INFO audit: zizmor: 🌈 completed .github\dependabot.yaml
 INFO audit: zizmor: 🌈 completed .github\workflows\ci.yaml
 WARN audit: zizmor: one or more inputs contains YAML anchors; see https://docs.zizmor.sh/usage/#yaml-anchors for details
 INFO audit: zizmor: 🌈 completed .github\workflows\macos.yaml
 INFO audit: zizmor: 🌈 completed .github\workflows\manylinux.yaml
 INFO audit: zizmor: 🌈 completed .github\workflows\musllinux.yaml
 INFO audit: zizmor: 🌈 completed .github\workflows\tests.yaml
 INFO audit: zizmor: 🌈 completed .github\workflows\wasm.yaml
 INFO audit: zizmor: 🌈 completed .github\workflows\windows.yaml
No findings to report. Good job! (9 ignored, 25 suppressed)
```

But if we run the command `zizmor .github/`, we'll see, for example, `.github/actions\build-pgo-wheel\action.yaml`

```bash
❯ zizmor .github/                                                                                                                              
 INFO zizmor: 🌈 zizmor v1.26.1                                                                                                                
 INFO audit: zizmor: 🌈 completed .github/actions\build-pgo-wheel\action.yaml
 INFO audit: zizmor: 🌈 completed .github/dependabot.yaml
 INFO audit: zizmor: 🌈 completed .github/workflows\ci.yaml
 WARN audit: zizmor: one or more inputs contains YAML anchors; see https://docs.zizmor.sh/usage/#yaml-anchors for details
 INFO audit: zizmor: 🌈 completed .github/workflows\macos.yaml
 INFO audit: zizmor: 🌈 completed .github/workflows\manylinux.yaml
 INFO audit: zizmor: 🌈 completed .github/workflows\musllinux.yaml
 INFO audit: zizmor: 🌈 completed .github/workflows\tests.yaml
[...]

PR #2154, opened 2026-06-26
fix: normalize path separators on Windows to ensure consistent output

chirizxc:
## Pre-submission checks

Please check these boxes:

      one first).

      use that violates the policy.

## Test Plan

Added test `test_input_key_windows_path_normalization` and run PR version `zizmor` locally:
изображение

### DIFF
--- a/crates/zizmor/src/registry/input.rs
+++ b/crates/zizmor/src/registry/input.rs
@@ -291,9 +291,7 @@ impl InputKey {
         // Normalize path separators on Windows to ensure consistent output
         // See: https://github.com/zizmorcore/zizmor/issues/2146
         #[cfg(windows)]
-        let given_path = {
-            Utf8PathBuf::from(path.as_str().replace('\\', "/"))
-        };
+        let given_path = { Utf8PathBuf::from(path.as_str().replace('\\', "/")) };
 
         #[cfg(not(windows))]
         let given_path = path.to_path_buf();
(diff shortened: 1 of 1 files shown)


woodruffw:
@chirizxc Could you please use our PR template? Thanks.

chirizxc:
> @chirizxc Could you please use our PR template? Thanks.

done

woodruffw (changes requested):
Thanks for the PR @chirizxc, but I'm not sure this is the approach we want -- post-hoc transformation of paths into a cross-platform normal form is pretty brittle. I need to look more at _why_ this changed with 1.26 and ensure that we propagate paths in the appropriate native form (except where required by formats like SARIF).

chirizxc:
> Thanks for the PR @chirizxc, but I'm not sure this is the approach we want -- post-hoc transformation of paths into a cross-platform normal form is pretty brittle. I need to look more at _why_ this changed with 1.26 and ensure that we propagate paths in the appropriate native form (except where required by formats like SARIF).

But it didn't change in version 1.26; it seems this behavior has always been there

woodruffw:
Oh, I misread the logs in #2146, sorry. I still think we want to preserve the native path format here, but I need to think about it more.

shaanmajid:
Maybe something the following would be acceptable?

```rust
  fn native_path(path: &Utf8Path) -> Utf8PathBuf {
        #[cfg(windows)]
        {
            path.components().collect()
        }

        #[cfg(not(windows))]
        {
            path.to_path_buf()
        }
    }
```

PR #2159, opened 2026-06-29
Fix `on.workflow_call.secrets: inherit` deserialization and flag it in secrets-inherit

### LINKED ISSUE #1673: [BUG]: `on.workflow_call.secrets: inherit` causes deserialization failure
### Pre-submission checks

### zizmor version

main

### Expected behavior

zizmor should parse workflows that use `secrets: inherit` in `on.workflow_call` without error. This is valid GitHub Actions syntax per the [reusing workflows documentation](https://docs.github.com/en/actions/how-tos/reuse-automations/reuse-workflows#using-inputs-and-secrets-in-a-reusable-workflow).

### Actual behavior

zizmor fails with `data did not match any variant of untagged enum Trigger`.

### Reproduction steps

```yaml
name: test
on:
  workflow_call:
    secrets: inherit
jobs:
  test:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v4
```

```bash
zizmor /tmp/secrets-inherit.yml
```

### Logs

```plain text
🌈 zizmor v1.23.0-rc5
fatal: no audit was performed
error: failed to load file:///tmp/secrets-inherit.yml as workflow
  |

Caused by:
    0: failed to load file:///tmp/secrets-inherit.yml as workflow
    2: data did not match any variant of untagged enum Trigger
```

### Additional context

Found in the wild in [arktypeio/arktype](https://github.com/arktypeio/arktype/blob/6d0639bf479cbe51a1df661d457517ac81411ac9/.github/workflows/pullfrog.yml#L14) with a valid GitHub Actions [run](https://github.com/arktypeio/arktype/actions/runs/19977380887).

david-wagih:
## Pre-submission checks

Please check these boxes:

      one first).

      use that violates the policy.

## Summary

Closes #1673.

A reusable workflow may declare `on.workflow_call.secrets: inherit` (a bare
string) to accept all of its caller's secrets. The callee-side
`WorkflowCall.secrets` field was typed as
`IndexMap<String, Option<WorkflowCallSecret>>`, so it only accepted a
_mapping_ — the bare `inherit` string failed to deserialize, which cascaded up
the untagged `Trigger` enum (`data did not match any variant of untagged enum
Trigger`) and caused zizmor to **skip all audits** on any workflow using this
valid syntax.

Per @woodruffw's note on the issue, this does two things in one PR:

**1. Model fix** — Introduce a `WorkflowCallSecrets` enum (`Inherit` | untagged
`Map`) that mirrors the existing caller-side `job::Secrets`, and type the field
as `Option<WorkflowCallSecrets>`. Nothing read the callee-side `secrets` field
before, so the blast radius is limited to the model.

**2. Audit fix** — Extend the `secrets-inherit` audit with `audit_workflow` so
it also flags a **callee** that declares `secrets: inherit`. The callee
requesting blanket inheritance is arguably worse than the caller, since it
forces _every_ caller to over-scope. The finding uses the same Medium severity
/ High confidence as the existing caller-side finding. The caller-side per-job
auditing is preserved.

## Test Plan

- Unit test for the new `WorkflowCallSecrets` model (`inherit`, explicit
  mapping, absent).
- Integration tests: a callee with `secrets: inherit` is flagged; a callee with
  an explicit mapping is not.
- Docs (`audits.md`) and release notes updated.

`cargo test`, `cargo clippy -- --deny warnings`, and `cargo fmt --check` all
pass.


david-wagih:
Hi @woodruffw 
Waiting for your review on this one.

woodruffw:
@david-wagih Please use our PR template, thanks.

(You also don't need to ping me for reviews -- I see everything that comes in, and I just haven't gotten to it yet. I will hopefully have time for it today.)

crates/github-actions-models/src/workflow/event.rs
```diff
@@ -237,8 +237,22 @@ pub struct WorkflowCall {
...
+/// which unconditionally forwards all of the calling workflow's secrets.
+#[derive(Deserialize, Serialize, Debug, PartialEq)]
+#[serde(rename_all = "kebab-case")]
+pub enum WorkflowCallSecrets {
+    /// `secrets: inherit`
+    Inherit,
+    /// An explicit mapping of named secrets.
+    #[serde(untagged)]
+    Map(IndexMap<String, Option<WorkflowCallSecret>>),
 }
```
woodruffw: This seems like it should reuse the existing `workflow::job::Secrets` type (which has the same shape, but also reuses `Env` internally).

(We can move that type to the top-level `workflow` module to make it clear that it's shared across the `workflow::job` and `workflow::event` APIs.)

crates/zizmor/src/models/workflow.rs
```diff
@@ -172,6 +172,23 @@ impl Workflow {
         }
     }
 
+    /// Whether this workflow declares `on.workflow_call.secrets: inherit`,
+    /// i.e. whether it's a reusable workflow that unconditionally inherits
+    /// all of its caller's secrets.
+    pub(crate) fn has_workflow_call_secrets_inherit(&self) -> bool {
```
woodruffw: This only has one callsite, so I don't think we need a new API here -- let's just inline this check into the `secrets-inherit` audit.

woodruffw (commented):
Thanks @david-wagih, just some nitpicks. This otherwise looks pretty good to me!

crates/zizmor/src/models/workflow.rs
david-wagih: Done

crates/github-actions-models/src/workflow/event.rs
david-wagih: Done in `5c1fb3b`, with one adjustment: the two types turned out not to have quite the same shape.

`job::Secrets::Env` reuses `Env`, meaning `IndexMap<String, EnvValue>`, where `EnvValue` is scalar-only. However, `on.workflow_call.secrets` maps secret names to `{ description, required }` declarations.

So reusing `job::Secrets` directly would fail to deserialize a valid explicit declaration like:

```yaml
on:
  workflow_call:
    secrets:
      my-secret:
        required: true
```

The existing `test_workflow_call_secrets` unit test catches this.

I moved the type to the top-level `workflow` module as `Secrets<V = EnvValue>`. The caller side, `jobs.<id>.secrets`, keeps using it exactly as before through the `EnvValue` default, while the callee side instantiates it as `Secrets<Option<WorkflowCallSecret>>`.

One serde note: the `#[serde(default)]` on the `Env` variant field had to be removed because serde’s derive infers a `V: Default` bound from it. It was effectively a no-op anyway, since an untagged newtype variant has no missing-field case for a default to fill.

david-wagih:
@woodruffw 

I am looking forward for your second review on this one.
I am sorry I am mentioning you here again for this PR.

david-wagih:
@woodruffw 
Hi 
if I may ask about the timeline for this milestone

PR #2166, opened 2026-07-05
feat: add timeout-minutes audit

### LINKED ISSUE #1023: Feature: Require or recommend the `timeout-minutes` property for all jobs

mchevestrier:
## Pre-submission checks

Please check these boxes:

      one first).

      use that violates the policy.

## Summary

Resolves #1023

This builds on #1347, but I also added a fix that will set `timeout-minutes` to a (configurable) default value on the job.

## Test Plan

Integration test at `crates/zizmor/tests/integration/audit/timeout_minutes.rs`

### DIFF
--- a/crates/zizmor/src/audit/timeout_minutes.rs
+++ b/crates/zizmor/src/audit/timeout_minutes.rs
@@ -31,7 +31,7 @@ impl Audit for TimeoutMinutes {
         if job.steps().all(|step| step.timeout_minutes().is_none()) {
             return Ok(vec![
                 Self::finding()
-                    .severity(Severity::Medium)
+                    .severity(Severity::Low)
                     .confidence(Confidence::High)
                     .persona(Persona::Pedantic)
                     .add_location(
@@ -53,7 +53,7 @@ impl Audit for TimeoutMinutes {
 
             findings.push(
                 Self::finding()
-                    .severity(Severity::Medium)
+                    .severity(Severity::Low)
                     .confidence(Confidence::High)
                     .persona(Persona::Pedantic)
                     .add_location(
--- a/crates/zizmor/src/audit/timeout_minutes.rs
+++ b/crates/zizmor/src/audit/timeout_minutes.rs
@@ -28,7 +28,7 @@ impl Audit for TimeoutMinutes {
             return Ok(vec![]);
         }
 
-        if job.steps().all(|step| step.timeout_minutes().is_none()) {
+        if job.steps().any(|step| step.timeout_minutes().is_none()) {
             return Ok(vec![
                 Self::finding()
                     .severity(Severity::Low)
@@ -44,29 +44,7 @@ impl Audit for TimeoutMinutes {
             ]);
         }
 
-        let mut findings = vec![];
-
-        for step in job.steps() {
-            if step.timeout_minutes().is_some() {
-                continue;
-            }
-
-            findings.push(
-                Self::finding()
-                    .severity(Severity::Low)
-                    .confidence(Confidence::High)
-                    .persona(Persona::Pedantic)
-                    .add_location(
-                        step.location()
-                            .primary()
-                            .annotated("step missing timeout-minutes"),
-                    )
-                    .fix(Self::create_add_timeout_fix_job(job, config))
-                    .build(job)?,
-            );
-        }
-
-        Ok(findings)
+        Ok(vec![])
     }
 }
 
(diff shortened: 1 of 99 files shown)


woodruffw:
Sorry, I haven't had time to review this. @mchevestrier if you can deconflict, I'm going to try and queue this for 1.28.0.

crates/zizmor/src/audit/timeout_minutes.rs
```diff
@@ -0,0 +1,215 @@
...
+        config: &crate::config::Config,
+    ) -> Result<Vec<Finding<'doc>>, AuditError> {
+        if job.timeout_minutes.is_some() {
+            return Ok(vec![]);
+        }
+
+        if job.steps().all(|step| step.timeout_minutes().is_none()) {
+            return Ok(vec![
+                Self::finding()
+                    .severity(Severity::Medium)
```
woodruffw: I think this should probably be `Severity::Low`, to match `concurrency-limits`.

crates/zizmor/src/audit/timeout_minutes.rs
```diff
@@ -0,0 +1,215 @@
...
+                            .annotated("job missing timeout-minutes"),
+                    )
+                    .fix(Self::create_add_timeout_fix_job(job, config))
+                    .build(job)?,
+            ]);
+        }
+
+        let mut findings = vec![];
+
+        for step in job.steps() {
```
woodruffw: I might be missing something here, but doesn't this mean we'd flag a step like this?

```yaml
uses: actions/checkout@...
```

I think that might be pretty surprising to users, since most users _probably_ don't want to put a timeout on that kind of action step.

woodruffw (commented):
Thanks @mchevestrier! The overall approach here looks reasonable to me, just some small nitpicks from an initial review.

At a high level, my main concern with this change is that it's going to be _very_ noisy for a lot of users for a mostly marginal security benefit: anybody who uses the pedantic persona will probably get quite a few hits here.

I'm not sure if that's a blocker yet, just something I'm abstractly concerned with/thinking about.

crates/zizmor/src/audit/timeout_minutes.rs
mchevestrier: Yes, I tried to keep the logic from #1347 (if there is no job level timeout but some step has one, flag the other steps) but we can simplify this behavior and just flag the job if any step is missing the property

mchevestrier: Done in https://github.com/zizmorcore/zizmor/pull/2166/changes/5f7924d75fd939dbec6bd5e153aeccd9d880ecb5

crates/zizmor/src/audit/timeout_minutes.rs
mchevestrier: Done in https://github.com/zizmorcore/zizmor/pull/2166/changes/2ef2494cb078dd905daa86c159dc1504d5c5a703

npapagna:
@woodruffw @mchevestrier would love to have this! (currently rolled my own check in a bash script).

What's keeping this PR from being merged?

woodruffw:
Two things:

- Review time from me
- A determination on whether this makes sense as a zizmor audit, given that it has limited security impact

It's in my queue to review, but the latter in particular means that it has less priority than other features and bugs.

PR #2179, opened 2026-07-11
fix/cleanup rainbow.svg

toy:
## Pre-submission checks

Please check these boxes:

- [ ] **Mandatory**: This PR corresponds to an issue (if not, please create one first).
  I didn't create an issue assuming the change is trivial.

- [ ] Having read the [AI policy], I hereby disclose the use of an LLM or other AI coding assistant in the creation of this PR. PRs will not be rejected for using AI tools, but *will* be rejected for undisclosed use or use that violates the policy.

## Summary

I noticed purple sticking out behind the red stripe rainbow, so checked the source and understood that the SVG has slightly offset coordinates (like `35.999` instead of `36`), uses Bézier curves instead of arcs and the purple stripe is oversized (best visible if it is ordered to be in the front rainbow).

This PR includes the cleaned up rainbow.svg with integer coordinates and using arcs for drawing stripes of reducing size.

Optimising using svgo doesn't reduce the size.


woodruffw (approved):
Thanks @toy!

toy:
@woodruffw Only after creating the PR I understood that I didn't try dividing coordinates by 3 which will reduce size a bit, I'll push as an additional commit

woodruffw:
> @woodruffw Only after creating the PR I understood that I didn't try dividing coordinates by 3 which will reduce size a bit, I'll push as an additional commit

I think that might be less compatible with browsers, e.g.:

image

toy:
> I think that might be less compatible with browsers, e.g.:

Sorry, I accidentally divided a digit in xmlns of svg making it `http://www.w1.org/2000/svg` 🤦 Unfortunately preview in my editor was ok with that

woodruffw (approved):
Thanks again @toy!

PR #2180, opened 2026-07-11
Add a warning in offline mode

### LINKED ISSUE #2178: Feature: be more explicit about not running online audits
### Pre-submission checks

Hi!
I am a very happy user of zizmor.
I would like to report a minor friction I had, and maybe an idea to improve the UX.

I use zizmor though zizmor-pre-commit, that I enforce locally and on CI (so I do not rely on [zizmor-action](https://github.com/marketplace/actions/zizmor-action), which matters for later).

For the record, the friction I am about to report is entirely my fault for not reading the docs, specially [this section](https://docs.zizmor.sh/usage/#operating-modes) about operating modes.

So basically, after seeing the `--offline` option in the help, I thought online mode was *on* by default, which is not true (zizmor needs a GitHub token for this). I also use pinact and pre-commit for other things, and those tools can use the GitHub API without input token (until you hit the rate limit), which reinforced my assumption that online checks happened by default.

So my feeling is that, one of 2 option might improve this:
* unless `--offline/--no-online-audits` is explicitly provided, zizmor should show a warning saying the online audits were not run for lack of a provided token
* add a `--online` option that trigger an error if online audits were not run

I understand zizmor-action would have solved this, as it runs online-audits by default, leveraging the default `github.token`.

### Additional context

_No response_

woodruffw:
This also contains some refactoring, moving our CLI types/helpers into `zizmor::cli`.

Closes #2178.


woodruffw:
Noting: I'm probably going to change this to some kind of `offline: Flag` state, where `Flag` is something like:

```rust
enum Flag {
  #[default]
  Default,
  Set(bool),
}
```

PR #2186, opened 2026-07-13
Include fix metadata in JSON output

### LINKED ISSUE #2175: Feature: emit some fix information in the JSON format
### Pre-submission checks

zizmor already computes concrete auto-fixes for many findings and can apply them
in-place with `--fix`. However, that fix information isn't available in any
machine-readable output:

- The SARIF output doesn't model or emit the SARIF `fix` object at all.
- The JSON v1 output has no `fixes` field.
- The only record of what a `--fix` run did is human-readable text on stderr.

This makes it hard to build automation on top of zizmor's fixes. My concrete
motivation is org-scale CI: I'd like a workflow that runs zizmor across many
repos and opens pull requests that are linked to the findings they resolve, so
engineers can just review and merge. That needs a structured, standard
representation of "which fix resolves which finding, and what does it change" —
which SARIF already defines but zizmor doesn't currently emit.

Populate the standard SARIF 2.1.0 `result.fixes` array for any finding that
carries one or more auto-fixes, within the existing `--format=sarif` output:

- `fix.description` from the fix's existing title.
- `fix.artifactChanges[].replacements[]` describing the edit, computed by
  diffing the original document against the fixed document.

Notes on scope:

- No new CLI flags; fixes just appear in existing SARIF output.
- No change to how fixes are computed or applied — `--fix` behavior is untouched.
- Deliberately kept to emitting standard SARIF only: no git, tokens, or
  PR-creation logic in the CLI. Any PR/workflow orchestration would live outside
  the CLI and consume this SARIF.

### Additional context

- This would surface the fix title that's already computed internally but
  currently unused.
- Seems complementary to the existing auto-fix roadmap (e.g. #858 and #1951).
- Emitting standard SARIF fixes also makes the data usable by any SARIF-aware
  tooling, not just my use case.
- Happy to work on this if greenlit, and to split it into small, tested PRs.
- AI assistance disclosure: I used an AI tool to help research the codebase and
  draft this issue; the proposal and wording reflect my own understanding.

david-wagih:
## Pre-submission checks

Please check these boxes:

      one first).

      use that violates the policy.

## Summary

Closes #2175 

I have added fixes array in each finding in the v1 JSON output.
have defined V1Fix which is a subset of the internal finding::Fix to only expose the fields we need like title, key, disposition.
updated the usage document with this new field added

## Test Plan
I have added new integration test called json which runs --format=json on the dependabot-execution fixture and assert that findings carries fixes list with the right title, key and description


crates/zizmor/tests/integration/json.rs
woodruffw, line 1: This should be part of the existing `json_v1` integration tests, not a new file.

docs/usage.md
```diff
@@ -265,7 +265,7 @@ For example, in the output above, `template-injection` within
 the [template-injection](./audits#template-injection) audit documentation.
 
 In addition to these OSC 8 links, `zizmor` also includes the full URL
-as part of each finding _if_ it detects a non-terminal output _or_
+as part of each finding *if* it detects a non-terminal output *or*
 a CI environment (e.g. GitHub Actions).
```
woodruffw, line 269: Can you revert these formatting changes?

docs/usage.md
```diff
@@ -411,10 +411,42 @@ zizmor --format=json . | jq .[0]
...
+    "key": {
+      "Local": {
+        "prefix": ".",
+        "given_path": "./.github/workflows/ci.yml"
+      }
+    },
+    "disposition": "unsafe"
+  }
+]
+```
```
woodruffw: I'd prefer if we remove this and just fold it into the example above.

woodruffw (commented):
Thanks @david-wagih. The approach looks good to me, just some nits about the test placement and docs.

docs/usage.md
david-wagih: on it

docs/usage.md
david-wagih: on it

crates/zizmor/tests/integration/json.rs
david-wagih: on it

docs/usage.md
```diff
@@ -411,10 +411,18 @@ zizmor --format=json . | jq .[0]
...
+      "fixes": []
     }
     ```
 
+Each finding also includes a `fixes` array of the auto-fixes available for
+it, or an empty array if none are available. Each entry has a `title`, the
+`key` of the input it applies to, and a `disposition` of either `safe` or
+`unsafe` (corresponding to the `--fix` modes). These entries are symbolic:
+they describe an available fix but do not contain concrete patch data, which
+is only produced when a fix is applied via `--fix`.
```
woodruffw: Let's just remove this. I think the JSON example above should provide the structure, i.e. give a full example instead of `fixes: []`.

(I don't want more prose here because it'll go stale over time.)

david-wagih: but i have a question here, for the example when running it, it doesn't populate any fixes automatically, so shall I switch the example to something that does ?

woodruffw: Yeah, switching to an example that does have fixes seems reasonable to me.

david-wagih: done

woodruffw (approved):
Thank you @david-wagih!

PR #2193, opened 2026-07-15
dependabot-cooldown: honor GitHub's new default

### LINKED ISSUE #2192: Tweak dependabot-cooldown based on GitHub's new defaults
All Dependabot updates now include a 3-day cooldown by default: https://github.blog/changelog/2026-07-14-dependabot-version-updates-introduce-default-package-cooldown/

Our recommendation is currently 7 days (and should probably remain that), so we need to tweak our findings slightly.

woodruffw:
~~WIP. This is not bug-free yet, since I need to fix `test_multi_ecosystem_group_without_cooldown`. The underlying problem there is that I've loosened the implicit structure in the multi-ecosystem check so now we're attempting to access YAML keys that aren't guaranteed to be present.~~

I decided to "fix" the multi-ecosystem stuff by assuming that GitHub has fixed it on their own end, given that cooldowns are now the default. If they haven't, that part of this PR will require a strategic revert.

Closes #2192.

### DIFF
--- a/crates/zizmor/src/audit/dependabot_cooldown.rs
+++ b/crates/zizmor/src/audit/dependabot_cooldown.rs
@@ -1,6 +1,7 @@
 use crate::{
     audit::{Audit, AuditError, audit_meta},
     finding::{Confidence, Fix, FixDisposition, Persona, Severity, location::Locatable as _},
+    models::dependabot,
 };
 use yamlpatch::{Op, Patch};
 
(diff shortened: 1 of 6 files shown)


hugovk:
https://docs.zizmor.sh/audits/#dependabot-cooldown needs updating about the new three-day cooldown default:

> By default, Dependabot does not perform any "cooldown" on dependency updates. [...]

Issue #1799, opened 2026-03-25
Support pre-commit configs for auditing?


woodruffw:
Raised by @henryiii.

pre-commit and prek are popular git hook managers/runners. They allow users to define a `.pre-commit-config.yaml` file that configures one or more plugins to run. For example:

```yaml
repos:
  - repo: https://github.com/henryiii/cibuildwheel
    rev: 58a0b274ea29c1e7899d45ab324b4ccdfc78d17d
    hooks:
      - id: mine
```

In the above `rev` is a Git reference, and is _also_ an impostor commit. In principle this makes it a good candidate for flagging by the `impostor-commit` audit.

Some considerations:

- Is this the only audit we could currently run on pre-commit, or are there other security audits that make sense to run?
- prek also supports `prek.toml`, which would be outside of zizmor's current scope (we're focused on YAML at the moment). This doesn't seem to be widely adopted at the moment, but if it were to be then we'd only be able to audit some portion of prek usage.
- This would be zizmor's first non-GitHub-specific audit input, which would mean a change in policy in terms of what I'm supporting. It's also a bit of a risk insofar as I know a lot less about pre-commit than I do about GitHub Actions/Dependabot.

Alternatives:

- This is possibly a better fit for a separate tool, either a pre-existing one or something new. OTOH tool proliferation is bad.

Opening this for discussion for now, since I'm not sure I want to support this (but I do agree it's a good idea). I think a significant motivator for me supporting this will be whether there are other things we can detect to improve the security of the pre-commit ecosystem, rather than just a single audit (`impostor-commit`).

henryiii:
You could check the `pre-commit autoupdate --freeze` / `prek auto-update --freeze` style (which I think Dependabot also supports, though I'm just now starting to try that out, because I want to try the 7 day cooldown, which pre-commit.ci doesn't have):

* `rev: 58a0b274ea29c1e7899d45ab324b4ccdfc78d17d # frozen: v1.2.3` - ensure tag and SHA match
* You could check for pinned hashes instead of directly using tags (`rev: v1.2.3` is valid, in fact `rev: main` is valid but strongly discouraged, pre-commit caches environments on this value, so it's assumed to be fixed)

henryiii:
Another thought if you were looking for more ideas - I just noticed superfluous-actions (from #1800), and one thing that could be added is renamed/replaced checks. (I already do this in sp-repo-review) For example, the ones there:
 
[...]

PR #2209, opened 2026-07-20
Initial pre-commit modeling

woodruffw:
See #1799 for context.

This does not add any audits yet; this is just to ensure we can model and inject pre-commit inputs into zizmor's machinery.

### DIFF
--- a/crates/zizmor/src/models/pre_commit.rs
+++ b/crates/zizmor/src/models/pre_commit.rs
@@ -1,0 +1,144 @@
+//! Pre-commit models.
+//!
+//! These models enrich the models under [`pre_commit_models`],
+//! providing higher-level APIs for zizmor to use.
+
+use terminal_link::Link;
+
+use crate::{
+    finding::location::{SymbolicFeature, SymbolicLocation},
+    models::AsDocument,
+    registry::input::{CollectionError, InputKey},
+};
+
+pub(crate) struct PreCommitConfig {
+    pub(crate) key: InputKey,
+    pub(crate) link: Option<String>,
+    document: yamlpath::Document,
+    inner: pre_commit_models::config::Config,
+}
+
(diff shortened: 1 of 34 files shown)


crates/zizmor/src/models/pre_commit.rs
```diff
@@ -0,0 +1,144 @@
...
+impl std::fmt::Debug for PreCommitConfig {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{key}", key = self.key)
+    }
+}
+
+impl PreCommitConfig {
+    pub(crate) fn from_string(contents: String, key: InputKey) -> Result<Self, CollectionError> {
+        // TODO: `from_str_with_validation` here.
+        let inner = yaml_serde::from_str(&contents)?;
```
woodruffw, line 44: This would be easy enough to stub-in with SchemaStore like we do for other inputs.

https://www.schemastore.org/pre-commit-config.json

That'll be a follow-up PR.

crates/zizmor/src/models/pre_commit.rs
```diff
@@ -0,0 +1,144 @@
...
+impl std::fmt::Debug for PreCommitHooks {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{key}", key = self.key)
+    }
+}
+
+impl PreCommitHooks {
+    pub(crate) fn from_string(contents: String, key: InputKey) -> Result<Self, CollectionError> {
+        // TODO: `from_str_with_validation` here.
+        let inner = yaml_serde::from_str(&contents)?;
```
woodruffw, line 110: Same as above:

https://www.schemastore.org/pre-commit-hooks.json

Issue #2218, opened 2026-07-22
[BUG]: "input does not match expected validation schema" for avm/pull_request.yaml


y-guyon:
### Pre-submission checks

- [x] I am **not** filing a feature request. These should be filed via the feature request form instead.
- [x] I have checked the [Troubleshooting Guide](https://docs.zizmor.sh/troubleshooting) for my problem.
- [x] I have looked through both the [open](https://github.com/zizmorcore/zizmor/issues?q=is%3Aissue+is%3Aopen+) and [closed](https://github.com/zizmorcore/zizmor/issues?q=is%3Aissue%20state%3Aclosed) issues for a duplicate report.
- [x] I affirm that I have followed the [AI policy](https://github.com/zizmorcore/.github/blob/main/AI_POLICY.md) and that this report constitutes my own words and understanding. 

### zizmor version

v1.25.2

### Expected behavior

I ran Zizmor on the avm repository. It worked for most workflows.

### Actual behavior

It failed for two workflows with the error "input does not match expected validation schema", see https://github.com/AOMediaCodec/avm/pull/5079.

How can I find out which part of the input Zizmor does not support?

### Reproduction steps

```
git clone https://github.com/AOMediaCodec/avm.git
zizmor --fix=all --gh-token=$(gh auth token) avm/.github/workflows
```

### Logs

```plain text
zizmor --fix=all --gh-token=$(gh auth token) --verbose .github/workflows/nightly.yaml

 INFO zizmor: \U0001f308 zizmor v1.25.2
DEBUG rustls_platform_verifier::verification::others: Loaded 153 CA root certificates from the system
DEBUG rustls_platform_verifier::verification::others: Loaded 153 CA root certificates from the system
DEBUG collect_inputs: zizmor::config: discovering config for local input `.github/workflows/nightly.yaml`
DEBUG collect_inputs: zizmor::config: attempting config discovery in `.github/workflows`
DEBUG collect_inputs: zizmor::config: found `/usr/local/google/home/yguyon/git/avm/.git`, stopping search
DEBUG collect_inputs: zizmor::registry::input: registering workflow input as with key file://.github/workflows/nightly.yaml
 WARN collect_inputs: zizmor::registry::input: failed to validate file://.github/workflows/nightly.yaml as workflow: input does not match expected validation schema
fatal: no audit was performed
error: no inputs collected
  |
  = help: collection yielded no auditable inputs
  = help: inputs must contain at least one valid workflow, action, or Dependabot config

[...]

Issue #1116, opened 2025-08-29
[BUG]: zizmor fails to scan/parse jfrog/setup-jfrog-cli


pyun96:
### Pre-submission checks

- [x] I am **not** filing a feature request. These should be filed via the feature request form instead.
- [x] I have looked through both the [open](https://github.com/zizmorcore/zizmor/issues?q=is%3Aissue+is%3Aopen+) and [closed](https://github.com/zizmorcore/zizmor/issues?q=is%3Aissue%20state%3Aclosed) issues for a duplicate report.

### zizmor version

1.12.1

### Expected behavior

I expected zizmor to be able to scan https://github.com/jfrog/setup-jfrog-cli and provide a detailed summary of the audit rules it ran against the action. Additionally, I expected zizmor to be able to fail gracefully and be able to continue scanning other actions that are queued up when passed to it as an input.

### Actual behavior

zizmor is met with an error and fails to continue execution.

### Reproduction steps

1. zizmor jfrog/setup-jfrog-cli
2. Observe the error.

### Logs

```plain text
zizmor -v jfrog/setup-jfrog-cli 
DEBUG zizmor::config: no config discovered; loading default
DEBUG zizmor::config: loaded config: Config { rules: {} }
DEBUG zizmor::state: using cache directory: "/Users/paul.yun/.cache/zizmor"
DEBUG collect_inputs:fetch_audit_inputs{slug=RepositoryUses { owner: "jfrog", repo: "setup-jfrog-cli", subpath: None, git_ref: None }}: zizmor::github_api: fetching repo: https://api.github.com/repos/jfrog/setup-jfrog-cli/tarball/HEAD
DEBUG collect_inputs:fetch_audit_inputs{slug=RepositoryUses { owner: "jfrog", repo: "setup-jfrog-cli", subpath: None, git_ref: None }}: reqwest::connect: starting new connection: https://api.github.com/    
DEBUG collect_inputs:fetch_audit_inputs{slug=RepositoryUses { owner: "jfrog", repo: "setup-jfrog-cli", subpath: None, git_ref: None }}: hyper_util::client::legacy::connect::http: connecting to 140.82.114.6:443
DEBUG collect_inputs:fetch_audit_inputs{slug=RepositoryUses { owner: "jfrog", repo: "setup-jfrog-cli", subpath: None, git_ref: None }}: hyper_util::client::legacy::connect::http: connected to 140.82.114.6:443
[...]

PR #2220, opened 2026-07-22
Handle `jobs.*.outputs` literals gracefully

woodruffw:
This fixes the first part of #2218, but reveals a much more surprising bug:

```
 INFO zizmor: 🌈 zizmor v1.28.0
fatal: no audit was performed
error: failed to load https://github.com/AOMediaCodec/avm/blob/HEAD/.github/workflows/pull_request.yaml as workflow
  |

Caused by:
    0: failed to load https://github.com/AOMediaCodec/avm/blob/HEAD/.github/workflows/pull_request.yaml as workflow
    1: failed to load internal pathing document
    2: input is not valid YAML
```

That suggests a `yamlpath` or `tree-sitter-yaml` bug, probably the latter.

### DIFF
--- a/crates/github-actions-models/src/workflow/job.rs
+++ b/crates/github-actions-models/src/workflow/job.rs
@@ -24,7 +24,7 @@ pub struct NormalJob {
     pub environment: Option<DeploymentEnvironment>,
     pub concurrency: Option<Concurrency>,
     #[serde(default)]
-    pub outputs: IndexMap<String, String>,
+    pub outputs: Env,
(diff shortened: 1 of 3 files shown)


woodruffw:
Ah, that looks like another variant of https://github.com/zizmorcore/zizmor/issues/1116 -- this was fixed by https://github.com/tree-sitter-grammars/tree-sitter-yaml/pull/36 but hasn't been put into a `tree-sitter-yaml` release yet.

Issue #2139, opened 2026-06-21
[BUG]: Failing test in 1.26.1: e2e::issue_1745


kpcyrd:
### Pre-submission checks

- [x] I am **not** filing a feature request. These should be filed via the feature request form instead.
- [x] I have checked the [Troubleshooting Guide](https://docs.zizmor.sh/troubleshooting) for my problem.
- [x] I have looked through both the [open](https://github.com/zizmorcore/zizmor/issues?q=is%3Aissue+is%3Aopen+) and [closed](https://github.com/zizmorcore/zizmor/issues?q=is%3Aissue%20state%3Aclosed) issues for a duplicate report.
- [x] I affirm that I have followed the [AI policy](https://github.com/zizmorcore/.github/blob/main/AI_POLICY.md) and that this report constitutes my own words and understanding.

### zizmor version

1.26.1

### Expected behavior

`cargo test` works

### Actual behavior

```
---- e2e::issue_1745 stdout ----
━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ Snapshot Summary ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
Snapshot: issue_1745
Source: crates/zizmor/tests/integration/e2e.rs:640
────────────────────────────────────────────────────────────────────────────────
Expression: zizmor()
    .offline(true)
    .no_config(true)
    .working_dir(input_under_test("issue-1745-repro"))
    .input(".github")
    .args(["--format=github"])
    .run()?
────────────────────────────────────────────────────────────────────────────────
-old snapshot
+new results
────────────┬───────────────────────────────────────────────────────────────────
    1       │-::warning file=@@WORKING_DIR@@/@@TEST_PREFIX@@/issue-1745-repro/@@INPUT@@/workflows/test.yml,line=10,title=artipacked::test.yml:10: credential persistence through GitHub Actions artifacts: does not set persist-credentials: false
    2       │-::error file=@@WORKING_DIR@@/@@TEST_PREFIX@@/issue-1745-repro/@@INPUT@@/workflows/test.yml,line=10,title=unpinned-uses::test.yml:10: unpinned action reference: action is not pinned to a hash (required by blanket policy)
          1 │+::warning file=workflows/test.yml,line=10,title=artipacked::test.yml:10: credential persistence through GitHub Actions artifacts: does not set persist-credentials: false
[...]

PR #2234, opened 2026-07-26
Make config discovery root-aware

### LINKED ISSUE #2229: [BUG]: zizmor.yaml (or other config names) is not found when the repo is called "workflows"
### Pre-submission checks

### zizmor version

1.28.0

### Expected behavior

Configuration should be loaded

### Actual behavior

Configuration should is not loaded:

```bash
$ ls -al zizmor.yaml
.rw-r--r--@ 116 xxx 24 Juli 11:48 zizmor.yaml
$ zizmor -v . 2>&1 | grep config
DEBUG collect_inputs: zizmor::config: discovering config for local input `.`
DEBUG collect_inputs: zizmor::config: attempting config discovery in `.`
DEBUG collect_inputs: zizmor::config: reached filesystem root without finding a config
```

### Reproduction steps

1. mkdir workflows
2. cd workflows
3. git init
4. touch zizmor.yaml
5. zizmor .

### Logs

```plain text
see above.
```

### Additional context

The repo is called `workflows` since we store shared workflows there...

woodruffw:
~~WIP.~~ Fixes #2229.

TL;DR: This takes the repository root that we discover for path presentation (see #2095) and uses it as another hint during config discovery. This is a lot more reliable than our previous heuristics, which involved trying to figure out where we were in some directory tree.

(Like with path presentation, the previous heuristics remain for when we don't have a Git repo to root against. This can happen e.g. if the user runs `zizmor .` on an extracted archive rather than a real repository.)

~~Needs more tests, plus another tweak to the heuristics to ensure we treat `.github/workflows` correctly.~~

Another perk of this is that it adds a proper "workspace" testing abstraction, which will make it possible for me to patch the downstream testing pain experienced in #2139.

### DIFF
--- a/crates/zizmor/tests/integration/common.rs
+++ b/crates/zizmor/tests/integration/common.rs
@@ -35,6 +35,19 @@ static TEST_PREFIX: LazyLock<Utf8PathBuf> = LazyLock::new(|| {
     Utf8PathBuf::try_from(file_path).expect("Cannot create UTF-8 path from test data directory")
 });
 
+const CONFIG_PLACEHOLDER: &str = "@@CONFIG@@";
+const INPUT_PLACEHOLDER: &str = "@@INPUT@@";
+const REPO_ROOT_PLACEHOLDER: &str = "@@REPO_ROOT@@";
+const WORKING_DIR_PLACEHOLDER: &str = "@@WORKING_DIR@@";
+const TEST_PREFIX_PLACEHOLDER: &str = "@@TEST_PREFIX@@";
+const VERSION_PLACEHOLDER: &str = "@@VERSION@@";
+
+/// Matches any `@@PLACEHOLDER@@` — including the custom ones introduced by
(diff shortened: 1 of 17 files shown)


crates/zizmor/tests/integration/common.rs
```diff
@@ -437,3 +463,93 @@ fn redact(haystack: &mut String, needle: &Utf8Path, placeholder: &str) {
...
+    ///
+    /// `dest` will be joined to the workspace's root.
+    pub fn copy<'a>(&self, source: impl Into<&'a Utf8Path>, dest: impl Into<&'a Utf8Path>) {
+        let source = source.into();
+        let dest = self.path().join(dest.into());
+
+        if source.is_file() {
+            fs::copy(source, &dest).unwrap();
+        } else {
+            dircpy::copy_dir(source, &dest).unwrap();
```
woodruffw: Note: I'm going to remove this dep, it adds a lot of transitive sprawl for a ~30 line open-coded operation.

PR #2247, opened 2026-07-28
Suggest tags belonging to the action being used

### LINKED ISSUE #2243: [BUG]:  The ref-version-mismatch suggests the wrong action's tag for subdirectory actions
### Pre-submission checks

### zizmor version

zizmor 1.28.0

### Expected behavior

For apache/infrastructure-actions/stash/save@<sha> (see reproduction below) a tag suggestion should name that action and not other actions which tags point to the smae commit: save/v1.0.0.

### Actual behavior

In a repository that publishes several actions from subdirectories and tags each one under its own prefix, `ref-version-mismatch`
 (and the other audits that suggest a tag) can suggest — and auto-fix to — a *different* action's version.

The `apache/infrastructure-actions` repository recently adopted exactly this layout in [#1032](https://github.com/apache/infrastructure-actions/
pull/1032): every action is released under its leaf-directory prefix, so one commit carries `allowlist-check/v1`, `allowlist-chec
k/v1.0.0`, `restore/v1`, `restore/v1.0.0`, `save/v1` and `save/v1.0.0` and for a while both dependabot and zizmor respect this format - the version regexp validates it and zizmor does not complain when the version is "right" and points to the right tag.

In the case explained below - zizmor suggests allowlist-check/v1.0.0, a different action that shares the commit, and offers it as an auto-fix. apache/infrastructure-actions tags each action under its leaf prefix (their #1032 (https://github.com/apache/infrastructure-actions/pull/1032)), so one commit carries allowlist-check/v1{,.0.0}, restore/v1{,.0.0}, save/v1{,.0.0}.

### Reproduction steps

The workflow I used:

```
name: control
on: push
permissions: {}
jobs:
  t:
    runs-on: ubuntu-latest
    steps:
      - uses: apache/infrastructure-actions/stash/save@61dcea11f19e2bbe1263f14d72235e8da17d3ad0  # save/v9.9.9
```

### Logs

```plain text
warning[ref-version-mismatch]: action's hash pin has mismatched or missing version comment
  - uses: apache/infrastructure-actions/stash/save@61dcea11…  # save/v9.9.9
    is pointed to by tag allowlist-check/v1.0.0                 ^^^^^^^^^^^ points to unknown ref
```

### Additional context

Cause is `Client::longest_tag_for_commit` (crates/zizmor/src/github.rs), which does .max_by_key(|t| t.name.len()) over every tag at the commit and is never told the action's subpath; RepositoryUses::subpath() already exists and none of the 6 call sites pass it. Affects ref-version-mismatch, unpinned-uses, known-vulnerable-actions, stale-action-refs, artipacked. 

Worth stating explicitly that validation is fine — commit_for_ref matches tag names exactly and peels annotated tags, so a correct # save/v1.0.0 comment produces no finding; only the suggestion side is wrong.

While - of course, this is mostly a convention - I think it's worth assuming that subpath match is the "preferred" match for the suggestion, not the longest match.

potiuk:
A repository that ships several actions from subdirectories tags each one under its own prefix, so a single commit can carry save/v1.0.0, restore/v1.0.0 and allowlist-check/v1.0.0 at once. Choosing purely by length hands every action in such a repository whichever sibling happens to have the longest name, and offers that as an auto-fix, so accepting it writes another action's version into the workflow.

Prefer a tag that names the action being used, keep the previous choice for everything else, and fall back to it when nothing names the action.

## Pre-submission checks

Please check these boxes:

      one first).

      use that violates the policy.

Disclosure: Generated with help of LLM.

## Summary

Closes https://github.com/zizmorcore/zizmor/issues/2243

In a repository that ships several actions from subdirectories and tags each under its own prefix, when Zizmor suggests auto-fix for actions, the longest_tag_for_commit picked the longest tag name among every tag at a commit, without knowing which action inside the repository was being used. , one commit carries several tags, so every action there was told about whichever sibling had the longest name — and that was offered as an auto-fix.

Concretely, for https://github.com/apache/infrastructure-actions/commit/61dcea11…, whose commit also carries restore/* and allowlist-check/* tags:

is pointed to by tag allowlist-check/v1.0.0
where the answer should be save/v1.0.0.

With this fix - the choice prefers a tag that names the action being used:

with a subpath, tags prefixed by the full subpath (stash/save/) or its leaf (save/);
without one, tags containing no /, so a root action does not use subdirectory action's tag;
otherwise the previous longest-tag-at-this-commit behaviour, unchanged, so repositories that do not use prefixed tags behave exactly as before — including the fallback where nothing names the action and a sibling is still better than no suggestion.
Validation was already correct and is untouched: commit_for_ref matches tag names exactly and peels annotated tags, so a correct # save/v1.0.0 comment resolves to the pinned commit and produces no finding. Only the suggestion side was **wrong.**

## Test Plan

Added unit tests covering the pure function that resolves best match.

Note for review: 7 tests in known_vulnerable_actions fail on my machine both with and without this change (108 passed / 7 failed on a clean tree, 109 / 7 with it — the extra pass being the new test), so I have assumed they are environmental rather than related.

### DIFF
--- a/crates/zizmor/src/audit/artipacked.rs
+++ b/crates/zizmor/src/audit/artipacked.rs
@@ -50,7 +50,7 @@ impl Artipacked {
             match self.client {
                 Some(ref client) => {
                     let tag = client
-                        .longest_tag_for_commit(&uses.into(), uses.git_ref())
+                        .longest_tag_for_commit(&uses.into(), uses.subpath(), uses.git_ref())
                         .await?;
 
                     match tag {
--- a/crates/zizmor/src/audit/artipacked.rs
+++ b/crates/zizmor/src/audit/artipacked.rs
@@ -51,7 +51,12 @@ impl Artipacked {
             match self.client {
                 Some(ref client) => {
                     let tag = client
-                        .longest_tag_for_commit(uses.owner(), uses.repo(), uses.git_ref())
+                        .longest_tag_for_commit(
+                            uses.owner(),
+                            uses.repo(),
+                            uses.subpath(),
+                            uses.git_ref(),
+                        )
                         .await?;
 
                     match tag {
--- a/crates/zizmor/src/audit/known_vulnerable_actions.rs
+++ b/crates/zizmor/src/audit/known_vulnerable_actions.rs
@@ -65,7 +65,7 @@ impl KnownVulnerableActions {
 
                 match self
                     .client
-                    .longest_tag_for_commit(uses.owner(), uses.repo(), &commit_ref)
+                    .longest_tag_for_commit(uses.owner(), uses.repo(), uses.subpath(), &commit_ref)
                     .await
                     .map_err(Self::err)?
                 {
@@ -84,7 +84,7 @@ impl KnownVulnerableActions {
             commit_ref => {
                 match self
                     .client
-                    .longest_tag_for_commit(uses.owner(), uses.repo(), commit_ref)
+                    .longest_tag_for_commit(uses.owner(), uses.repo(), uses.subpath(), commit_ref)
                     .await
                     .map_err(Self::err)?
                 {
--- a/crates/zizmor/src/audit/known_vulnerable_actions.rs
+++ b/crates/zizmor/src/audit/known_vulnerable_actions.rs
@@ -73,7 +73,7 @@ impl KnownVulnerableActions {
 
                 match self
                     .client
-                    .longest_tag_for_commit(&slug, commit_ref.commit())
+                    .longest_tag_for_commit(&slug, repo_ref.subpath(), commit_ref.commit())
                     .await
                     .map_err(Self::err)?
                 {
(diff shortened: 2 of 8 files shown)


potiuk:
A bit of a comment - It took me 4 times to create the PR - besides the first time (3 times from the template) - all the previous attemps were auto-closed as not following the template (but ... I did). 

Yes. It was generated with LLM (Disclosed) - but... for whatever reason it was auto closed few times before.. The 5th time succeeded.

woodruffw:
From looking at the closed PRs, it looks like none of them have the hidden marker that the bot checks for. Did you fill out the template on the web, or through some other mechanism? If you copy-pasted it that might explain it.

(This is a line I'm still figuring out how to draw -- I don't mind people submitting LLM-assisted PRs, but I _do_ want them to submit the PR itself manually and to fully affirm that they're read the template. The bot is my best attempt at that ATM, but it's definitely not perfect.)

woodruffw (approved):
Thanks @potiuk! The basic approach here seems reasonable to me, although I think it's unfortunate that repos use this pattern at all 😅 -- FWICT there's going to be a lot of edge cases with this, and I think it might also not interact super well with GitHub's upcoming actions lockfile feature.

One request: could you add a real-world (online) test that exercises this path? I think something under `ref-version-mismatch` would be fine.

woodruffw:
(NB: I'm probably going to queue this for 1.30, since I'm planning on sending 1.29 out today.)

woodruffw (approved):
Thank you @potiuk!

PR #2249, opened 2026-07-31
Support self-repository syntax for local uses

### LINKED ISSUE #2248: Feature: Support Self-Repository Syntax For Actions
### Pre-submission checks

GitHub Actions now [supports](https://github.blog/changelog/2026-07-30-reference-same-repository-actions-with-self-repository-syntax/) a dedicated syntax for sibling actions within a repository ([docs](https://docs.github.com/en/actions/how-tos/write-workflows/choose-what-workflows-do/find-and-customize-actions#adding-an-action-from-the-same-repository)).  References like `uses: my_org/my_repo/actions/foo@v1` can be replaced with `uses: $/./actions/foo`, avoiding the need to keep the internal references consistent and making it easier to work on one action and test it in another.  Currently an action with that syntax fails with:

```
ERROR collect_inputs: github_actions_models::common: msg="malformed `uses` ref: missing `@<ref>` in $/git-sign"
fatal: no audit was performed
error: failed to load file://./tag-version/action.yml as action
  |
Caused by:
    0: failed to load file://./tag-version/action.yml as action
    2: data did not match any variant of untagged enum Runs
```

Add support for the new self-reference syntax.  

### Additional context

_No response_

woodruffw:
Closes #2248.

### DIFF
--- a/crates/github-actions-models/src/common.rs
+++ b/crates/github-actions-models/src/common.rs
@@ -289,7 +289,7 @@ impl Uses {
         let uses = uses.into();
         let uses = uses.trim();
 
-        if uses.starts_with("./") || uses.starts_with('$') {
+        if uses.starts_with("./") || uses.starts_with("$/") {
             Ok(Self::Local(LocalUses::new(uses)))
         } else if let Some(image) = uses.strip_prefix("docker://") {
             Ok(Self::Docker(DockerUses::parse(image)))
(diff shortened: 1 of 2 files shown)


crates/github-actions-models/src/common.rs
```diff
@@ -289,7 +289,7 @@ impl Uses {
         let uses = uses.into();
         let uses = uses.trim();
 
-        if uses.starts_with("./") {
+        if uses.starts_with("./") || uses.starts_with('$') {
```
woodruffw: I guess this technically overaccepts, since `uses: $` is not valid.

PR #2255, opened 2026-08-02
Generalize repo ref handling

woodruffw:
~~WIP.~~

The idea here is to generalize how we handle "repository references" away from assuming that they're `uses:`-shaped, so that the common machinery in some of the audits can be re-used for pre-commit repo checks.

Adds `impostor-commit` analysis for pre-commit configs as a demonstration.

See #2256

### DIFF
--- a/crates/zizmor/src/models/repo_ref.rs
+++ b/crates/zizmor/src/models/repo_ref.rs
@@ -97,7 +97,7 @@ impl<'doc> RepoRef<'doc> {
             owner: extra.map(|e| e.0),
             repo: extra.map(|e| e.1),
             slug: extra.map(|e| e.2),
-            git_ref: git_ref,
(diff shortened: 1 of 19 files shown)


crates/zizmor/src/models/repo_ref.rs
```diff
@@ -0,0 +1,256 @@
...
+}
+
+impl std::fmt::Display for Slug<'_> {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{}", self.slug())
+    }
+}
+
+#[derive(Debug)]
+pub(crate) enum RepoRef<'doc> {
```
woodruffw, line 42: I don't love this name, but it's fine for now.

Issue #2256, opened 2026-08-02
[META] pre-commit audit parity


woodruffw:
From https://github.com/zizmorcore/zizmor/issues/1799#issuecomment-5023653606:

- [x] `impostor-commit`: #2255 
- [ ] `unpinned-uses`
- [x] `forbidden-uses`: #2263
- [x] `archived-uses`: #2272
- [ ] `typosquat-uses`
- [x] `ref-confusion`: https://github.com/zizmorcore/zizmor/pull/2274
- [ ] `ref-version-mismatch`
- [ ] `known-vulnerable-actions`
  - This one may make sense to rename to `known-vulnerabilities`
- [ ] `misfeature` (TBD: how?)
- [ ] `obfuscation` (TBD: how?)
- New audits:
  - [ ] Checking for `additional_dependencies`?
  - [ ] Checking for EOL Python and other versions?
  - [ ] Checking for ambiguous Git refs, e.g. short Git hashes?

pygarap:
@woodruffw Could you add support for `prek.toml` as part of this checklist? Thanks 🙏

woodruffw:
> [@woodruffw](https://github.com/woodruffw) Could you add support for `prek.toml` as part of this checklist? Thanks 🙏

The tricky part here is that zizmor doesn't actually support TOML inputs at all at the moment 😅 -- the internals (turning symbolic paths into source spans) are 100% YAML-specific.

That's fixable over time, but it's a much larger task than reaching audit parity for YAML-shaped inputs. I'll create a separate issue to track that with some rough notes on what it'll take.

shaanmajid:
`ref-version-mismatch` also seems like it'd be a pretty clean fit -- pre-commit and prek both emit and bump version specifier comments with their native  `auto-update --freeze` command. This version specifier takes the format `# frozen: <version>` ([example](https://github.com/j178/prek/blob/master/.pre-commit-config.yaml#L21)), though, so the audit would need to be expanded to recognize and parse this format (optionally this format could narrowly _only_ be accepted for the pre-commit config input type, which may be considered more correct).

woodruffw:
Yep! I'll add to the list above.

woodruffw:
@shaanmajid @reubenwong97 let me know if you're interested in working on any of these, BTW (and no pressure of course) -- the APIs/degree of sharing between the different input types is still pretty rough, so I'd appreciate others' opinions/eyes.

shaanmajid:
[...]

PR #2272, opened 2026-08-06
feat(archived-uses): add support for auditing pre-commit-config.yaml

reubenwong97:
## Pre-submission checks

Please check these boxes:

- [x] **Mandatory**: This PR is part of #2256 .

      use that violates the policy.

## Summary

Adapts the `archived-uses` audit for auditing `.pre-commit-config.yaml` files.

## Test Plan

An integration test was added on an archived repo, [facebook/pyre-check](github.com/facebook/pyre-check). The archived repo generation script was updated to include another source (though it only has 1 file just for validating this idea).


reubenwong97:
Hi @woodruffw , would you mind taking a look to see if this is reasonable? If the idea works out, then the main thing is that I am not too sure how to curate the list for archived pre-commit repos...

woodruffw:
Thanks @reubenwong97, I'll try and take a look at this tonight. 

And yeah, seeding a list of archived pre-commit repos is going to be interesting. One idea: if there's a good dataset of repos that *use* pre-commit somewhere, we could work backwards from that (using GitHub's API to test for archivedness). I think that would also generalize well for archived actions too.

support/archived-pre-commit-repos.txt
```diff
@@ -0,0 +1,7 @@
+# archived-action-repos.txt
```
woodruffw: Filename here needs updating 🙂

reubenwong97:
> Thanks @reubenwong97, I'll try and take a look at this tonight.
> 
> And yeah, seeding a list of archived pre-commit repos is going to be interesting. One idea: if there's a good dataset of repos that _use_ pre-commit somewhere, we could work backwards from that (using GitHub's API to test for archivedness). I think that would also generalize well for archived actions too.

Hi @woodruffw , I wasn't able to find such a dataset.. However, I think a decent list could be seeded with some repos indexed by sourcegraph with [this query](https://sourcegraph.com/search?q=file:%5E%5C.pre-commit-hooks%5C.yaml%24+archived:only+count:all+select:repo+fork:yes&patternType=keyword&sm=0). Does this look ok to you?

support/archived-repos.py
```diff
@@ -7,20 +7,24 @@
...
 
 def main() -> None:
     lines = []
-    for line in _ARCHIVED_ACTION_REPOS.read_text().splitlines():
-        line = line.strip()
-        if not line or line.startswith("#"):
-            continue
-
-        lines.append(line.lower())
+    sources = [_ARCHIVED_ACTION_REPOS, _ARCHIVED_PRE_COMMIT_REPOS]
```
woodruffw, line 20: At some point we might want to split these two out (so that they become two distinct FSTs), since I expect there isn't a ton of overlap. But I think it's fine for now (and actually I'm not sure it would even be faster, I guess the search is independent of FST size).

woodruffw (approved):
Thanks @reubenwong97!

And yeah, that Sourcegraph query looks reasonable. We can keep seeding this over time, this seems like a solid starting point.

PR #2295, opened 2026-08-10
fix(yamlpatch): preserve nesting in Add and Replace

### LINKED ISSUE #2270: [BUG]: `Op::Add`/`Op::Replace` corrupt nested blocks
### Pre-submission checks

### zizmor version

1.29.0

### Expected behavior

yamlpatch should be able to parse nested blocks.  There are two unexpected behaviours:

`Op::Add`, **expected**: 

```yaml
top:
  existing: 1
  added:
    outer:
      inner: value
```

`Op::Add` **Actual**:

```yaml
top:
  existing: 1
  added:
    outer:
    inner: value
```

`Op::Replace`:

For a scalar replacement that is correct. For a block-mapping replacement it places a multi-line mapping on the same line as its key, yielding `target: outer:`, which is not valid YAML, and fails. The expected output is:

```yaml
top:
  target:
    outer:
      inner: value
```

### Actual behavior

Patch operations that insert a **mapping value nested more than one level deep** do not round-trip. `Op::Add` (and `Op::MergeInto`, which lowers to it) produces well-formed YAML that parses to a *different document* than the one
requested, silently. 

On the other hand,`Op::Replace` produces invalid YAML and fails.

### Reproduction steps

```rust
#[test]
fn test_add_nested_block_mapping() {
    let original = r#"
top:
  existing: 1
"#;

    let document = yamlpath::Document::new(original).unwrap();

    // A value nested two levels deep.
    let value: yaml_serde::Value = yaml_serde::from_str("outer:\n  inner: value\n").unwrap();

    let operations = vec![Patch {
        route: route!("top"),
        operation: Op::Add {
            key: "added".to_string(),
            value,
        },
    }];

    let result = apply_yaml_patches(&document, &operations).unwrap();

    // This assertion fails: `inner` is emitted as a sibling of `outer`
    // rather than as its child.
    insta::assert_snapshot!(format_patch(result.source()), @r"
    --- PATCH ---

    top:
      existing: 1
      added:
        outer:
          inner: value

    --- END PATCH ---
    ");
}

#[test]
fn test_replace_nested_block_mapping() {
    let original = r#"
top:
  target:
    old: 1
"#;

    let document = yamlpath::Document::new(original).unwrap();

    // Control: replacing the same route with a *scalar* works, confirming
    // the route itself is correct.
    let scalar_ops = vec![Patch {
        route: route!("top", "target"),
        operation: Op::Replace(yaml_serde::Value::String("replaced".to_string())),
    }];
    let scalar_result = apply_yaml_patches(&document, &scalar_ops);
    assert!(
        scalar_result.is_ok(),
        "control case should succeed: {:?}",
        scalar_result.err()
    );

    // The same route with a nested mapping value fails.
    let value: yaml_serde::Value = yaml_serde::from_str("outer:\n  inner: value\n").unwrap();
    let operations = vec![Patch {
        route: route!("top", "target"),
        operation: Op::Replace(value),
    }];

    // This fails: the serialized mapping is spliced onto the key's own line,
    // producing the invalid intermediate `target: outer:`, which then fails
    // to re-parse.
    let result = apply_yaml_patches(&document, &operations)
        .expect("Op::Replace with a block mapping value should succeed");

    insta::assert_snapshot!(format_patch(result.source()), @r"
    --- PATCH ---

    top:
      target:
        outer:
          inner: value

    --- END PATCH ---
    ");
}
```

### Logs

```plain text
...
```

### Additional context

Suggested fix for the first case.
`handle_block_mapping_addition`, `src/lib.rs:838-847`:

```diff
let value_lines = new_value_str.lines();
let mut result = format!("\n{indent}{key}:");
for line in value_lines {
    if !line.trim().is_empty() {
        result.push('\n');
        result.push_str(&indent);
        result.push_str("  "); // 2 spaces for nested content
-        result.push_str(line.trim_start());   // <-- drops the line's own depth
+       result.push_str(line);
    }
}
```

dmbuil:
## Pre-submission checks

Please check these boxes:

      one first).

      use that violates the policy.

## Summary

Both operations assumed a serialized value is a single-line scalar, so mapping values nested more than one level deep did not round-trip.

`serialize_yaml_value` anchors its output at column 0, which means each line's leading whitespace is its depth relative to the value's root.

In `handle_block_mapping_addition`, `trim_start()` discarded that depth, flattening every nested level into a sibling of the top-level key. The result was well-formed YAML that parsed to a different document than the one requested, silently.

In `apply_value_replacement`, the value was unconditionally formatted as `key: value` on one line. A block collection was therefore spliced onto the key's own line, leaving the remaining lines at the key's depth and producing invalid YAML such as `target: outer: `, which failed to re-parse.

Sequences are covered alongside mappings, since a multi-line block sequence hits the same splice in the replacement path.

Closes  #2270 

Claude was used to help me review the tests added.

## Test Plan

- `test_add_nested_block_mapping`: `Op::Add` with a value nested two levels deep. Previously produced well-formed YAML that parsed to a *different* document (`inner` emitted as a sibling of `outer` instead of its child).
- `test_replace_nested_block_mapping`: `Op::Replace` with a block-mapping value. Previously emitted `target: outer:` and failed to re-parse. Includes a scalar-replacement control case on the same route, so a failure can't be mistaken for a bad route.
- `test_replace_block_sequence`: the `Value::Sequence` arm of the replacement fix; a multi-line block sequence hits the same splice as a mapping.
- `test_replace_deeply_nested_block_mapping`: four levels deep, confirming each level keeps its own relative depth rather than only the first.
- `test_merge_into_nested_block_mapping`: `Op::MergeInto` lowers to `Op::Add`, so it was subject to the same flattening; nothing covered that path.
- `test_add_nested_block_mapping_preserves_comments`: asserts the fix doesn't regress comment preservation, pinning top-of-file, entry-attached, trailing-inline and end-of-file comments alongside the corrected nesting.

### DIFF
--- a/crates/yamlpatch/src/lib.rs
+++ b/crates/yamlpatch/src/lib.rs
@@ -1084,16 +1084,17 @@ fn apply_value_replacement(
         // Regular block style - use standard formatting
         let val_str = serialize_yaml_value(value)?;
 
-        // A multi-line block collection can't share the key's line: splicing it
-        // after `key:` puts the collection's first entry on that line and leaves
-        // the remaining lines at the key's own depth, e.g. `target: outer:`,
-        // which is not valid YAML. Emit it as a nested block instead.
-        if val_str.contains('\n')
-            && matches!(
-                value,
-                yaml_serde::Value::Mapping(_) | yaml_serde::Value::Sequence(_)
-            )
-        {
+        // A non-empty block collection can't share the key's line: splicing it
+        // after `key:` produces invalid YAML even when its serialized form is a
+        // single line, e.g. `target: outer: value` or `target: - value`. Emit it
+        // as a nested block instead, while keeping empty collections inline.
+        let is_nonempty_block_collection = match value {
+            yaml_serde::Value::Mapping(mapping) => !mapping.is_empty(),
+            yaml_serde::Value::Sequence(sequence) => !sequence.is_empty(),
+            _ => false,
+        };
+
+        if is_nonempty_block_collection {
             let key_indent = &key_part[..key_part.len() - key_part.trim_start().len()];
             let mut result = key_part.to_string();
             for line in val_str.lines() {
--- a/crates/yamlpatch/src/lib.rs
+++ b/crates/yamlpatch/src/lib.rs
@@ -842,7 +842,11 @@ fn handle_block_mapping_addition(
                     result.push('\n');
                     result.push_str(&indent);
                     result.push_str("  "); // 2 spaces for nested content
-                    result.push_str(line.trim_start());
+                    // `serialize_yaml_value` anchors its output at column 0, so
+                    // each line's own leading whitespace *is* its depth relative
+                    // to the value's root. Trimming it here would flatten every
+                    // nested level into a sibling of the top-level key.
+                    result.push_str(line);
                 }
             }
             result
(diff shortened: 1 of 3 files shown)


woodruffw:
Thanks @dmbuil, this looks reasonable to me.

(I fixed a handful of cases by generalizing the `Remove` op a bit more. I've also added some more backstop cases that still aren't handled correctly, but those require a larger fix than I think it appropriate for this PR.)

woodruffw (approved):

PR #2310, opened 2026-08-14
docs: add note to 'concurrency-limit' about pedantic persona

trueberryless:
## Pre-submission checks

Please check these boxes:

      one first).

      use that violates the policy.

## Summary

I noticed that the [`concurrency-limit` section](https://docs.zizmor.sh/audits/#concurrency-limits) does not mention that it is a persona `pedantic`-only rule currently. The reason for this being that

> "[...] they'll be pretty noisy for the average user" - [William](https://github.com/zizmorcore/zizmor/pull/1227#discussion_r2427753229)

While it might change in the future, I still think it is best to document it in the meantime.

### DIFF
--- a/docs/audits.md
+++ b/docs/audits.md
@@ -406,8 +406,7 @@ by workflow and job identifiers, rather than run IDs.
 
 !!! note
 
-    This is a `--pedantic` only audit because it is pretty noisy for the average user.
-    In the future, this rule might apply to persona `regular` as well.
+    This is a `--pedantic` only audit, as it is pretty noisy for the average user.
 
 Other resources:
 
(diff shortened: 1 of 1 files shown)


docs/audits.md
```diff
@@ -404,6 +404,11 @@ resource waste vector for attackers, particularly on billed runners. Separately,
 it can be a source of subtle race conditions when attempting to locate artifacts
 by workflow and job identifiers, rather than run IDs.
 
+!!! note
+
+    This is a `--pedantic` only audit because it is pretty noisy for the average user.
+    In the future, this rule might apply to persona `regular` as well.
```
woodruffw:
```suggestion
    This is a `--pedantic` only audit, as it is pretty noisy for the average user.
```

woodruffw (approved):
Thank you @trueberryless!

woodruffw:
NB: I removed the "future" qualification, since it's true in general that everything about zizmor can change in the future. I think it's fine for the docs to just state what's true at the moment.

PR #2317, opened 2026-08-17
Split zizmor internal crates

woodruffw:
This is an attempt to do a machine refactor of zizmor's internals into a few different crates, as part of reducing compile times and generally making the codebase more structured.

I don't know if I'm going to merge this yet.


woodruffw:
Not happy with the results here.

PR #2327, opened 2026-08-21
fix(yamlpatch): reject Op::Replace on multi-key flow mappings

### LINKED ISSUE #1579: [BUG]: `Op::Replace` overrides multi-key flow mappings
### Pre-submission checks

### zizmor version

main

### Expected behavior

`handle_flow_mapping_value_replacement` in `yamlpatch` incorrectly handles flow mappings with multiple keys. When replacing a value inside a multi-key flow mapping like `{ a: 1, b: 2, c: 3 }`, the function loses all other keys and assigns the new value to the wrong key.

When replacing a value at route `foo/b` in:
```yaml
foo: { a: 1, b: 2, c: 3 }
```

Expected result:
```yaml
foo: { a: 1, b: newvalue, c: 3 }
```

### Actual behavior

```yaml
foo: { a: newvalue }
```

### Reproduction steps

Add this test to `crates/yamlpatch/tests/unit_tests.rs`:

```rust
#[test]
fn test_replace_value_in_multi_key_flow_mapping() {
    let original = r#"
foo: { a: 1, b: 2, c: 3 }
"#;

    let document = yamlpath::Document::new(original).unwrap();

    let operations = vec![Patch {
        route: route!("foo", "b"),
        operation: Op::Replace(serde_yaml::Value::String("newvalue".to_string())),
    }];

    let result = apply_yaml_patches(&document, &operations).unwrap();

    // This assertion fails
    insta::assert_snapshot!(format_patch(result.source()), @r"
    --- PATCH ---

    foo: { a: 1, b: newvalue, c: 3 }

    --- END PATCH ---
    ");
}
```

### Logs

```plain text
❯ cargo test --package yamlpatch test_replace_value_in_multi_key_flow_mapping
    Finished `test` profile [unoptimized + debuginfo] target(s) in 0.16s
     Running unittests src/lib.rs (target/debug/deps/yamlpatch-411c94bd3b1527ab)

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s

     Running tests/unit_tests.rs (target/debug/deps/unit_tests-b5ac80fe741d4558)

running 1 test
test test_replace_value_in_multi_key_flow_mapping ... FAILED

failures:

---- test_replace_value_in_multi_key_flow_mapping stdout ----
━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ Snapshot Summary ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
Snapshot: replace_value_in_multi_key_flow_mapping
Source: crates/yamlpatch/tests/unit_tests.rs:3549
────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
Expression: format_patch(result.source())
────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
-old snapshot
+new results
────────────┬───────────────────────────────────────────────────────────────────────────────────────────────────────────────────
    1     1 │ --- PATCH ---
    2     2 │ 
    3       │-foo: { a: 1, b: newvalue, c: 3 }
          3 │+foo: { a: newvalue }
    4     4 │ 
    5     5 │ --- END PATCH ---
────────────┴───────────────────────────────────────────────────────────────────────────────────────────────────────────────────
To update snapshots run `cargo insta review`
Stopped on the first failure. Run `cargo insta test` to run all snapshots.

thread 'test_replace_value_in_multi_key_flow_mapping' (3231255) panicked at /Users/dm/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/insta-1.46.1/src/runtime.rs:719:13:
snapshot assertion for 'replace_value_in_multi_key_flow_mapping' failed in line 3549
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

### Additional context

While this is a bug in the API, I'm unsure if it was meant to be generic (like the `handle_flow_mapping_addition` that unserialize, add the key and serialize again).
Feel free to close the issue if that's not something worth to fix.

hdimer:
## Pre-submission checks

Please check these boxes:

      one first).

      use that violates the policy.

## Summary

Closes #1579.

`handle_flow_mapping_value_replacement` rebuilds the whole flow mapping around
one member: it finds the first colon and emits `{ key: newvalue }`. That's
fine when there's exactly one member, but with more than one it just drops
the rest, so replacing `foo/b` in `foo: { a: 1, b: 2, c: 3 }` produced
`foo: { a: newvalue }` instead of `foo: { a: 1, b: newvalue, c: 3 }`.

Same code path also runs under `Op::MergeInto` whenever the merge target
already exists in a flow mapping, so merging into an inline `with:` with more
than one key hit this too (e.g. it silently rewrote
`with: { fetch-depth: 0, persist-credentials: true }` down to
`with: { fetch-depth: false }`).

You said in the issue the fix could be "handle it correctly" or "reject it as
unsupported," and reject was fine — that's what I did. Before rebuilding
the mapping, it now re-parses the fragment as a `Mapping` and bails with an
error if it has more than one key, or if it doesn't parse at all (which
happens for a flow mapping containing an alias whose anchor lives elsewhere
in the document — you can't tell how many members it has without resolving
that first).

I did look at actually fixing the replacement instead of rejecting it — there's
a `query_exact` call nearby that resolves the exact span of a single member,
which would let you splice in the new value without touching the rest. But
that's a bigger, riskier change than what the issue asked for, so I left it as
a TODO next to the guard and kept this PR to the reject path.

## Test Plan

Added three tests: replacing in a two-key flow mapping, replacing in the
unparsable-alias case above, and the `MergeInto`-into-`with:` case, since that
one goes through a different lowering path than a plain `Op::Replace` and
wasn't otherwise covered.

`cargo test` and `cargo fmt --check` are both clean on the workspace.

Disclosure per the AI policy: I used Claude for the implementation and the
tests, and to help me dig into the `query_exact` alternative above.

### DIFF
--- a/crates/yamlpatch/src/lib.rs
+++ b/crates/yamlpatch/src/lib.rs
@@ -362,7 +362,8 @@ fn apply_single_patch(
             let feature = route_to_feature_pretty(&patch.route, document)?;
 
             // Get the replacement content
-            let replacement = apply_value_replacement(&feature, document, value, true)?;
+            let replacement =
+                apply_value_replacement(&feature, document, value, true, &patch.route)?;
 
             // Extract the current content to calculate spans
             let current_content = document.extract(&feature);
@@ -1010,6 +1011,7 @@ fn apply_value_replacement(
     doc: &yamlpath::Document,
     value: &yaml_serde::Value,
     support_multiline_literals: bool,
+    route: &yamlpath::Route<'_>,
 ) -> Result<String, Error> {
     // Extract the current content to see what we're replacing
     let current_content_with_ws = doc.extract_with_leading_whitespace(feature);
@@ -1033,6 +1035,7 @@ fn apply_value_replacement(
             end_byte,
             current_content_with_ws,
             value,
+            route,
         );
     }
 
@@ -1126,12 +1129,29 @@ fn handle_flow_mapping_value_replacement(
     _end_byte: usize,
     current_content: &str,
     value: &yaml_serde::Value,
+    route: &yamlpath::Route<'_>,
 ) -> Result<String, Error> {
     let val_str = serialize_yaml_value(value)?;
     let val_str = val_str.trim();
 
     // Parse the flow mapping content to understand the structure
     let trimmed = current_content.trim();
+
+    // TODO: Remove this limitation.
+    // Everything below rebuilds the mapping around a single member, so it's
+    // only correct when the mapping has exactly one. Against a larger mapping
+    // it drops every other member, so reject rather than emit an incorrect
+    // patch. A failure to parse is also a rejection: we can't confirm the
+    // member count, and guessing is what produced the bad patch in the first
+    // place.
+    match yaml_serde::from_str::<yaml_serde::Mapping>(trimmed) {
+        Ok(mapping) if mapping.len() <= 1 => (),
+        _ => {
+            return Err(Error::InvalidOperation(format!(
+                "replace operation is not permitted against multi-key flow mapping route: {route:?}"
+            )));
+        }
+    }
 
     // Case 1: { key: } - has colon, empty value
     if let Some(colon_pos) = trimmed.find(':') {
--- a/crates/yamlpatch/tests/unit_tests.rs
+++ b/crates/yamlpatch/tests/unit_tests.rs
@@ -4322,3 +4322,74 @@ outer:
     --- END PATCH ---
     ");
 }
+
+/// `Op::Replace` rebuilds a flow mapping around a single member, so against a
+/// larger mapping it would drop the others.
+///
+/// This is a backstop test; at some point this should be supported.
+#[test]
+#[should_panic = "replace operation is not permitted against multi-key flow mapping"]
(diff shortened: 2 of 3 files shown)


woodruffw:
Hi @hdimer, please edit the PR description to match your own words and understanding. Per our AI policy, we don't review AI contributions unless the submitter has demonstrated that they understand their own change using their own words.

hdimer:
Fair, the original description read more like an AI writeup than my own explanation. Rewrote it in my own words above — happy to walk through any part of the fix or tests in more detail if useful.

woodruffw:
That's clearly still an AI comment. Please follow our AI policy when interacting with this projects; subsequent examples of this will result in a ban.

hdimer:
Will,

It wasn't, sorry to say. I'll stay away from this project.

PR #2330, opened 2026-08-22
Express version bounds in action coordinates

### LINKED ISSUE #2320: [BUG]: `cache-poisoning` should pass on `astral-sh/setup-uv` on v10+ with `enable-cache: auto`
### Pre-submission checks

### zizmor version

1.29.0

### Expected behavior

Zizmor should allow the use of `enable-cache: auto` with `setup-uv` so long as it's v10 or later

### Actual behavior

Zizmor yields the cache poisoning audit

### Reproduction steps

```yml
uses: astral-sh/setup-uv@ae62891fec2bb8e7d6c99fc78c9fec3a63790f8d # v10.0.0
with:
  enable-cache: auto
```

### Logs

```plain text

```

### Additional context

`enable-cache: auto` now disables caching on events that are sensitive to cache poisoning. See the release notes: https://github.com/astral-sh/setup-uv/releases#release-v10.0.0

Note: `enable-cache: auto` is also the default setting if the value is omitted, which currently also raises an audit

woodruffw:
This will fix #2320, but I need to think more through the implementation.

There are really two things here:

1. The coordinate API is now aware of "version bounds," i.e. it can express a constraint like "match this `uses:` clause but only for certain versions."
2. Doing (1) revealed that we were conflating "coordinate is satisfied" (i.e. not default satisfied) with "the `uses:` clause has a sibling `with:` clause," which is not actually guaranteed to be true now that we're treating version bounds as potentially satisfying on their own.

The solution to (2) is a pretty fundamental refactor to how coordinate matching is implemented: the `usage` API now returns a `Usage` that splits the satisfaction state (Enabled/Conditional/Always) from one or more "control origins," which encode the "why" of the satisfaction state. The `cache-poisoning` audit then consumes those origins to produce its locations. A nice consequence of that is that our other diagnostics from this audit become cleaner, which can be seen in the snapshots.

Doing this has regressed the fix generation for `cache-poisoning`, since we can no longer easily "invert" an `ActionCoordinate` to produce its fixes.

### DIFF
--- a/crates/zizmor/tests/integration/audit/cache_poisoning.rs
+++ b/crates/zizmor/tests/integration/audit/cache_poisoning.rs
@@ -454,18 +454,7 @@ fn test_issue_1081() -> anyhow::Result<()> {
        |
        = note: audit confidence → Low
 
-    error[cache-poisoning]: runtime artifacts potentially vulnerable to a cache poisoning attack
-      --> @@INPUT@@:23:9
-       |
-     5 | on: release
-       | ----------- generally used when publishing artifacts generated at runtime
-    ...
-    23 |       - uses: astral-sh/setup-uv@d9e0f98d3fc6adb07d1e3d37f3043649ddad06a1
-       |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ action version may enable caching
-       |
-       = note: audit confidence → Low
-
-    4 findings (1 suppressed): 0 informational, 0 low, 0 medium, 3 high
+    3 findings (1 suppressed): 0 informational, 0 low, 0 medium, 2 high
     "
     );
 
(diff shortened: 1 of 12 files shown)


crates/zizmor/tests/integration/test-data/cache-poisoning/issue-1081-repro.yml
```diff
@@ -19,6 +19,9 @@ jobs:
         with:
           enable-cache: true
 
+      # TRUE POSITIVE: action version is unknown
+      - uses: astral-sh/setup-uv@d9e0f98d3fc6adb07d1e3d37f3043649ddad06a1
```
woodruffw: NB: Need to move this.

crates/zizmor/tests/integration/audit/cache_poisoning.rs
```diff
@@ -82,25 +81,23 @@ fn test_caching_opt_in_expression() -> anyhow::Result<()> {
...
-    21 | |           python-version: "3.12"
-    22 | |           enable-cache: ${{ github.ref == 'refs/heads/main' }}
-       | |______________________________________________________________- may enable caching here
+    22 |           enable-cache: ${{ github.ref == 'refs/heads/main' }}
+       |           ---------------------------------------------------- may enable caching here
        |
        = note: audit confidence → Low
-       = note: this finding has an auto-fix
 
-    3 findings (1 ignored, 1 suppressed, 1 unsafe fixes): 0 informational, 0 low, 0 medium, 1 high
```
woodruffw, line 102: NB need to investigate why the unsafe fix disappeared.

woodruffw: Deferring this for another PR.

PR #2346, opened 2026-08-29
Use Docker's github-builder for image builds

woodruffw:
This is a nice simplification over the existing Docker build/push mess. Plus it does the attestation stuff by default that I haven't had time to figure out.

(Thank you @crazy-max! The experience of integrating this was very smooth.)


woodruffw:
Working as expected: https://github.com/zizmorcore/zizmor/actions/runs/33267666790

PR #2349, opened 2026-08-29
expressions: add a recursion budget

### LINKED ISSUE #2348: Add a depth guard for recursions during expression parsing
I received a slop security report for this. It's not a vulnerability (we don't consider availability to be a risk in zizmor, since you can "crash" it with any number of invalid inputs), but something we should still produce a useful error for instead of a hard termination.

TL;DR: `GitHub-actions-expressions` is currently a naive recursive descent parser. We should set a recursion budget (256?) to prevent unbounded recursion on user inputs. Inputs that exceed this budget will get a normal `Error` variant (`Error::Depth` or similar?) rather than a hard termination.

woodruffw:
Fixes #2348

### DIFF
--- a/crates/github-actions-expressions/src/parser.rs
+++ b/crates/github-actions-expressions/src/parser.rs
@@ -16,6 +16,9 @@ use crate::{
 /// The result of parsing a single grammar production.
 type PResult<'src> = Result<SpannedExpr<'src>, Error>;
 
+/// GitHub's maximum expression depth.
+const RECURSION_BUDGET: usize = 50;
+
 /// Parse a complete GitHub Actions expression. The whole input must be
 /// consumed (modulo whitespace).
 pub(crate) fn parse(src: &str) -> Result<SpannedExpr<'_>, Error> {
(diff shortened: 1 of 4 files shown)


crates/github-actions-expressions/src/parser.rs
```diff
@@ -16,20 +16,24 @@ use crate::{
 /// The result of parsing a single grammar production.
 type PResult<'src> = Result<SpannedExpr<'src>, Error>;
 
+/// GitHub's maximum expression depth.
```
woodruffw, line 19: NB I originally picked 256, but GitHub's own public parser uses this limit. Seems fine to align.

crates/github-actions-expressions/tests/languageservices_kat.rs
```diff
@@ -76,8 +76,8 @@ fn to_evaluation(result: &TestResult) -> Result<Evaluation, String> {
 /// Entire test suite groups to skip.
 static SKIPPED_GROUPS: LazyLock<HashSet<(&str, &str)>> = LazyLock::new(|| {
     HashSet::from_iter([
-        // We don't enforce recursion limits at the moment.
-        // TODO: Do so?
+        // GitHub also applies its depth limit to iterative postfix chains,
+        // while we only limit recursive parser productions.
         ("syntax-errors.json", "depth-errors"),
```
woodruffw, line 81: NB we still can't enable these tests since our parser doesn't recurse for as many productions as GitHub's does. So we don't fail on many of these. I guess in the future we could try to separate them out more, or force ourselves to fail on them by consuming budget even when unnecessary.

PR #2364, opened 2026-09-04
Add zizmor-http-cache

woodruffw:
WIP. This is essentially a wholesale lift of uv's `httpcache` (with attribution).

I'm not sure if I'll use this yet.

### DIFF
--- a/crates/zizmor/src/cli.rs
+++ b/crates/zizmor/src/cli.rs
@@ -4,12 +4,10 @@ use std::io::stdout;
 
 use annotate_snippets::renderer::{AnsiColor, Effects};
 use anstream::stream::IsTerminal;
-use camino::Utf8PathBuf;
 use clap::builder::{NonEmptyStringValueParser, Styles};
 use clap::{ArgAction, Args, Parser, ValueEnum, ValueHint};
 use clap_complete::Generator;
 use clap_verbosity_flag::InfoLevel;
-use etcetera::AppStrategy as _;
 
 use crate::config::Config;
 use crate::finding::Persona;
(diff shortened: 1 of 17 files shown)


crates/zizmor/src/cli.rs
```diff
@@ -213,17 +197,6 @@ pub(crate) struct NetworkArgs {
...
-    /// The directory to use for HTTP caching. By default, a
-    /// host-appropriate user-caching directory will be used.
-    #[arg(
-        long,
-        value_name = "DIR",
-        default_value_t = App::default_cache_dir(),
-        hide_default_value = true,
-        value_hint = ValueHint::DirPath
-    )]
-    pub(crate) cache_dir: Utf8PathBuf,
```
woodruffw, line 226: NB probably need to retain this as a no-op for compatibility, or maybe re-consider adding back a FS caching backend.

crates/zizmor/src/github.rs
```diff
@@ -282,15 +271,11 @@ pub(crate) struct Client {
     token: GitHubToken,
     base_client: ClientWithMiddleware,
     api_client: ClientWithMiddleware,
-    ref_cache: MokaCache<String, Vec<RemoteHead>>,
```
woodruffw, line 285: Not sure about this yet, maybe we should keep moka.

docs/usage.md
```diff
@@ -947,32 +947,11 @@ To do this, you can pass `--no-ignores` during audits:
 zizmor --no-ignores example/example
 ```
 
-## Caching between runs
+## Caching within a run
```
woodruffw, line 950: This should be removed and should become a doc stub.

woodruffw:
I think these perf numbers are unreliable.

PR #2394, opened 2026-09-18
ref-version-mismatch: handle unusual comment placement

### LINKED ISSUE #2393: [BUG]: Crash when `uses:` is represented as a literal block scalar with a comment
### Pre-submission checks

### zizmor version

1.30.1

### Expected behavior

I expect zizmor to handle a valid GitHub Actions uses: value represented as a YAML literal block scalar.

For example:

```yaml
uses: |-  # v1.2.3

  astral-sh/setup-uv@20cfd1bf945f4377ade1205e4dbc17946fc9a30d
```

This is valid YAML and the resulting scalar value is:

astral-sh/setup-uv@20cfd1bf945f4377ade1205e4dbc17946fc9a30d

I would expect zizmor to treat this equivalently to the conventional representation:

```yaml
uses: astral-sh/setup-uv@20cfd1bf945f4377ade1205e4dbc17946fc9a30d  # v1.2.3
```

In particular, the ref-version-mismatch audit should process the uses: reference without crashing.

### Actual behavior

```
Panic occurred in file 'crates/zizmor/src/finding/location.rs' at line 354
cause = "subfeature does not occur within feature"

The relevant part of the backtrace is:

<zizmor::finding::location::Feature>::from_subfeature::<
    zizmor::models::workflow::ReusableWorkflowCallJob
>
<zizmor::audit::ref_version_mismatch::RefVersionMismatch
    as zizmor::audit::Audit>::audit_step
```

### Reproduction steps

Create a workflow containing:

```yaml
jobs:
  test:
    uses: |-  # v1.2.3

      astral-sh/setup-uv@20cfd1bf945f4377ade1205e4dbc17946fc9a30d
```

Run zizmor with the auditor persona:

```
GH_TOKEN="$(gh auth token)" uv run zizmor \
  --persona=auditor \
  --strict-collection \
  .github
```

zizmor crashes with:
subfeature does not occur within feature

### Logs

```plain text
name = "zizmor"
operating_system = "Linux (Ubuntu 24.04) [x86_64]"
crate_version = "1.30.1"

explanation = """
Panic occurred in file 'crates/zizmor/src/finding/location.rs' at line 354
"""

cause = "subfeature does not occur within feature"
method = "Panic"

The panic occurs through:

Feature::from_subfeature::<ReusableWorkflowCallJob>
RefVersionMismatch::audit_step
RefVersionMismatch::audit_normal_job
RefVersionMismatch::audit_workflow
```

### Additional context

The important difference from a normal uses: value is the YAML representation.

The uses: value is a literal block scalar:

```
uses: |-
  astral-sh/setup-uv@20cfd1bf945f4377ade1205e4dbc17946fc9a30d
```

and the version comment is attached to the block scalar indicator:

```
uses: |-  # v1.2.3
```

rather than being an end-of-line comment on the scalar value.

This representation is useful when tooling needs to preserve comments while constructing YAML, and it is valid YAML. I believe zizmor should support it rather than assuming that uses: values are always plain/quoted scalars.

It may be worth testing the same case without the version comment as well, but the crash I encountered occurs with the # v1.2.3 comment present and appears to go through the ref-version-mismatch audit.

The crash also appears to occur while handling ReusableWorkflowCallJob, so the issue may be related to how the audit maps the block scalar/comment location back to the workflow feature.

woodruffw:
~~WIP, needs more tests.~~

The underlying bug here is that `ref-version-mismatch` assumed that a `uses:` clause that has a comment will always have that comment _after_ the actual content. However, that's not true if the user does something like this:

```yaml
uses: |- # 1.2.3
  abc/def
```

This is inadvisable since tools like Renovate and Dependabot won't handle it correctly, but it still shouldn't crash zizmor. The fix on zizmor's side is to track each comment's concrete location and use that for the subfeature rather than trying to guesstimate the subfeature's location relative to the end of the content in `uses:`.

Closes #2393.


crates/zizmor/src/audit/ref_version_mismatch.rs
```diff
@@ -30,26 +30,31 @@ audit_meta!(
...
 );
 
-#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+#[derive(Debug)]
 enum CommentVersionState<'doc> {
     Missing,
-    Version(&'doc str),
+    Version {
+        version: &'doc str,
+        comment: &'doc Comment<'doc>,
```
woodruffw, line 38: This feels wrong, I need to think a bit more about the representation here.

crates/zizmor/src/audit/ref_version_mismatch.rs
```diff
@@ -30,26 +30,31 @@ audit_meta!(
...
         for comment in comments {
             if let Some(version) = RawVersion::from_comment(comment) {
-                return Some(version.as_raw());
+                return Some((version.as_raw(), comment));
             }
         }
         None
     }
 
     fn comment_version_state<'doc>(comments: &'doc [Comment<'doc>]) -> CommentVersionState<'doc> {
```
woodruffw, line 55: This should probably be `version_comment_state` or `detect_version_comment`.

crates/zizmor/src/finding/location.rs
```diff
@@ -303,17 +303,37 @@ static_regex!(IGNORE_EXPR, r"# zizmor: ignore\[(.+)\](?:\s+.*)?$");
 /// Represents a single source comment.
 #[derive(Debug, Serialize)]
 #[serde(transparent)]
-pub(crate) struct Comment<'doc>(&'doc str);
+pub(crate) struct Comment<'doc> {
+    raw: &'doc str,
+    /// The comment's concrete location within its containing document.
+    #[serde(skip)]
```
woodruffw, line 309: I guess we could not skip this.

crates/zizmor/src/audit/ref_version_mismatch.rs
woodruffw: Seems unavoidable for now.

Issue #2366, opened 2026-09-05
Audits should be able to spot findings based on indirect matrix expansions


ubiratansoares:
### Pre-submission checks

- [x] I am **not** reporting a bug (crash, false positive/negative, etc). These must be filed via the bug report template.
- [x] I have looked through both the [open](https://github.com/zizmorcore/zizmor/issues?q=is%3Aissue+is%3Aopen+) and [closed](https://github.com/zizmorcore/zizmor/issues?q=is%3Aissue%20state%3Aclosed) issues for a duplicate request.
- [x] I affirm that I have followed the [AI policy](https://github.com/zizmorcore/.github/blob/main/AI_POLICY.md) and that this request constitutes my own words and understanding. 

### What's the problem this feature will solve?

While working with https://github.com/zizmorcore/zizmor/pull/2350 I took your suggestion, and I had a chance to check `matrix.rs`, learning about the some opportunities to improve matrix expansions. I'm materializing this issue to back an implementation proposal :) 

Some audits that use `matrix::Expansions` API (`crates/zizmor/src/models/workflow/matrix.rs` ) may want to know whether the entire matrix is indirectly defined or whether inclusions or exclusions happen via indirection/expressions. These signals could enable audits to report findings based on such indeterminations with the proper confidence / persona (low / pedantic I guess?).

As an example : for zizmor 1.30.0, if one runs the `pendantic` persona against this [Homebrew workflow](https://github.com/Homebrew/homebrew-core/blob/main/.github/workflows/tests.yml) one will see several `unpinned-images` errors, but we could also spot something like:

```
error[unpinned-images]: unpinned image references
   --> path/to/homebrew-core/.github/workflows/tests.yml:365:5
    |
360 |       matrix:
    |       ------ this matrix
361 |         include: ${{fromJson(needs.setup_dep_runners.outputs.runners)}}
    |         --------------------------------------------------------------- values are populated indirectly here
...
365 |     container: ${{matrix.container}}
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ container image may be unpinned
    |
    = note: audit confidence → Low
```

since [the related matrix](https://github.com/Homebrew/homebrew-core/blob/main/.github/workflows/tests.yml#L360) is fully indirect.

### Describe the solution you'd like

`matrix::Expansions` struct already holds the all walked expansions, but we could also track indetermination sources like

```rust
pub(crate) struct Expansions<'doc> {
[...]

PR #2395, opened 2026-09-20
feat(unpinned-images): add support to indirect matrices

ubiratansoares:
## Pre-submission checks

Please check these boxes:

      one first).

      use that violates the policy.

## AI disclosure

I wrote the code by hand, and I asked Claude to review this implementation before raising this PR. I also asked it to prepare a test sample that I reviewed and refined afterwards.

## Summary

This PR adds better handling for matrix indirect expansions, improving a few bits regarding precision and behaviour along the way. Here we :

- extend the `Expansions` API to track whether there are any indetermintations/indirections when expanding matrix values
- improve the internal matrix expansion logic, mirroring what Github does, ie, computing matrix exclusions before inclusions
- also adapt the code so `Expansions` is constructed only once alongside with the `Matrix` instance, being now part of this struct
- use these changes to improve the `unpinned-images` audit 

This PR is the first step towards https://github.com/zizmorcore/zizmor/issues/2366. I'm rising it to validate the direction and using the `unpinned-images` audit as a validation target. If everything looks ok, `self-hosted-runner` and `template-injection` to follow on another PR (for the sake of easier reviews).

Although this PR improves the precision of `unpinned-images` and allows us to flag more cases, I did not change the confidence of related findings, they are still tracked as `low`.

## Test Plan

I added some unit tests covering the new `Expansions` APIs and also an integration test for `unpinned-images`, mixing (on purpose) indirections with static expansions this audit should already flag. The new inner implementation of `Matrix` does not trigger any regressions based on integration tests, so `cargo test` should be green.


woodruffw:
Thank youn @ubiratansoares! I'll try and review this today.

crates/zizmor/src/audit/unpinned_images.rs
```diff
@@ -186,42 +186,74 @@ fn candidates_for_leaf<'doc>(
                 return vec![];
             };
 
-            matrix
-                .expansions()
+            let expansions = matrix.expansions();
+
+            // Evaluate first candidates from static expansions
```
woodruffw: Nitpick: 


```suggestion
            // First: evaluate candidates from static expansions
```

crates/zizmor/src/audit/unpinned_images.rs
```diff
@@ -186,42 +186,74 @@ fn candidates_for_leaf<'doc>(
...
+                    ImageCandidate::concrete(
+                        &DockerUses::parse(&expansion.value),
+                        location.clone(),
+                        annotations,
+                    )
                 })
-                .collect()
+                .collect::<Vec<_>>();
+
+            // Evaluate also indirect expansions
```
woodruffw:
```suggestion
            // Next: evaluate indirect expansions
```

crates/zizmor/src/audit/unpinned_images.rs
```diff
@@ -186,42 +186,74 @@ fn candidates_for_leaf<'doc>(
...
+            // An indirect matrix, dimensions block means this path may
+            // take values we never saw -- possibly all of them.
+            if expansions.has_indirect_expansions() {
+                let mut annotations = vec![matrix.location().key_only().annotated("this matrix")];
+
+                let indirect_matrix = annotated_indetermination(
+                    expansions.indirectly_expanded(),
+                    "indirect `matrix` adds combinations we can't see",
```
woodruffw: Nit: maybe something like "adds unanalyzable combinations" here and below? We don't generally use "we" elsewhere in zizmor's annotations 🙂

crates/zizmor/src/audit/unpinned_images.rs
```diff
@@ -186,42 +186,74 @@ fn candidates_for_leaf<'doc>(
...
+
+            candidates
         }
         // Any other leaf (non-`matrix` context, function call, etc.) can't be
         // analyzed statically.
         _ => vec![ImageCandidate::opaque(location, vec![])],
     }
 }
 
+fn annotated_indetermination<'doc>(
```
woodruffw: I'd be okay with removing this API and just inlining the call it makes, it seems relatively trivial 🙂

crates/zizmor/src/models/workflow/matrix.rs
```diff
@@ -235,19 +298,20 @@ impl<'doc> Matrix<'doc> {
...
+    pub(crate) fn expansions(&self) -> &Expansions<'doc> {
+        &self.expansions
     }
 
     /// Checks whether some expanded path leads to an expression
     pub(crate) fn expands_to_static_values(&self, context: &Context) -> bool {
         // If we have an indirect matrix, we can't determine whether it expands to
         // static values or not.
-        if matches!(self.inner, LoE::Expr(_)) {
+        if self.expansions().indirectly_expanded.is_some() {
```
woodruffw, line 314: Nice!

crates/zizmor/src/models/workflow/matrix.rs
```diff
@@ -235,19 +298,20 @@ impl<'doc> Matrix<'doc> {
...
-    pub(crate) fn expansions(&self) -> Expansions<'doc> {
-        Expansions::new(self)
+    /// The expansions for this matrix
+    pub(crate) fn expansions(&self) -> &Expansions<'doc> {
+        &self.expansions
     }
 
     /// Checks whether some expanded path leads to an expression
     pub(crate) fn expands_to_static_values(&self, context: &Context) -> bool {
         // If we have an indirect matrix, we can't determine whether it expands to
```
woodruffw: Nit: I think my old comment here is now imprecise:


```suggestion
        // If any part of the matrix is indirect, we can't determine whether it expands to
```

woodruffw (commented):
Thanks @ubiratansoares! This looks great to me overall, just a few tiny nitpicks.

ubiratansoares:
Thanks @woodruffw ! I'll get your suggestions sorted asap :)

crates/zizmor/src/audit/unpinned_images.rs
ubiratansoares: I tried to dedup a bit and extracted this one out, but happy to inline (back) 😄

ubiratansoares:
Hopefully all nits are sorted :)

woodruffw (approved):
LGTM, thanks a ton @ubiratansoares!

crates/zizmor/tests/integration/audit/unpinned_images.rs
```diff
@@ -226,6 +226,91 @@ fn test_matrix_in_image_regular() -> anyhow::Result<()> {
...
+      --> @@INPUT@@:45:7
+       |
+    45 |       image: ${{ matrix.image }}
+       |       ^^^^^^^^^^^^^^^^^^^^^^^^^^ container image may be unpinned
+    ...
+    48 |       matrix:
+       |       ------ this matrix
+    ...
+    54 |         exclude: ${{ fromJSON(vars.KNOWN_BROKEN_COMBINATIONS) }}
+       |         -------------------------------------------------------- `exclude` may remove unanalyzable combinations
```
woodruffw, line 303: Sorry, I didn't notice this one before: shouldn't this one be fine?

My thinking is that `exclude` is subtractive so it actually can't _add_ anything unanalyzable here, only remove things.

ubiratansoares: Although `exclude` can't add combinations, on that particular matrix one of the expansions (`ubuntu-latest`) is a legit finding, and we can't know whether   `fromJSON(vars.KNOWN_BROKEN_COMBINATIONS)` filters it out. 

Hence, I opted for show it, but I don't have a strong opinion regarding this take.

If you agree with, then ... I think we should also skip flagging these indeterminate exclusions if there is no image candidates, ie, a matrix like:

```yml
jobs:
  indirect-matrix:
    name: indirect-matrix
    runs-on: ubuntu-latest
    container:
      image: ${{ matrix.image }}
    strategy:
      matrix: ${{ fromJSON(vars.CUSTOM_TARGETS) }}
      exclude: ${{ fromJSON(vars.KNOWN_BROKEN_COMBINATIONS) }}
    steps:
      - run: true
```

should not report `exclude`, I guess? Eventually there is a missing `if` statement left in this case.

Happy to push another commit, either improveing the current take or just don't flagging exclusions at all. :)

woodruffw: Ah, I see what you mean -- you're thinking that because the indirect `exclude` could _remove_ the finding, we should flag it? 

I think my 0.02c is for leaving that out for now -- IIUC we're already flagging the legitimate `ubuntu-latest` finding in that case, so flagging the `exclude` is probably going to be too noisy/not helpful to users who already have a finding.



> If you agree with, then ... I think we should also skip flagging these indeterminate exclusions if there is no image candidates, ie, a matrix like:

Yes, agreed! I think for now we just don't want to flag `exclude` at all 🙂 -- we can always revisit down the road.

ubiratansoares: > Ah, I see what you mean -- you're thinking that because the indirect exclude could remove the finding, we should flag it?

Yes, that's what I meant. :) 

Removed [here](https://github.com/zizmorcore/zizmor/pull/2395/changes/f3a4c0c98a8880fdd9da5ef54b92e5fe5031d29b), but keep the `indirect_exclusions` in case we find uses for it in the future.

PR #2396, opened 2026-09-20
dangerous-triggers: mark issue_comment as dangerous

### LINKED ISSUE #1606: Enhance `dangerous-triggers`: flag `issue_comment` as a dangerous trigger

woodruffw:
This was a long-running oversight. I've been hesitant to add it in the past since I suspect it'll be disruptive, but it's exactly as insecure as `pull_request_target` and deserves to be flagged as such.

Closes #1606 

Closes #1614


woodruffw:
Note that this is much simpler than the approach in #1614, which mediates the finding through a check for expressions that check the `author_association`. That may be what we want long term.

